PR #5, opened 2019-12-17
Make target scaffolding for lint, vet, fmt, builds, etc...

davidvossel:


davidvossel:
/cc
@slintes 
@fedepaol 
@cynepco3hahue 

fyi, I'm not following normal merge procedures here. This is a bunch of boiler plat stuff for Dev enablement. I'm self merging in order to quickly get this repo setup for CI and dev.

Retroactively review and make comments if you have questions.

openshift-ci/Dockerfile.tools
```diff
@@ -0,0 +1,37 @@
+FROM fedora:30 as build-tools
```
fedepaol: I made some changes on the baremetal-deploy repo. One of those was using fedora:31.
Can you look at that?
Also, since it's shipping ginkgo, I'd remove the rule from the Makefile or (at least) remove the dependency of the `functests` rule from the `ginkgo` rule.

PR #7, opened 2019-12-17
updates to prow build root container

davidvossel:
Signed-off-by: David Vossel <davidvossel@gmail.com>


davidvossel:
this aligns closer with what we have in baremetal-deploy repo

PR #8, opened 2019-12-17
Only download ginkgo if it isn't in PATH

davidvossel:


Makefile
```diff
@@ -24,14 +24,9 @@ deps-update:
...
-	go install github.com/onsi/ginkgo/ginkgo
-
-functests: ginkgo
-	FOCUS=$$(echo $(FEATURES) | tr ' ' '|') && \
-	echo "Focusing on $$FOCUS" && \
-	GOFLAGS=-mod=vendor ginkgo --focus=$$FOCUS functests -- -junit /tmp/artifacts/unit_report.xml
-	#TODO - copy in functional test suite
+functests:
+	@echo "Running Functional Tests"
+	hack/build-functest.sh
```
slintes: a bit nitpicking, but for consistency I'd prefer `functests` (plural). And why `build-` and not `run-`?

davidvossel: yep, fixed both.

PR #11, opened 2019-12-18
Code generation make targets

davidvossel:
adds ```make generate``` and ```make verify-generate``` make targets to generate deep copy functions and crd manifests from our api. Validation occurs during unit-test execution to verify we don't commit api changes without the corresponding generated logic.


Makefile
```diff
@@ -13,6 +13,14 @@ export FEATURES?=mcp performance sctp
 TARGET_GOOS=linux
 TARGET_GOARCH=amd64
 
+CACHE_DIR="_cache"
+TOOLS_DIR="$(CACHE_DIR)/tools"
+
+OPERATOR_SDK_VERSION="v0.12.0"
```
slintes: v0.13.0 please :)

davidvossel: yep yep, will do!

davidvossel: fixed

davidvossel:
/cc @slintes @fedepaol 

can i get a review on this? Lets exercise the prow workflow now.

This PR adds logic to automatically generate deepcopy and crd manifests for us. I had to add a Dockerfile for the generation logic to work correctly for some reason. We'll be using that dockerfile soon, but for this PR it's just there to make everything pass.

I also added a "verify-generate" command that makes sure we don't commit changes to our api that haven't had their corresponding generated files updated. For example, this helps us avoid making an API change and then not updating the deepcopy functions.  For now I added this verification to the unit-tests make target simply because I a place to prove it works in CI. we might want to consider another CI job for this in the future.

hack/verify-generated.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if [[ -n "$(git status --porcelain pkg/apis deploy/crds)" ]]; then
+        echo "uncommitted generated files. run 'make update-generated' and commit results."
```
slintes: there is no update-generated, should probably be just generate?

PR #14, opened 2019-12-20
Add a makefile rule to setup the cluster.

fedepaol:
The setup is supposed to happen in a declarative way, using customize.
The cluster-setup/base folder needs to contain a per-feature folder with the base manifest.
The cluster-setup/ci-cluster needs to contain the related kustomized version of the base folder.
Inside the ci-cluster folder, there must be an optional wait_for_it.sh file that must return only if the given
feature is deployed and the cluster is ready.

In this way, we are going to be able to deploy both the features delivered by our operator, but also those delivered by other operators. 
The rule still relies on the `FEATURES` variable which makes the deployment AND the test modular.


MarSik:
The only thing I do not like is the base64 content. We should find a way to keep it reviewable.

Otherwise this looks good to me.

/lgtm

PR #16, opened 2019-12-20
Have the test suite consume the tests from baremetal-deploy.

fedepaol:
The per feature tests are exposed in baremetal-deploy as packages.
Blind importing them would make them available in this suite without any additional effort.

`_ "github.com/openshift-kni/baremetal-deploy/features/functests/sctp"` is an example of how to make the sctp suite available to our make functests.

This should be a nice way to have tests under baremetal-deploy until they are 4.4 ready. Once they are, they can be consumed from this side and eventually be removed from there.
In the meanwhile, new tests can be added on this side without collisions.


slintes:
nice idea! :+1:

davidvossel:
/hold

I'm putting this on hold to ensure it's discussed fully before we commit to a plan.

This is a clever approach. I have some questions though.

1. how do we handle disjoint "util like" packages that tests use for helper functions? 
2. these feature functional tests have setup logic as well in baremetal-deploy. 
3. How do we handle syncing repos (re-vendoring). There will be drift between the two repos unless some sort of merge bot is used to keep the test vendors updated.

I think there's also a larger topic we need clarity on as well.

It's unclear to me what the future is for the baremetal-deploy repo's feature dir. Would it make sense for us to bring both the features/ dir and functests dir from that repo wholesale over to performance-addon-operators? A transitional path for us then would be removing items from "features/" as they gain controller reconcile loops. Items that we have no intention of creating reconcile loops for would remain in "features/" as example feature deployments our functional tests execute against.

If we have a commitment to continued support of the baremetal-deployment's features/ dir exactly as it is today, then a transition path to performance-addon-operator would be to bring in the features dir, snapshot a supportable branch of the repo that we continue to contributed/backport fixes to, and move forward with transitioning features/ to reconcile loops in master.  

thoughts? (fyi i'm going on PTO until Jan 6th, so my future response will be delayed)

fedepaol:
Closing as per the development in repos organization

PR #19, opened 2019-12-22
Use leader election mechanism from controller-runtime library

cynepco3hahue:
There are two options for the leader election. The SDK defaults to the simpler one ("Leader for life") which trades off slower failover (5 minutes by default - the pod-eviction-timeout) for a guarantee that there can never be two leaders. The other one (used here) fails over faster, but there may be two leaders if one node is running a clock 50% faster(!) than the others unless you tweak the default timeouts to accommodate the clock skew rate you expect.

The controller-runtime library provides a way to set leader among pods via config map annotation, that includes leader lease. In case when the pod fails to update the lease under the config map, the other pod will acquire the leader.

Otherwise, the operator-SDK do not provide the lease feature and just creates the
config map, so until the config map exists, another pod can not acquire the leader,
but the config map will exist as long as pod exists(also in terminating state),
so it will prevent HA in cases when node not responsive for some reason(network, kernel panic, ...).

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -7,24 +7,25 @@ import (
 	"os"
 	"runtime"
 
-	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
-	"k8s.io/client-go/rest"
-
 	"github.com/openshift-kni/performance-addon-operators/pkg/apis"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller"
 	"github.com/openshift-kni/performance-addon-operators/version"
 
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	kubemetrics "github.com/operator-framework/operator-sdk/pkg/kube-metrics"
-	"github.com/operator-framework/operator-sdk/pkg/leader"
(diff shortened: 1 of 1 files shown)


fedepaol:
Just out of curiosity, why do we need this change? 
Since what we are applying here is "day 2 configuration", it feels like having a slow failover is bearable and we can afford not needing a quick reaction time.

fedepaol:
Apart from that
/lgtm

cynepco3hahue:
> Just out of curiosity, why do we need this change?
> Since what we are applying here is "day 2 configuration", it feels like having a slow failover is bearable and we can afford not needing a quick reaction time.

It is a good point, but I always prefer the more robust solution, also if controller-runtime provides us leader election out of the box I do not see the reason for additional import(in general IMHO operator-SDK provides a lot of stuff that by default you do not really need or controller-runtime already provides solution for it)

fedepaol:
> > Just out of curiosity, why do we need this change?
> > Since what we are applying here is "day 2 configuration", it feels like having a slow failover is bearable and we can afford not needing a quick reaction time.
> 
> It is a good point, but I always prefer the more robust solution, also if controller-runtime provides us leader election out of the box I do not see the reason for additional import(in general IMHO operator-SDK provides a lot of stuff that by default you do not really need or controller-runtime already provides solution for it)

Understood. Thanks for taking the time to explain.

PR #20, opened 2019-12-23
SR-IOV topology manager tests

cynepco3hahue:
- provide basic tests that will verify topology manager configuration
- provide SR-IOV test, the kubelet should use the same NUMA node for SR-IOV and CPU's

Tested locally until we will have SR-IOV bare metal environment under CI:
```
Running Suite: Performance Addon Operators e2e tests
====================================================
Random Seed: 1577119246
Will run 2 of 2 specs

• [SLOW TEST:13.194 seconds]
[performance]Topology Manager
/home/alukiano/go/src/github.com/openshift-kni/performance-addon-operators/functests/performance/topology_manager.go:22
  should be enabled with the best-effort policy
  /home/alukiano/go/src/github.com/openshift-kni/performance-addon-operators/functests/performance/topology_manager.go:33
------------------------------
• [SLOW TEST:62.955 seconds]
[performance]Topology Manager
/home/alukiano/go/src/github.com/openshift-kni/performance-addon-operators/functests/performance/topology_manager.go:22
  with the SR-IOV devices and static CPU's
  /home/alukiano/go/src/github.com/openshift-kni/performance-addon-operators/functests/performance/topology_manager.go:46
    should allocate resources from the same NUMA node
    /home/alukiano/go/src/github.com/openshift-kni/performance-addon-operators/functests/performance/topology_manager.go:96
------------------------------

JUnit report was created: /tmp/artifacts/unit_report.xml

Ran 2 of 2 Specs in 132.401 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```


fedepaol:
I'd like to see the `util` package split in different packages where the name declares what the package provides.
It would improve readability on the calling site, like `pods.WaitForDeletion` as opposed to `utils.WaitForPodDeletion`, or `nodes.FilterByResource`.

See https://blog.golang.org/package-names

cynepco3hahue:
> I'd like to see the `util` package split in different packages where the name declares what the package provides.
> It would improve readability on the calling site, like `pods.WaitForDeletion` as opposed to `utils.WaitForPodDeletion`, or `nodes.FilterByResource`.
> 
> See https://blog.golang.org/package-names

Done

functests/performance/topology_manager.go
```diff
@@ -0,0 +1,194 @@
...
+		Expect(kubeletConfig.TopologyManagerPolicy).To(Equal(kubeletconfigv1beta1.BestEffortTopologyManagerPolicy))
+	})
+
+	Context("with the SR-IOV devices and static CPU's", func() {
+		var testpod *corev1.Pod
+		var sriovNode *corev1.Node
+
+		BeforeEach(func() {
+			sriovNodes := nodes.FilterByResource(cs, workerRTNodes, testutils.ResourceSRIOV)
+			if len(sriovNodes) < 1 {
```
fedepaol: Don't we risk skipping if our deployment failed for some reason? Maybe it's better to fail (a bit like what you are doing above with worker-rt nodes). If we want to run the suite against a cluster without this feature, we can drive that through FOCUS / SKIP parameters)

functests/performance/topology_manager.go
```diff
@@ -0,0 +1,194 @@
...
+						"The environment does not have nodes with role %q and available %q resources",
+						testutils.RoleWorkerRT,
+						string(testutils.ResourceSRIOV),
+					),
+				)
+			}
+			sriovNode = &sriovNodes[0]
+
+			var err error
+			if testpod != nil {
```
fedepaol: How could this happen?

functests/test_suite_test.go
```diff
@@ -24,10 +31,24 @@ func TestTest(t *testing.T) {
...
 		rr = append(rr, reporters.NewJUnitReporter(*junitPath))
 	}
-	RunSpecsWithDefaultAndCustomReporters(t, "Test Suite", rr)
+	RunSpecsWithDefaultAndCustomReporters(t, "Performance Addon Operators e2e tests", rr)
 }
 
 var _ = BeforeSuite(func() {
-	//clients.Setup()
-	// Add here the setup for additional features
+	cs := testutils.NewClientSet("")
```
fedepaol: could be clients.New() (with a separate module). In that way only const remains in testutils which can then be switched to testconsts

functests/performance/topology_manager.go
cynepco3hahue: I assumed that we will have more SR-IOV topology manager tests under this context, so in general on the first test under the context `testpod` will be `nil`, but for all following tests it will be not `nil`

functests/test_suite_test.go
cynepco3hahue: will do

functests/performance/topology_manager.go
fedepaol: Ah you are right :+1:

functests/performance/topology_manager.go
cynepco3hahue: It still unclear how our baremetal CI will look, like will we have a different job for SR-IOV, or will we have a number of different environments under a single job, so we will need to check the existence of SR-IOV devices on tests runtime
Let's leave this check for the near future to ease run of all tests(locally), once we will have CI we can remove it.

fedepaol: Ok, but please add a big TODO then :-)

cynepco3hahue: hh ok :)

cynepco3hahue: done

functests/test_suite_test.go
cynepco3hahue: done

functests/test_suite_test.go
```diff
@@ -24,10 +32,24 @@ func TestTest(t *testing.T) {
...
 		rr = append(rr, reporters.NewJUnitReporter(*junitPath))
 	}
-	RunSpecsWithDefaultAndCustomReporters(t, "Test Suite", rr)
+	RunSpecsWithDefaultAndCustomReporters(t, "Performance Addon Operators e2e tests", rr)
 }
 
 var _ = BeforeSuite(func() {
-	//clients.Setup()
-	// Add here the setup for additional features
+	cs := testclient.New("")
```
fedepaol: Sorry I just noticed, how about making this global and using it in all the tests?
In that case I would have a longer (and more descriptive) name since the scope is wider.

cynepco3hahue: what name do you prefer?

fedepaol: clientSet maybe?

cynepco3hahue: hm maybe instead of global variable we can provide getter for the clientset
```
var clientSet *ClientSet

func Get() *ClientSet {
    if clientSet == nil {
        clientSet = New()
    }
    return clientSet
}
```
WDT?

fedepaol: We risk races here I think? Don't remember if ginkgo tests are executed in parallel.
On the other hand, I wouldn't waste a mutex here.

Another option could be calling New() in the package's init() func and having the ClientSet variable exposed. 

something along the line of 
```
package client

var ClientSet ClientSet

func init() {
    // init here
}
```

On the calling site it would just be `client.ClientSet`.

cynepco3hahue: done

PR #21, opened 2019-12-24
Rename and extend performance API

cynepco3hahue:
The idea to have all performance related parameters in one place,
it does not mean, that you should specify any of them, all parameters will have
default value nil.

This PR also adds additional field to the `spec`, for now it should be look:

```yaml
apiVersion: performance.openshift.io/v1alpha1
kind: PerformanceProfile
metadata:
  name: example-performanceprofile
spec:
  cpu:
    reserved: "0-3"
    isolated: "4-5"
    nonIsolated: "0-1"
  hugepages:
    defaultHugepagesSize: "1G"
    pages:
    - size: "1G"
      count: 4
    - size: "2M"
      count: 1024
  nodeSelector:
    node-role.kubernetes.io/worker-rt: ""
  realTimeKernel:
    enabled: true
```


cynepco3hahue:
@yanirq Can you please review?

yanirq:
@cynepco3hahue did you drop some of the kargs we used ?

yanirq:
What about unitests ?

cynepco3hahue:
> @cynepco3hahue did you drop some of the kargs we used ?

I added only parameters that the user should change, all other I will add by default.

> What about unitests ?

It is only the API, I will add unit tests together with operator logic.
Currently I more interested to know if the looks good enough for us.

PR #24, opened 2019-12-25
Codify performance manifests

cynepco3hahue:
Codify performance manifests to make it easier to introduce complex logic for applying this manifest.

- [x] cluster feature gate resource
- [x] kubelet config resource
- [x] machine config resource
- [x] machine config pool
- [x] tuned resource
- [x] add unittest for all generation methods


build/assets/scripts/pre-boot-tuning.sh
```diff
@@ -0,0 +1,123 @@
...
+        ver=`awk '/version/ {print $2}' $line`
+        if [ "$ver" -gt "$current_ver" ]; then
+           current_ver=$ver
+           entry_file=$line
+        fi
+    done <<<$(egrep $(uname -r) -lr /boot/loader/entries/)
+
+    sed -i "s^initrd .*\$^& ${RHCOS_OSTREE_BOOTLOADER_PATH}iso_initrd.img^" $entry_file
+
+    #TODO - once RHCOS image contains the initrd content we can set parameters with rpm-ostree:
```
yanirq: We might actually be able to test that already with latest RHCOS changes , but ill do it in another PR

build/assets/tuned/openshift-node-network-latency
```diff
@@ -0,0 +1,52 @@
+[main]
+summary=Openshift node optimized for deterministic performance at the cost of increased power consumption, focused on low latency network performance
+include=openshift-node
+
+[cpu]
+# https://github.com/redhat-performance/tuned/blob/master/profiles/latency-performance/tuned.conf
+# https://github.com/redhat-performance/tuned/blob/master/profiles/network-latency/tuned.conf
+force_latency=cstate.id:1|3
```
yanirq: there might be an issue with this format of setting latency  (using the OR operand) , investigating this but worth checking if tuned pod doesn't have any errors with that here .

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	mc.Spec.Config = *ignitionConfig
+	mc.Spec.KernelArguments = getKernelArgs(profile.Spec.HugePages, profile.Spec.CPU.Isolated)
+
+	return mc, nil
+}
+
+func getKernelArgs(hugePages *performancev1alpha1.HugePages, isolatedCPUs *performancev1alpha1.CPUSet) []string {
+	kargs := []string{
```
yanirq: Can we pass these values in the CR as well ?  Can we and do we want to pass then in the CR ?

yanirq:
@cynepco3hahue  overall looks ok ,  If there are specific files that were not migrated as they are and are worth a second look please specify them.
Also need to see e2e tests pass

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
cynepco3hahue: Currently, I pass the whole performance profile.

build/assets/tuned/openshift-node-network-latency
cynepco3hahue: We can fix it under the separate PR

cynepco3hahue:
@yanirq e2e tests 

> @cynepco3hahue overall looks ok , If there are specific files that were not migrated as they are and are worth a second look please specify them.
> Also need to see e2e tests pass

Fixed unittest, e2e tests requires the deployment of the operator, will create following PR once #25  will be merged.

pkg/apis/performance/v1alpha1/performanceprofile_types.go
```diff
@@ -44,15 +44,15 @@ type HugePages struct {
...
-	Count *int32 `json:"count,omitempty"`
+	Count int32 `json:"count,omitempty"`
 }
 
 // RealTimeKernel defines the set of parameters relevant for the real time kernel.
 type RealTimeKernel struct {
-	// Enabled enables real time kernel on relevant nodes.
-	Enabled *bool `json:"enabled,omitempty"`
+	// RepoURL defines the URL to the repository with real time kernel packages
+	RepoURL *string `json:"repoURL,omitempty"`
```
davidvossel: is this going to eventually turn back into a boolean soon?

pkg/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -0,0 +1,40 @@
...
+  - match:
+    - label: label1
+      match:
+      - label: label2
+        value: label2
+      value: label1
+`
+
+var _ = Describe("Tuned", func() {
+	Context("with worker real time kerbnel profile", func() {
```
davidvossel: kerbnel/kernel

PR #25, opened 2019-12-31
Implement operator logic

cynepco3hahue:
The controller will watch resources:

- `PerformanceProfile`
- `MachineConfigPool`
- `MachineConfig`
- `KubeletConfig`
- `FeatureGate`
- `Tuned`

once the PerformanceProfile created the controller will start to reconcile
and will create all needed resources for performance tuning and it will continue
to save on the desired state of all components, the only way to update resources
owned by the controller is to update the PerformanceProfile resource.

I tried to make the controller less noisy as possible, that the reason why I provided custom predicates for all resources except `PerformanceProfile` it due to the fact that for all other resources we have openshift controllers that can update different fields of these resources.

An additional thing that mentions saying, I changed the scope of `PerformanceProfile` from namespaced to the cluster, because we can not set owner reference on cluster scope resources if the owner will have namespaced scope.

- [X] test creation of all components
- [x] test deletion of all components
- [x] provide unit tests
- [ ] integrate CI with the operator to deploy all components(I believe I will do it under additional PR)
- [ ] provide functional test(I believe I will do it under additional PR)


pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -140,54 +143,230 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+	if performanceProfile.Spec.CPU == nil {
+		return fmt.Errorf("you should provide CPU section")
+	}
+
+	if performanceProfile.Spec.CPU.Isolated == nil {
+		return fmt.Errorf("you should provide isolated CPU set")
+	}
+
+	if performanceProfile.Spec.CPU.NonIsolated == nil {
+		return fmt.Errorf("you should provide non isolated CPU set")
```
fromanirh: Why? I mean, why can't we just compute the non-isolated set by difference? Or do we want to do it in a future PR?

cmd/manager/main.go
```diff
@@ -66,6 +69,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	// we have two namespaces that we need to watch
+	// 1. tuned namespace
+	// 2. None namespace
+	namespaces := []string{
+		components.NamespaceNodeTuningOperator,
+		metav1.NamespaceNone,
```
fromanirh: Could you please add a one-line comment to remind our future selves what "None namespace" means in this context? (e.g. watch all the namespaces, even though IIRC this is not the case)

deploy/crds/performance.openshift.io_v1alpha1_performanceprofile_cr.yaml
```diff
@@ -2,6 +2,18 @@ apiVersion: performance.openshift.io/v1alpha1
 kind: PerformanceProfile
 metadata:
   name: example-performanceprofile
+  namespace: openshift-performance-addon
 spec:
-  # Add fields here
-  size: 3
+  cpu:
+    isolated: "2-3"
+    nonIsolated: "0"
```
fromanirh: Could you please clarify the state of CPU 1? how could it be neither `isolated` nor `nonIsolated`?

yanirq: maybe even add a validation hook on that later on ?

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: @fromanirh welcome back:)
Ok, in general, we can have 3 different workflows under the host:
- OS workflows
- some user application workflows(not the k8s one)
- containers workflows

and for each of these workflows, we can have the different CPU sets (one of the CPU set can be a subset of the other, but we can not precisely calculate values).

For example:
We set `reserved` to 0-3, it means these CPU's will not be used for container workflows, so these CPU's can be used or for OS workflows or for other user application that requires CPU isolation, and now we should separate CPU's that will be used for OS workflows and another user application, so we can now have CPU's 0-1 for OS workflows(`nonIsolated`), that leaves us with additional variable `isolated` that we can set to 2-5, it means that we have CPU's 2-3 for another user application that requires isolated CPU's and CPU's 4-5 for container workloads that requires isolated CPU's.

I hope it was clear enough, give me know if you need some additional information.

BTW, I do not say that we can not optimize it somehow, but for sure it will be part of another PR.

cmd/manager/main.go
cynepco3hahue: done

deploy/crds/performance.openshift.io_v1alpha1_performanceprofile_cr.yaml
cynepco3hahue: it is just example that will be injected under the CSV and does not the real case
regarding validation webhook I left `TODO` - https://github.com/openshift-kni/performance-addon-operators/blob/fd9054e83fef41af7f7683af9255a452dd7bd09e/pkg/controller/performanceprofile/performanceprofile_controller.go#L229

pkg/controller/performanceprofile/performanceprofile_controller.go
davidvossel: > and now we should separate CPU's that will be used for OS workflows and another user application, so we can now have CPU's 0-1 for OS workflows(nonIsolated), that leaves us with additional variable isolated that we can set to 2-5, it means that we have CPU's 2-3 for another user application that requires isolated CPU's and CPU's 4-5 for container workloads that requires isolated CPU's.

It seems like everything can be expressed with ```reserved``` and ```isolated```

example. 8 core machine with 1 reserved isolated core, 1 reserved non-isolated core, and 6 isolated cores for container workloads. 

```
reserved: "0-1"
isolated: "1-7"
```

If you wanted to leave a general cores for container workloads as well, you could do something like this.

```
reserved: "0-1"
isolated: "1-5"
```

That would be 1 isolated core for reserved, 1 non-isolated core for reserved, 4 isolated cores for container workloads, and 2 non-isolated cores for container workloads. 

then non-isolated would be assumed to be "0,6-7" 

would that optimization make sense?

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +208,225 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
-					Command: []string{"sleep", "3600"},
-				},
-			},
-		},
+	if err := r.createOrUpdateMachineConfigPool(mcp); err != nil {
+		return err
+	}
+
+	// pause machine config pool
+	if err := r.pauseMachineConfigPool(mcp.Name, true); err != nil {
```
davidvossel: we don't want to issu APIs call every time the reconcile loop pops unless something changes.  We'll need to check to see if modifications are going to occur and only pause/unpause as modifications happen.

Also, this pause/unpause logic seems like it could have a race condition. For example, what guarantees us that the machine config operator sees that we paused a mcp before seeing a new machine config is posted?

deploy/crds/performance.openshift.io_v1alpha1_performanceprofile_cr.yaml
fromanirh: Ack, good enough for me now

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: At first glance, it looks fine. I will give it another thought and will reflect it under the PR :)

deploy/service_account.yaml
```diff
@@ -1,4 +0,0 @@
-apiVersion: v1
```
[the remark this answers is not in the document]
cynepco3hahue: technically it required only for manual deployment(not via OLM), CSV generator does not use service account resources at all. Give me know if you prefer to have it.

hack/clean-deploy.sh
```diff
@@ -32,3 +32,5 @@ spec:
   publisher: Red Hat
   sourceType: grpc
 EOF
+
+$OC_TOOL -n openshift-performance-addon delete csv --all
```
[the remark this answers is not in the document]
cynepco3hahue: done

pkg/controller/performanceprofile/components/consts.go
```diff
@@ -26,5 +28,7 @@ const (
 
 const (
 	// FeatureGateLatencySensetiveName defines the latency sensetive feature gate name
-	FeatureGateLatencySensetiveName = "latency-sensetive"
-)
\ No newline at end of file
+	// TOOD: uncomment once https://bugzilla.redhat.com/show_bug.cgi?id=1788061 fixed
+	// FeatureGateLatencySensetiveName = "latency-sensetive"
+	FeatureGateLatencySensetiveName = "cluster"
```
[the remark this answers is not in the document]
cynepco3hahue: it does not feature set, it the name of the feature gate resource.
The initial idea was to create additional feature gate resource that should enable topology manager(that will be managed by us), but because of the bug, the kubelet controller does not render correct machine config(it uses only default `cluster` feature gate resource)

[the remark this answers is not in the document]
cynepco3hahue: fixed typo

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -54,12 +68,103 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			}
+			if e.MetaNew == nil {
+				klog.Error("Update event has no new metadata")
+				return false
+			}
+
+			mcpNew := e.ObjectOld.(*mcov1.MachineConfigPool)
+			mcpOld := e.ObjectNew.(*mcov1.MachineConfigPool)
+
+			mcpNew.Spec.Paused = mcpOld.Spec.Paused
```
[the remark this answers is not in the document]
cynepco3hahue: will add deep copy just to be sure:)

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +208,225 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+			return reconcile.Result{}, nil
+		}
+	}
+
+	// TODO: we need to check if all under performance profiles values != nil
+	// first we need to decide if each of values required and we should move the check into validation webhook
+	// for now let's assume that all parameters needed for assets scrips are required
+	if err := r.verifyPerformanceProfileParameters(instance); err != nil {
+		// we do not want to reconcile again in case of error, because a user will need to update the PerformanceProfile anyway
+		klog.Errorf("failed to reconcile: %v", err)
```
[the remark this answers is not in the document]
cynepco3hahue: Can we leave events and status sections for following PR's, this PR already too big?
Created:
- https://github.com/openshift-kni/performance-addon-operators/issues/29
- https://github.com/openshift-kni/performance-addon-operators/issues/30

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: What do you mean by modifications? Modifications to performance profile resource? We anyway run a very small number of reconciling loops because of our predicates(only when spec or labels was updated, so we really pause only on important modifications).

Regarding the race condition, you are right once we will have `MaxConcurrentReconciles` bigger than one we will have troubles, but to be honest I do not have a good solution for it.
I can lock with the mutex `applyComponents` method, but it the same as running reconcile only with the one thread, and you can not really know if it is an additional update in progress.
For now, I will add a lock, but maybe do you have better ideas?

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +208,225 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
-	// Define a new Pod object
-	pod := newPodForCR(instance)
+	if instance.DeletionTimestamp != nil {
+		// delete components
+		if err := r.deleteComponents(instance); err != nil {
```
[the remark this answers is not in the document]
cynepco3hahue: mm I do not remember that under some operator we waited for the real deletion, but if you insist I can add it.

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: > If you wanted to leave a general cores for container workloads as well, you could do something like this.
> 
> ```
> reserved: "0-1"
> isolated: "1-5"
> ```
> 
> That would be 1 isolated core for reserved, 1 non-isolated core for reserved, 4 isolated cores for container workloads, and 2 non-isolated cores for container workloads.
> 
> then non-isolated would be assumed to be "0,6-7"

On the second though, non-isolated in our context means that these CPUs will hardly be used by the operating system, so it does not good to use them for container workloads, so you still need the non-isolated parameter to specify what part of 0,6-7 will be used for OS workflows.

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +208,225 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
-	// Check if this Pod already exists
-	found := &corev1.Pod{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, found)
-	if err != nil && errors.IsNotFound(err) {
-		reqLogger.Info("Creating a new Pod", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
-		err = r.client.Create(context.TODO(), pod)
-		if err != nil {
+	// add finalizer
+	if !hasFinalizer(instance, finalizer) {
+		instance.Finalizers = append(instance.Finalizers, finalizer)
```
slintes: To my best knowledge it's better to update a copy of the instance, for the same reason David mentioned at another place (updating a chached instance from informer is bad)?

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +208,225 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+		if err := r.client.Update(context.TODO(), instance); err != nil {
 			return reconcile.Result{}, err
 		}
 
-		// Pod created successfully - don't requeue
+		// we exit reconcile loop because we will have additional update reconcile
```
slintes: it's not clear to me why we exit here and not go on with processing this instance?

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +208,225 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+	// TODO: we need to update the status
+
 	return reconcile.Result{}, nil
 }
 
-// newPodForCR returns a busybox pod with the same name/namespace as the cr
-func newPodForCR(cr *performancev1alpha1.PerformanceProfile) *corev1.Pod {
-	labels := map[string]string{
-		"app": cr.Name,
+func (r *ReconcilePerformanceProfile) applyComponents(profle *performancev1alpha1.PerformanceProfile) error {
```
slintes: typo in arg name `profile`

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +208,225 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+	// deploy machine config pool
+	mcp := machineconfigpool.NewPerformance(profle)
```
slintes: maybe `NewPerformancePool` is a better name?

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +208,225 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+		return err
+	}
+
+	// pause machine config pool
+	if err := r.pauseMachineConfigPool(mcp.Name, true); err != nil {
+		return err
+	}
+
+	// deploy machine config
+	mc, err := machineconfig.NewPerformance(r.assetsDir, profle)
```
slintes: `NewPerformanceConfig`?

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +208,225 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+	if err := controllerutil.SetControllerReference(profle, mc, r.scheme); err != nil {
+		return err
+	}
+
+	if err := r.createOrUpdateMachineConfig(mc); err != nil {
+		return err
+	}
+
+	// deploy kubelet config
+	kc := kubeletconfig.NewPerformance(profle)
```
slintes: `NewPerformanceConfig`?

slintes: same for following NewXyz() funcs...

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +208,225 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+	updatedMcp, err := r.getMachineConfigPool(mcpName)
+	if errors.IsNotFound(err) {
+		klog.Warning("failed to set pause, the machine config pool does not exist, probably it was deleted")
+		return nil
+	}
+
+	if err != nil {
+		return err
+	}
+	if err != nil {
```
slintes: duplicate :)

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +208,225 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+		return err
+	}
+
+	updatedMcp.Spec.Paused = pause
+	return r.client.Update(context.TODO(), updatedMcp)
+}
+
+func (r *ReconcilePerformanceProfile) verifyPerformanceProfileParameters(performanceProfile *performancev1alpha1.PerformanceProfile) error {
```
slintes: nit: not sure what the exact difference between the 2 words is, but mostly we talk about `validate` instead of `verify` (validation webhook) :thinking:

hack/clean-deploy.sh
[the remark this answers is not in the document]
cynepco3hahue: Yes, I waited for your review to commit all together:)

cmd/manager/main.go
```diff
@@ -38,53 +43,50 @@ var (
...
-	// be propagated through the whole operator, generating
-	// uniform and structured logs.
-	logf.SetLogger(zap.Logger())
-
 	printVersion()
 
 	namespace, err := k8sutil.GetWatchNamespace()
 	if err != nil {
-		log.Error(err, "Failed to get watch namespace")
+		klog.Error(err.Error())
```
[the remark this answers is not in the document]
cynepco3hahue: done

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -54,12 +68,103 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	// TODO(user): Modify this to be the types you create that are owned by the primary resource
-	// Watch for changes to secondary resource Pods and requeue the owner PerformanceProfile
+	// we want to initate reconcile loop only on change under labels or spec of the object
+	p := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.MetaOld == nil {
```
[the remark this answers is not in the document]
cynepco3hahue: I unsure when we will get this, but the default update predicate has same checks so I left it as it is

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -54,12 +68,103 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	// Watch for changes to tuned owned by our controller
+	err = c.Watch(&source.Kind{Type: &tunedv1.Tuned{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &performancev1alpha1.PerformanceProfile{},
+	}, p)
+	if err != nil {
+		return err
+	}
+
+	// we do not want initiate reconcile loop on the pause of machine config pool
```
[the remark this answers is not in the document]
cynepco3hahue: done

deploy/crds/performance.openshift.io_v1alpha1_performanceprofile_cr.yaml
cynepco3hahue: @slintes this CR example injected into CSV, so I do not add some comments to it(unsure regarding parsing), what example do you prefer?

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -54,12 +68,103 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			if e.ObjectNew == nil {
+				klog.Error("Update event has no new runtime object for update")
+				return false
+			}
+			if e.MetaNew == nil {
+				klog.Error("Update event has no new metadata")
+				return false
+			}
+
+			mcpNew := e.ObjectOld.(*mcov1.MachineConfigPool)
```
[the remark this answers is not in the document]
cynepco3hahue: it is a mistake, fixed

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: I unsure that we modify cached instance, we create a new one and fill it with the details from the cached instance
```
instance := &performancev1alpha1.PerformanceProfile{}
err := r.client.Get(context.TODO(), request.NamespacedName, instance)
```

I believe the main reason to copy, it when you want to change something under the instance for the update, but still want to continue reconcile loop and work on the instance without update

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: the only reason it because, once we will have status section, we will need to update the status section as well, but each update will raise resource version, and so the second update will fail because of different resource version

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: done

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: we already have `machineconfigpool` in the name of the package, see a nice link that @fedepaol sent me - https://blog.golang.org/package-names

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: see the link above

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: see the link above

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: done

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: changed

cynepco3hahue:
@davidvossel @slintes A lot of thanks for the review!

pkg/controller/performanceprofile/performanceprofile_controller.go
davidvossel: >  so you still need the non-isolated parameter to specify what part of 0,6-7 will be used for OS workflows.

reserved is 0-1. so 0 would be used for OS and 6-7 for containers.  What am i missing here?

deploy/service_account.yaml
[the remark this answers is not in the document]
davidvossel: ah, okay. I actually like that these manifests are being removed then if we aren't using them directly. It forces everyone down the CSV path, which is what we want at the moment.

pkg/controller/performanceprofile/components/consts.go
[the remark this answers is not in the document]
davidvossel: > but because of the bug, the kubelet controller does not render correct machine config(it uses only default cluster feature gate resource)

oh, so does this mean we can't enable topology manager until that BZ is resolved?

pkg/controller/performanceprofile/performanceprofile_controller.go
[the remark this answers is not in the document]
davidvossel: > mm I do not remember that under some operator we waited for the real deletion, but if you insist I can add it.

Waiting for deletion guarantees us that we have 100% successfully cleaned up all dependent resources before the parent resource is completely removed. Without this, sometimes issues during tear down are masked. CI wouldn't necessarily catch this because the parent resource was deleted while dependent resources might still remain in a pending deletion type phase.

so basically, it ensures our teardown logic works. otherwise these issues go un-noticed and only manifest themselves in odd situations where people attempt to delete a resource and then re-post it again.

pkg/controller/performanceprofile/performanceprofile_controller.go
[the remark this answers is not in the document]
davidvossel: sounds good :+1:

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +212,253 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+	if errors.IsNotFound(err) {
+		klog.Warning("failed to set pause, the machine config pool does not exist, probably it was deleted")
+		return nil
+	}
+
+	if err != nil {
+		return err
+	}
+
+	updatedMcp.Spec.Paused = pause
```
davidvossel: only perform the update here if updatedMcp.Spec.Paused does not already match the desired pause value.

pkg/controller/performanceprofile/performanceprofile_controller.go
davidvossel: > What do you mean by modifications? Modifications to performance profile resource? We anyway run a very small number of reconciling loops because of our predicates(only when spec or labels was updated, so we really pause only on important modifications).

What I'm suggesting is to determine up front whether any config changes will occur that involve a restart, and only perform the "pause/unpause" logic when it's detected that changes will occur.

so the flow would be something like this.

* look though cache to determine if any machine configs, kubelet configs, etc... either need to be posted or updated.
    * If not, fast succeed and move on to status calculations since there's nothing to be done
    * else pause, apply changes, unpause then move on to status calculations.

I know what I'm suggesting is kind of a pain, the issue is if we get into an error loop or some other unexpected reconcile loop spin, things like this mean we spam the api server with writes.

In practice we need to only GET or POST to the api server when it's absolutely necessary. This is why we do things like only posting an update to the status section when ```if ! reflect.DeepEquals(newStatus, oldStatus)``` We're avoiding a post to the API server that would otherwise occur every reconcile loop if we didn't determine that the change wasn't necessary by looking at cache.

davidvossel: > Regarding the race condition, you are right once we will have MaxConcurrentReconciles bigger than one we will have troubles, but to be honest I do not have a good solution for it.
I can lock with the mutex applyComponents method, but it the same as running reconcile only with the one thread, and you can not really know if it is an additional update in progress.
For now, I will add a lock, but maybe do you have better ideas?

And with regards to the race condition.

What I was referring to isn't related to our reconcile loop. Regardless of how many concurrent reconcile loops we allow, we are guaranteed a single key is only being acted on by a single reconcile execution. There's no parallel execution of the same key, only parallel execution of separate keys. 

What I was referring to is actually the machine config operator's reconcile loop. Is there anything guaranteeing us that the operator will be notified that we paused a pool before it's notified we posted new machine configs or kubelet configs?

Basically, do we have any guarantees that the order we post changes in our reconcile loop matches the order another operator will receive them in? so, If I post resource A then resource B, is it possible for another operator to observe resource B's creation before resource A? if resources A and B are of different kinds, i think this might actually be possible.

pkg/controller/performanceprofile/components/consts.go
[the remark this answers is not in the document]
cynepco3hahue: nope, it just means that we should update the default feature gate resource `cluster` instead of creating a new one(that can create some conflicts with user configuration, so I wanted to avoid it), also it should be updated before the kubelet config creation, see - https://bugzilla.redhat.com/show_bug.cgi?id=1788061#c3

pkg/controller/performanceprofile/performanceprofile_controller.go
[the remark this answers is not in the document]
cynepco3hahue: added

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: > so the flow would be something like this.
> 
> * look though cache to determine if any machine configs, kubelet configs, etc... either need to be posted or updated.
>   
>   * If not, fast succeed and move on to status calculations since there's nothing to be done
>   * else pause, apply changes, unpause then move on to status calculations.

@davidvossel But we already have this functionality because of custom update predicates, it just does not enter to the reconcile loop at all, if it wasn't important update.

cynepco3hahue: > And with regards to the race condition.
> 
> What I was referring to isn't related to our reconcile loop. Regardless of how many concurrent reconcile loops we allow, we are guaranteed a single key is only being acted on by a single reconcile execution. There's no parallel execution of the same key, only parallel execution of separate keys.
> 
> What I was referring to is actually the machine config operator's reconcile loop. Is there anything guaranteeing us that the operator will be notified that we paused a pool before it's notified we posted new machine configs or kubelet configs?
> 
> Basically, do we have any guarantees that the order we post changes in our reconcile loop matches the order another operator will receive them in? so, If I post resource A then resource B, is it possible for another operator to observe resource B's creation before resource A? if resources A and B are of different kinds, i think this might actually be possible.

Ah, I see, good point, but I unsure how we can avoid it, I can divide reconcile loop, something like:
- create or update MCP and pause it(`return reconcile.Result{Requeue: true, RequeueAfter: time.Second}`)
- before creating all other components, check that MCP paused, if return again `reconcile.Result{Requeue: true, RequeueAfter: 10*time.Second}`
- create all components
- unpause MCP
@davidvossel WDT?

pkg/controller/performanceprofile/performanceprofile_controller.go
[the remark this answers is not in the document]
slintes: ack, works for me

pkg/controller/performanceprofile/performanceprofile_controller.go
slintes: ah, I see, interesting read about package names and `New()`. Thanks!
But `NewPerformance` still sounds a bit wrong to me... :thinking: 
Maybe go into the other direction and make the func name even shorter, only `New()`?
Or sth different than New, e.g. `FromProfile()`?
WDYT?

pkg/controller/performanceprofile/performanceprofile_controller.go
slintes: > we create a new one
oh yes, sure, sorry, missed that

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: Only `New()` works for me, will change it

pkg/controller/performanceprofile/performanceprofile_controller.go
davidvossel: Looks good, the only thing I'd add is to only unpause after you observe all dependent resources have been created.

pkg/controller/performanceprofile/resources.go
```diff
@@ -0,0 +1,350 @@
...
+
+	mutated := existing.DeepCopy()
+
+	mergeMaps(mcp.Annotations, mutated.Annotations)
+	mergeMaps(mcp.Labels, mutated.Labels)
+
+	mutated.Spec = mcp.Spec
+
+	// do not update the pause and configuration fields
+	mutated.Spec.Paused = existing.Spec.Paused
```
davidvossel: I think we should remove this now. it shouldn't matter anymore. Reconciling based on whether the mcp is paused or not shouldn't impact us now with the new logic.

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +210,350 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+		return nil, err
+	}
+
+	// we want to give time to the machine-config controllers to get updated values
+	if updated {
+		return &reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+	}
+
+	// update the machine config pool after that we paused it
+	if mcpMutated != nil && !created {
```
davidvossel: lets only update the mcp object once.  If the mcp needs to be updated to be paused or updated for any other reason, that should be a single update.

writing to the mcp object to pause it, and then writing to it once again right afterwards for some other reason doesn't look right. Figure out what the state of the mcp object should be, and issue a single create or update.

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +210,350 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
-		Spec: corev1.PodSpec{
-			Containers: []corev1.Container{
-				{
-					Name:    "busybox",
-					Image:   "busybox",
-					Command: []string{"sleep", "3600"},
-				},
-			},
-		},
+	mcpMutated, created, err := r.getMutatedMachineConfigPool(mcp)
```
davidvossel: the ```created``` value here is confusing.  created=false when mcp exists. and created=true when mcp needs to be created.

can we call ```created``` something like ```nonCreated```, or reverse the meaning of ```created``` to reflect the status of the object being observed?

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +210,350 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+		}
+
+		if updated {
+			return &reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+		}
+		return nil, nil
+	}
+
+	// create machine config pool and pause it
+	if mcpMutated != nil && created {
```
davidvossel: if we know the mcp needs to be created here, and we're just going to immediately pause it after creation, then why not post the manifest with paused=true to begin with?

so

```
if mcpMutated != nil && needsToBeCreated {
// post a new paused mcp. 
} else {
// get and pause existing mcp
}

```

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +210,350 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+
+	// does not update any resources, if it no changes to relevant objects and just continue to the status update
+	if mcpMutated == nil &&
+		mcMutated == nil &&
+		kcMutated == nil &&
+		fgMutated == nil &&
+		networkLatencyTunedMutated == nil &&
+		realTimeKernelTunedMutated == nil {
+		// if nothing needed to be updated and machine config pool paused, unpause it
+		mcpUpdate, err := r.getMachineConfigPool(mcp.Name)
```
davidvossel: the mcp object has references to objects in the status section.

the ```mcp.status.configuration.source``` list. Can we look at that to make sure the mcp picked up all our configs before unpausing it?

pkg/controller/performanceprofile/performanceprofile_controller.go
davidvossel: do we have any more thoughts here on removing non-isolated?

slintes: your suggestion makes sense to me on a first look, but I'm not familiar enough with isolated / reserved CPUs to have a strong opinion here...

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +210,350 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+
+	if performanceProfile.Spec.CPU.Isolated == nil {
+		return fmt.Errorf("you should provide isolated CPU set")
+	}
+
+	if performanceProfile.Spec.CPU.NonIsolated == nil {
+		return fmt.Errorf("you should provide non isolated CPU set")
+	}
+
+	if performanceProfile.Spec.RealTimeKernel == nil {
```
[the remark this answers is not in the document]
slintes: @cynepco3hahue @davidvossel heads up, I am already working on a follow up PR for making this a boolean, because with OCP 4.4 we don't need the repo url anymore at all, the RPMs are in the RHCOS image already. Also I remove validation for this and just enable the RT kernel systemd unit when the boolean is set and true.

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +210,350 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+	if err := controllerutil.SetControllerReference(profile, networkLatencyTuned, r.scheme); err != nil {
+		return nil, err
+	}
+	networkLatencyTunedMutated, err := r.getMutatedTuned(networkLatencyTuned)
+	if err != nil {
+		return nil, err
+	}
+
+	// get mutated real time kernel tuned
+	realTimeKernelTuned, err := tuned.NewWorkerRealTimeKernel(r.assetsDir, profile)
```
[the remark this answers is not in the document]
fromanirh: Filed issue: https://github.com/openshift-kni/performance-addon-operators/issues/32

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: ok so our assumption that nonIsolatedCpu's should be a subset of `reserved` CPU's and does not have an intersection with `isolated` Cpu's, I believe it will be correct in most cases, still, I feel unsure regarding this kind of calculations for all use cases
@fromanirh @vladikr Guys, do you have some thoughts?

@davidvossel Can we leave this discussion for another PR, to avoid unmerge of this one?

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: Hm, we unpause just when all mutated objects equal to nil https://github.com/openshift-kni/performance-addon-operators/blob/c3dd292b406b8ae95b4047ea8c5f12661c159b64/pkg/controller/performanceprofile/performanceprofile_controller.go#L366, it means that the object that we get from the cache and the object that we expect to have it the same, so it means that all objects already exist, or do I miss something?

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: done

pkg/controller/performanceprofile/resources.go
cynepco3hahue: I am unsure it correct. for example, our code will pause the machine config pool, but the MCP that we will generate will have spec.paused false, so it will recognize the change and will unpause.

pkg/controller/performanceprofile/performanceprofile_controller.go
fromanirh: My take: first, let's move this (important) discussion to another issue/PR to let this PR move forward.

Second: `isolated` and `reserved` *in the current implementation* must not overlap. So the question boils down if we actually need cpus which are neiter isolated or reserved, which form a shared pool of cores on which container workload can run.
After a bit more thought I agree that non-reserved, non-isolated cores is a worthy concept, but we should present it nicely.

I think the takeaway is this little topic deserves more attention and perhaps a better abstraction, which is another hint we should move this discussion to a separate ticket.

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: done

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: After merging `pause` and `createOrUpdate` of MCP I dropped it, I just overthought it X_X

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: done

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: mm feature gate and kubelet config machine configs rendered always with the same name, so I unsure how can we check it, see - https://github.com/openshift/machine-config-operator/blob/04cd2198cae247fabcd3154669618d74f124f27f/pkg/controller/kubelet-config/helpers.go#L60

pkg/controller/performanceprofile/performanceprofile_controller.go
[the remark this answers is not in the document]
cynepco3hahue: @davidvossel  According to Marc comment, can we leave it for following PR, because it will need change under scripts and API?

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +210,350 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+func (r *ReconcilePerformanceProfile) pauseMachineConfigPool(mcp *mcov1.MachineConfigPool, pause bool) (bool, error) {
+	if mcp.Spec.Paused == pause {
+		return false, nil
+	}
+
+	mcp.Spec.Paused = pause
+	return true, r.client.Update(context.TODO(), mcp)
+}
+
+func (r *ReconcilePerformanceProfile) validatePerformanceProfileParameters(performanceProfile *performancev1alpha1.PerformanceProfile) error {
```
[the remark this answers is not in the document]
cynepco3hahue: I added all parameters that needed for scripts as required, otherwise, we can have some unpredicted behavior under our scripts. To change it we will need to change our scripts, so I prefer to leave it for another PR, opened https://github.com/openshift-kni/performance-addon-operators/issues/33

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +210,350 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+		}
+
+		// delete components
+		if err := r.deleteComponents(instance); err != nil {
+			klog.Errorf("failed to delete components: %v", err)
+			return reconcile.Result{}, err
+		}
+
+		if r.isComponentsExist(instance) {
+			return reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
```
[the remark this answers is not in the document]
cynepco3hahue: done

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +210,350 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+			return nil, err
+		}
+
+		updated, err := r.pauseMachineConfigPool(mcpUpdate, false)
+		if err != nil {
+			return nil, err
+		}
+
+		if updated {
+			return &reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
```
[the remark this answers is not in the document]
cynepco3hahue: hm wasn't aware of it, but I see under the controller code
```
} else if result.RequeueAfter > 0 {
		// The result.RequeueAfter request will be lost, if it is returned
		// along with a non-nil error. But this is intended as
		// We need to drive to stable reconcile loops before queuing due
		// to result.RequestAfter
		c.Queue.Forget(obj)
		c.Queue.AddAfter(req, result.RequeueAfter)
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, "requeue_after").Inc()
		return true
	}
```
Will remove it.

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -103,54 +210,350 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+	}
+
+	updated, err := r.pauseMachineConfigPool(mcpUpdate, true)
+	if err != nil {
+		return nil, err
+	}
+
+	// we want to give time to the machine-config controllers to get updated values
+	if updated {
+		return &reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
```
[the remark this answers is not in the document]
cynepco3hahue: done

pkg/controller/performanceprofile/performanceprofile_controller.go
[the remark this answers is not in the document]
slintes: it's also in the doc of the `Result` struct ;)

pkg/controller/performanceprofile/performanceprofile_controller.go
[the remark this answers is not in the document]
davidvossel: alrighty, as long as we're tracking this, thanks

pkg/controller/performanceprofile/performanceprofile_controller.go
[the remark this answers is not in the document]
davidvossel: ack, @slintes thanks for following up on this.

pkg/controller/performanceprofile/performanceprofile_controller.go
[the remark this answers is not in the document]
davidvossel: perfect, thanks

pkg/controller/performanceprofile/performanceprofile_controller.go
davidvossel: so, we'd only be able to detect the new MC was applied?

here's an issue  #35 lets follow up on this discussion and see if there's a way to detect when unpause is safe. or at least detect the mc we posted has been picked up before unpausing.

pkg/controller/performanceprofile/resources.go
davidvossel: that's a bug. the mcp we generate needs to reflect whether or not the pause is expected or not. 

we can't prevent the reconcile loop from popping early, this logic sounds like we're depending on the reconcile loop not executing too early which would result in a early unpause.

PR #26, opened 2020-01-02
Add ptp feature using kustomize

SchSeba:
This PR add the ptp feature using kustomize.

Signed-off-by: Sebastian Sch <sebassch@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/openshift-kni/performance-addon-operators/26)

### DIFF
--- a/cluster-setup/ci-cluster/feature_ptp/kustomization.yaml
+++ b/cluster-setup/ci-cluster/feature_ptp/kustomization.yaml
@@ -1,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+bases:
+  - ../../base/feature_ptp
+patches:
+  - subscription.yaml
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 
-export FEATURES?=mcp performance sctp
+export FEATURES?=mcp performance sctp ptp
 
 .PHONY: build \
 	deps-update \
--- a/cluster-setup/base/feature_ptp/01_namespace.yaml
+++ b/cluster-setup/base/feature_ptp/01_namespace.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: v1
+kind: Namespace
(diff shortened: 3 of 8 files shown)


cluster-setup/ci-cluster/feature_ptp/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+bases:
+  - ../../base/feature_ptp
+patches:
+  - subscription.yaml
```
fedepaol: Why this patch? The target should be 4.4

fedepaol:
Also, if the ptp operator requires some configuration (such as node labelling, iirc) this setup should be here.

SchSeba:
Hi @fedepaol 

> Also, if the ptp operator requires some configuration (such as node labelling, iirc) this setup should be here.

I will like to do the label and everything related to that in the functional test itself it will give us better control and leave the deploy to using only kustomize

fedepaol:
> Hi @fedepaol
> 
> > Also, if the ptp operator requires some configuration (such as node labelling, iirc) this setup should be here.
> 
> I will like to do the label and everything related to that in the functional test itself it will give us better control and leave the deploy to using only kustomize

Sounds good to me.

cluster-setup/ci-cluster/feature_ptp/kustomization.yaml
SchSeba: here we use the test source.

Production: `redhat-operators`
Test: `opsrctest` the should point to the ART source catalog

slintes:
with the new cnf deploy repo being available, I guess we do not want to add more features to this repo, but only test the operator itself?
@davidvossel ?

slintes:
just found the confirmation of my assumption over here: https://github.com/openshift-kni/performance-addon-operators/pull/28#issuecomment-572595831

/close

PR #28, opened 2020-01-05
Add sriov feature deployment using kustomize

SchSeba:
This PR add the SRIOV features.

There is an open question here that I will like to discuss on this PR.
There are some configuration that are specific to the hardware the sriov operator is running on like number of virtual functions and the PCI addresses of the network cards

Example:
```yaml
apiVersion: sriovnetwork.openshift.io/v1
kind: SriovNetworkNodePolicy
metadata:
  name: sriov-network-node-policy
  namespace: openshift-sriov-network-operator
spec:
  deviceType: netdevice
  isRdma: false
  mtu: 1500
  nicSelector:
    pfNames:
    - eno1
    rootDevices:
    - "0000:01:00.0"
  nodeSelector:
    feature.node.kubernetes.io/network-sriov.capable: true
  numVfs: 5
  priority: 10
  resourceName: sriovnic
```

I wasn't able to find a way to do it using kustomize this is why this PR should be on hold until we agree on a solution.

/hold

Signed-off-by: Sebastian Sch <sebassch@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/openshift-kni/performance-addon-operators/28)

### DIFF
--- a/cluster-setup/ci-cluster/feature_sriov/10-sriov-daemonset.yaml
+++ b/cluster-setup/ci-cluster/feature_sriov/10-sriov-daemonset.yaml
@@ -1,0 +1,66 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: sriov-unsupported-nics-entrypoint
+data:
+  entrypoint.sh: |
+    #!/bin/sh
+    set -euo pipefail
+    echo "########################################"
+    echo "# Apply udev rules in unsupported NICs #"
+    echo "########################################"
+    OPERATORUDEVRULES="/etc/udev/rules.d/10-nm-unmanaged.rules"
+    CUSTOMUDEVRULES="/etc/udev/rules.d/11-custom-nm-unmanaged.rules"
+
+    if [ -f ${CUSTOMUDEVRULES} ] || [ -f ${OPERATORUDEVRULES} ]; then
+      while true ; do
+        echo "udev rules already applied, sleep loop"
+        sleep 3600
+      done
+    else
+      UDEVDEVICE=$(cat /sys/class/net/${nic}/device/sriov_vf_device)
+      echo "ACTION==\"add|change\", ATTRS{device}==\"0x${UDEVDEVICE}\", ENV{NM_UNMANAGED}=\"1\"" > ${CUSTOMUDEVRULES}
+      udevadm control --reload-rules 
+      udevadm trigger
+    fi
+---
+
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: sriov-unsupported-nics-ds
+  labels:
+    app: sriov-unsupported-nics-ds
+spec:
+  selector:
+    matchLabels:
+      app: sriov-unsupported-nics-ds
+  template:
+    metadata:
+      labels:
+        app: sriov-unsupported-nics-ds
+    spec:
+      serviceAccountName: sriov-network-config-daemon
+      nodeSelector:
+        feature.node.kubernetes.io/network-sriov.capable: true
+      containers:
+      - name: sriov-unsupported-nics
+        image: ubi8/ubi-minimal
+        command: ['sh', '-c', 'cp /script/entrypoint.sh /host/tmp && chmod +x /host/tmp/entrypoint.sh && echo "Creating udev rules" && chroot /host /tmp/entrypoint.sh && sleep infinity']
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - mountPath: /script
+          name: sriov-unsupported-nics-script
+        - mountPath: /host
+          name: host
+      restartPolicy: Always
+      terminationGracePeriodSeconds: 10
+      volumes:
+      - configMap:
+          name: sriov-unsupported-nics-entrypoint
+        name: sriov-unsupported-nics-script
+      - hostPath:
+          path: /
+          type: Directory
+        name: host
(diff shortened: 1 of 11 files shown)


fedepaol:
The `cluster-setup` folder is meant to solve two different purpouses:
- provide an example of how is it possible to deploy sriov with manifests and kustomize
- provide an entry point to prepare a working cluster to run the sr-iov tests.

This is not intended to be the final solution that a customer will use becase, as you said, there are a lot of variables that depend on how the end user will configure his environment. In sctp for example, I just "kustomize" the role of the node we want to apply the config.

As such you can, for example, kustomize only the nic vendor in the  `SriovNetworkNodePolicy` manifest (which should be enough to select a SR-IOV nic, removing  `rootDevices` and `pfNames`).
This could also help once we have a CI cluster we may want to run this against, since we could ideally have different vendors.

The "cluster preparation" part is performed by https://github.com/openshift-kni/performance-addon-operators/blob/master/hack/setup_test_cluster.sh . Please note that it expects a `wait_for_it.sh` script to be notified that the feature is ready to be tested.

This is important in order to be sure the cluster is ready to run our tests.

cluster-setup/ci-cluster/feature_sriov/sriov-subscription.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: sriov-network-operator-subscription
+  namespace: openshift-sriov-network-operator
+spec:
+  channel: "4.3"
```
fedepaol: This file should be removed in favour of https://github.com/openshift-kni/performance-addon-operators/pull/28/files#diff-b4137bb24933e2f96ff56df5649dcc80 I think?

cluster-setup/ci-cluster/feature_sriov/10-sriov-daemonset.yaml
```diff
@@ -0,0 +1,66 @@
+kind: ConfigMap
```
fedepaol: Please add a comment that says that this is a workaround for nics not officially supported by the operator.
In theory, if we manage to get a CI cluster with a supported nic, this could be removed because the operator will put the udev rule for us.

Also, if this doesn't end kustomized should be in base

cluster-setup/ci-cluster/feature_sriov/11-sriov-networknodepolicy.yaml
```diff
@@ -0,0 +1,19 @@
...
+kind: SriovNetworkNodePolicy
+metadata:
+  name: sriov-network-node-policy
+  namespace: openshift-sriov-network-operator
+spec:
+  deviceType: netdevice
+  isRdma: false
+  mtu: 1500
+  nicSelector:
+    pfNames:
```
fedepaol: As I wrote in the comment, I would put this under base and then kustomize the nic vendor.

cluster-setup/ci-cluster/feature_sriov/12-sriov-network.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: sriovnetwork.openshift.io/v1
```
fedepaol: This should go in base

cluster-setup/ci-cluster/feature_sriov/20-sriov-testing-namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
fedepaol: This should belong to the test setup (once we have it).
Ideally, the tests find the cluster with the sr-iov feature ready to be used and runs its tests, setting up the namespace and launching the pods, verifying what has to be verified.

cluster-setup/ci-cluster/feature_sriov/21-sriov-testing-deployment.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apps/v1
```
fedepaol: Same as above, the test pods should be created by the gingko tests.

cluster-setup/ci-cluster/feature_sriov/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+bases:
+  - ../../base/feature_sriov
+patches:
+  - sriov-subscription.yaml
+resources:
+  - 10-sriov-daemonset.yaml
```
fedepaol: Here we should have only the patch

cluster-setup/ci-cluster/feature_sriov/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+bases:
+  - ../../base/feature_sriov
+patches:
+  - sriov-subscription.yaml
```
fedepaol: Why this patch?

cluster-setup/ci-cluster/feature_sriov/11-sriov-networknodepolicy.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - eno1
+    rootDevices:
+    - "0000:01:00.0"
```
mmirecki: This should be taken from SriovNetworkNodeState.

davidvossel:
fyi, these kustomize scripts are moving to a centralized repo outside of performance addon operators. more info coming shortly.

slintes:
@SchSeba can you move this to https://github.com/openshift-kni/cnf-features-deploy please? Thanks!

/close

PR #31, opened 2020-01-08
Force latency is not supported with cstate in tuned 2.11

yanirq:
force_latency parameter in tuned profile is supported only for versions greater than 2.11. 
Tuned 2.11 is the current version we are using (matching RHEL 7.7/RHCOS in ocp 4.4) 
Change for supporting cstate can be found here:
https://github.com/redhat-performance/tuned/commit/0ec40e036019c4c062d76a31d676565a09c615dd#diff-6abf9f63e8d9fbfe6bddc23b8addfa1a


slintes:
are we supporting RHEL worker nodes? I assumed RHCOS only until now...?

yanirq:
> are we supporting RHEL worker nodes? I assumed RHCOS only until now...?

Fixed the title and description

yanirq:
This will  be solved by https://bugzilla.redhat.com/show_bug.cgi?id=1789750  : Rebase Tuned in FDP to the latest upstream version (tuned-2.13)

PR #34, opened 2020-01-09
Install RT kernel from RPMs included in RHCOS

slintes:
With this PR we install the RT kernel from the RPMs provided in RHCOS 44.81.
This removes the need for a yum repo.

Not sure if the installation of the RT kernel should be default, but can be disabled on the CRD? Atm it's the other way around, but probably in most cases we want the RT kernel.

TODO:
- [x] investigate failing reboot of worker node -> didn't happen again after rebase
- [x] add functest
- [x] rebase after #25 and #27 are merged


hack/label-worker-rt.sh
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+# expect oc to be in PATH by default
+OC_TOOL="${OC_TOOL:-oc}"
+
+# Label worker nodes
+echo "[INFO]:labeling worker nodes"
+for node in $(${OC_TOOL} get nodes --selector='!node-role.kubernetes.io/master' -o name); do
+    ${OC_TOOL} label $node node-role.kubernetes.io/worker-rt=""
+done
```
cynepco3hahue: nit: new line

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -166,11 +167,15 @@ func getIgnitionConfig(assetsDir string, rtRepoURL string, nonIsolatedCpus *perf
...
+	enableRTKernel := profile.Spec.RealTimeKernel != nil &&
+		profile.Spec.RealTimeKernel.Enabled != nil &&
+		*profile.Spec.RealTimeKernel.Enabled
+
 	ignitionConfig.Systemd = igntypes.Systemd{
 		Units: []igntypes.Unit{
 			{
 				Contents: rtKernelService,
-				Enabled:  pointer.BoolPtr(true),
+				Enabled:  &enableRTKernel,
```
cynepco3hahue: Should we create at all RT kernel unit if the RT disabled, also pre-tunning also depends on the RT kernel unit, so I am interesting if it will work as expected?

slintes: ah yes, good point
I'm wondering if we need the option to not install the RT kernel at all, when other performance features depend on it (no clue if all of them do?)

slintes: other than that, pre-tuning uses `wants` and not `requires`, so it would be started without rt-kernel (to my best understanding)

hack/label-worker-rt.sh
slintes: fixed

davidvossel:
weird, the e2e test failed, but it actually looks like it happened after our e2e test suite successfully completed. I'll run it again. 

```
2020/01/09 17:38:37 Container test in pod e2e-gcp completed successfully
2020/01/09 17:38:38 error: unable to gather container logs: [error: Unable to retrieve logs from pod container cli: container "cli" in pod "e2e-gcp" is terminated, error: Unable to retrieve logs from pod container artifacts: container "artifacts" in pod "e2e-gcp" is terminated, error: Unable to retrieve logs from pod container setup: container "setup" in pod "e2e-gcp" is terminated, error: Unable to retrieve logs from pod container teardown: container "teardown" in pod "e2e-gcp" is terminated, error: Unable to retrieve logs from pod container test: container "test" in pod "e2e-gcp" is terminated]
2020/01/09 17:38:38 error: unable to signal to artifacts container to terminate in pod e2e-gcp, triggering deletion: could not run remote command: unable to upgrade connection: container not found ("artifacts")
2020/01/09 17:38:38 error: unable to retrieve artifacts from pod e2e-gcp: could not read gzipped artifacts: unable to upgrade connection: container not found ("artifacts")
```

functests/performance/rt-kernel.go
```diff
@@ -0,0 +1,69 @@
...
+			testpod.Spec.NodeSelector = map[string]string{
+				fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerRT): "",
+			}
+
+			var err error
+			testpod, err = testclient.Client.Pods(testutils.NamespaceTesting).Create(testpod)
+			if err != nil {
+				return ""
+			}
+			defer func() {
```
davidvossel: can we use an AfterEach() instead. I'm not 100% sure what happens if you do a `ctrl-c` to kill execution when a defer is in use. I know that ginkgo will attempt to run the AfterEach() functions even if you try to kill the test execution early though.

build/assets/scripts/rt-kernel.sh
```diff
@@ -1,55 +1,55 @@
...
-    fi
-}
+# find booted image sha
+sha=$(rpm-ostree status --json | jq -r '.deployments[] | select(.booted == true) | .["custom-origin"][0]' | sed -E "s|.*@(.*)|\1|")
+imageID=$(podman images --digests | grep $sha | awk '{print $4}')
 
-trap exit_handler EXIT
+# create and mount new container
+podman create --net=none --name "$osContainerName" "$imageID" > /dev/null
+kernelDir=$(podman mount "$osContainerName")
```
davidvossel: wow, so we're delivering kernel pkgs in containers? interesting.

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
fromanirh: AFAIK it is correct, if an unit `wants` another unit and the latter can't be started, the former should run anyway (see: https://unix.stackexchange.com/questions/388586/systemd-requires-vs-wants - which in turn quotes manpages).
Question is however is if makes sense to do pretuning without rt-kernel, like Marc mentionedin on his grandparent comment.

build/assets/scripts/rt-kernel.sh
fromanirh: AFAIU this is used to access the packages in the basesystem (so we are not shipping anything anymore). But I may have out of date information.

hack/label-worker-rt.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
fromanirh: who calls this? only humans or also the operator proper? I guess the former, asking to be sure.

build/assets/scripts/rt-kernel.sh
slintes: this is complicated, let me try to summarize:
- the installer's bootstrap node installs an "old" os bootimage on cluster nodes, which is just new enough to be able to do everything which is needed on first boot
- then the MCO installs a newer version of the os ("early pivot"). That one is delivered via container image. Maybe you noticed that OCP comes with a `machine-os-content` image: that one contains the os
- with RHCOS version 44 the image contains the RT kernel RPMs. They were just put into the root directory of the image
- so what I do here (stolen from MCO code): mount the image (which is not possible directly, so create (but don't run) a container from it) which contains the os version which is currently booted, and install the RPMs from the mount path.

longer version: https://github.com/openshift/machine-config-operator/blob/master/docs/OSUpgrades.md and others

hack/label-worker-rt.sh
slintes: we use a `MachineConfig` for installing this script and a systemd unit, and systemd will execute the script then.

build/assets/scripts/rt-kernel.sh
fromanirh: Thanks @slintes great summary. I think it would be great to record this as comment in the `rt-kernel.sh` script itself. WDYT?

functests/performance/rt-kernel.go
slintes: fair point, will change it

build/assets/scripts/rt-kernel.sh
slintes: The operator is for 4.4 only. There we have the RT kernel RPMs already available in the RHCOS image, so no need for a yum repo anymore, but the MCO is not ready to use it yet, see WIP PR [0]. So this is a new temporary workaround for 4.4, unrelated to 4.3.

[0] https://github.com/openshift/machine-config-operator/pull/1330

slintes: The operator is for 4.4 only. There we have the RT kernel RPMs already available in the RHCOS image, so no need for a yum repo anymore, but the MCO is not ready to use it yet, see WIP PR [0]. So this is a new temporary workaround for 4.4, unrelated to 4.3.

[0] openshift/machine-config-operator#1330

slintes: @fromanirh ah missed your comment
will do

slintes: @fromanirh done

hack/label-worker-rt.sh
slintes: oh sorry, just noticed I was looking at the wrong script
I called this one manually on test clusters

PR #36, opened 2020-01-12
Expose status under performance profile resource

### LINKED ISSUE #22: Add status field under the performance profile API
We should decide what fields are actual for `.status` section of the `PerformanceProfile`

cynepco3hahue:
We will get a status update from the machine-config-pool resource that the controller created, and reflect part of these fields under the performance profile status.
We still need to decide how we will monitor tuned changes because the `Tuned` resource does not have any status fields at all.

- [x] add code
- [x] verify that code works manually
- [x] add unittest

Fixes #22


cynepco3hahue:
@slintes @davidvossel Can you please review?

pkg/apis/performance/v1alpha1/performanceprofile_types.go
```diff
@@ -55,11 +56,57 @@ type RealTimeKernel struct {
...
+	// and ready for action.
+	UpdatedMachineCount int32 `json:"updatedMachineCount"`
+
+	// UnavailableMachineCount represents the total number of unavailable (non-ready) machines targeted by performance profile.
+	// A node is marked unavailable if it is in updating state or NodeReady condition is false.
+	UnavailableMachineCount int32 `json:"unavailableMachineCount"`
+
+	// conditions represents the latest available observations of current state.
+	// +optional
+	Conditions []PerformanceProfileCondition `json:"conditions,omitempty"`
```
davidvossel: use structs in  ```github.com/openshift/custom-resource-status/conditions/v1```

there are also a bunch of helper functions in there like ```SetStatusCondition``` and ```IsStatusConditionTrue```

pkg/apis/performance/v1alpha1/performanceprofile_types.go
```diff
@@ -55,11 +56,57 @@ type RealTimeKernel struct {
 	Enabled *bool `json:"enabled,omitempty"`
 }
 
+// PerformanceProfileConditionType valid conditions of a PerformanceProfile.
+type PerformanceProfileConditionType string
+
+const (
```
davidvossel: there are some standard condition types we can probably align with pretty easily.

from the conditions pkg, we have these, which are actually interpreted by the olm to mean something when performing updates and checking availability. 

```
const (         
        // ConditionAvailable indicates that the resources maintained by the operator,
        // is functional and available in the cluster.
        ConditionAvailable ConditionType = "Available"
                
        // ConditionProgressing indicates that the operator is actively making changes to the resources maintained by the
        // operator     
        ConditionProgressing ConditionType = "Progressing"
                
        // ConditionDegraded indicates that the resources maintained by the operator are not functioning completely.
        // An example of a degraded state would be if not all pods in a deployment were running.
        // It may still be available, but it is degraded
        ConditionDegraded ConditionType = "Degraded"
                        
        // ConditionUpgradeable indicates whether the resources maintained by the operator are in a state that is safe to upgrade.
        // When `False`, the resources maintained by the operator should not be upgraded and the
        // message field should contain a human readable description of what the administrator should do to
        // allow the operator to successfully update the resources maintained by the operator.
        ConditionUpgradeable ConditionType = "Upgradeable"
)
```

available=true
progressing=false
degraded=false
upgradable=true

^ that would indicate everything is stable.

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -275,11 +276,19 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
 	// apply components
 	result, err := r.applyComponents(instance)
 	if err != nil {
-		klog.Errorf("failed to deploy components: %v", err)
+		klog.Errorf("failed to deploy performance profile %q components: %v", instance.Name, err)
 		return reconcile.Result{}, err
```
davidvossel: we shouldn't be breaking out early here. If an error occurs during ```applyComponents``` we would treat that as a degraded condition and want to report the error msg in the condition.

pkg/apis/performance/v1alpha1/performanceprofile_types.go
cynepco3hahue: done

pkg/apis/performance/v1alpha1/performanceprofile_types.go
cynepco3hahue: done

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -275,11 +276,19 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
 	}
 
-	// TODO: we need to update the status
+	if err := r.updateStatus(instance); err != nil {
```
slintes: We never get here when `applyComponents()` fails a few lines above. I think we want to update the status based on such an error.

pkg/apis/performance/v1alpha1/performanceprofile_types.go
slintes: I still see updated, updating and degraded, forgot to push? :)

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: done

pkg/controller/performanceprofile/performanceprofile_controller.go
cynepco3hahue: done

pkg/apis/performance/v1alpha1/performanceprofile_types.go
cynepco3hahue: worked on it:) done

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -272,14 +264,41 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+		}
+		// we do not want to reconcile again in case of error, because a user will need to update the PerformanceProfile anyway
+		return reconcile.Result{}, nil
+	}
+
 	// apply components
 	result, err := r.applyComponents(instance)
 	if err != nil {
-		klog.Errorf("failed to deploy components: %v", err)
+		klog.Errorf("failed to deploy performance profile %q components: %v", instance.Name, err)
```
davidvossel: you could collapse all this if/else logic and just pass the applyComponents error result to updateStatus. applyComponents would need to always return a results struct instead of nil though for this to work.

so

```
result, syncErr := applyComponents(instance)
if syncErr != nil {
    klog.Errorf("failed to deploy performance profile %q components: %v", instance.Name, syncErr)
}

if err := r.updateStatus(instance, syncErr); err != nil {
    klog.Errorf("failed to update performance profile %q status: %v", instance.Name, err)
}

return *result, syncErr
```

updateStatus would then set the degraded condition if syncErr != nil

PR #38, opened 2020-01-15
makefile: add a target to only build the binary

fromanirh:
To better integrate with isolated build systems, we could make use
of a target which just builds the binary, without the vet/lint/fmt
steps which we can take for granted - code not compliant should
never be committed in the first place.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,9 +54,9 @@ OPERATOR_DEV_CSV="0.0.1"
 # Export GO111MODULE=on to enable project to be built from within GOPATH/src
 export GO111MODULE=on
 
-build: gofmt golint govet binary
+build: gofmt golint govet dist
 
-binary:
+dist:
 	@echo "Building operator binary"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -54,7 +54,9 @@ OPERATOR_DEV_CSV="0.0.1"
 # Export GO111MODULE=on to enable project to be built from within GOPATH/src
 export GO111MODULE=on
 
-build: gofmt golint govet
+build: gofmt golint govet binary
```
MarSik: I am used to all fashioned

all: for everything
dist: for binaries :)

But otherwise looks good to me.

cynepco3hahue:
lgtm, target names does not critical for me, just decide with Martin what preferred one

Makefile
fromanirh: Since @cynepco3hahue doesn't mind, I'll keep the old target for backward compat and add `dist` as new target

PR #42, opened 2020-01-16
Exclusively select worker node

gsr-shanks:
Some clusters could have master nodes labeled as 'worker'.
This patch only selects worker node and excludes nodes with
'master' label.


slintes:
> Some clusters could have master nodes labeled as 'worker'.

And some other clusters might have no exlusive workers... :)
Just curious, what is the problem with the combined master/worker node?

gsr-shanks:
> > Some clusters could have master nodes labeled as 'worker'.
> 
> And some other clusters might have no exlusive workers... :)
> Just curious, what is the problem with the combined master/worker node?

I was under the impression that we deploy RT Kernel only on worker nodes. Is it not the case?

davidvossel:
> I was under the impression that we deploy RT Kernel only on worker nodes. Is it not the case?

yeah, actually you're right. I bet our configuration would fail on a node that's both a master and worker. That's due to the MCO not layering roles like we originally thought it did

davidvossel:
we're seeing that same govet build error that is showing up in other PRs. this isn't directly related to this PR. I'm working to determine what's going on here. 

```
Running go vet
go vet github.com/openshift-kni/performance-addon-operators/...
go: github.com/openshift/machine-config-operator@v4.2.0-alpha.0.0.20190917115525-033375cbe820+incompatible requires
	github.com/cri-o/cri-o@v1.9.0-beta.2.0.20191003162030-4775e1c05c26 requires
	github.com/containers/image/v5@v5.0.0 requires
	github.com/mtrmac/gpgme@v0.0.0-20170102180018-b2432428689c: invalid version: unknown revision b2432428689c
```

davidvossel:
/hold

something got messed up during the rebase to master. Somehow there's now 22 commits in this PR.

gsr-shanks:
> /hold
> 
> something got messed up during the rebase to master. Somehow there's now 22 commits in this PR.

Yeah, looks like I messed it up. Closing this one.

PR #40, opened 2020-01-15
Removed creation of MCP by operator

slintes:
- The MCP needs to be created by an admin, the operator does not create it anymore, and we don't modify it.
- In order to set the right relationships between existing MCP and resources created by the operator (MC, KC, Tuned), we need some additional information in the CR now.
- Deploy MCP, operator and profile using kustomize.
- Cleanup: removed SCTP and setup_test_cluster.

Untested due to lack of test cluster atm, let's see what CI thinks.

PR #44, opened 2020-01-20
Add ability to "pause" a performance profile's reconcile loop

davidvossel:
Placing the "performance.openshift.io/pause-reconcile=true" annotation on a performance profile object will prevent the operator from reconciling on that object.

This is an escape hatch that allows someone to manually modify a configuration owned by the performance profile and not have our operator stomp on their changes.

### DIFF
--- a/pkg/controller/performanceprofile/performanceprofile_controller.go
+++ b/pkg/controller/performanceprofile/performanceprofile_controller.go
@@ -296,7 +296,27 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
 	return reconcile.Result{}, nil
 }
 
+func isPaused(profile *performancev1alpha1.PerformanceProfile) bool {
+
+	if profile.Annotations == nil {
+		return false
+	}
+
+	isPaused, ok := profile.Annotations[performancev1alpha1.PerformanceProfilePauseAnnotation]
+	if ok && isPaused == "true" {
+		return true
+	}
+
+	return false
+}
+
 func (r *ReconcilePerformanceProfile) applyComponents(profile *performancev1alpha1.PerformanceProfile) (*reconcile.Result, error) {
+
+	if isPaused(profile) {
+		klog.Infof("Ignoring reconcile loop for pause performance profile %s", profile.Name)
+		return nil, nil
+	}
+
 	// get mutated machine config
 	mc, err := machineconfig.New(r.assetsDir, profile)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -316,7 +316,27 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
 	return reconcile.Result{}, nil
 }
 
+func isPaused(profile *performancev1alpha1.PerformanceProfile) bool {
```
slintes: in case this is merged after #40: this would be a good fit for the new `pkg/controller/performanceprofile/components/profile/profile.go`

pkg/controller/performanceprofile/performanceprofile_controller_test.go
```diff
@@ -234,6 +234,50 @@ var _ = Describe("Controller", func() {
 			Expect(mcp.Spec.Paused).To(BeFalse())
 		})
 
+		It("should do nothing when pause annotation is set", func() {
```
slintes: would be nice if this weren't between the "should do xyz on nth reconcile loop" tests

davidvossel: yup, i can move it in a more logical place

simon3z:
> /cc @simon3z

Thanks!

@davidvossel @MarSik BTW what's our plan to document (or generate documentation) for these?
ATM we're not planning to document it? (Because it's an escape hatch, etc.)

davidvossel:
> @davidvossel @MarSik BTW what's our plan to document (or generate documentation) for these?
ATM we're not planning to document it? (Because it's an escape hatch, etc.)

@simon3z I'd prefer not to document this and keep it as a hidden "unsupported" feature. It's here for emergencies.

pkg/controller/performanceprofile/performanceprofile_controller.go
davidvossel: yup, fixed. I added the function to that pkg, but had to import the profile package with the `profileutils` alias to prevent conflicting with the commonly named `profile` struct that represents the performance profile.

pkg/controller/performanceprofile/performanceprofile_controller_test.go
davidvossel: fixed, and I updated the logic to reflect your #40 changes

PR #46, opened 2020-01-23
Make it possible to run tests against an external cluster

fedepaol:
`make functests` also setup labels and deploy the operator (see https://github.com/openshift-kni/performance-addon-operators/blob/f8ca736d9c83ef39dd1e581d855c94865a565400/Makefile#L126)

Also, the tests are pinned to the labels / profiles applied in the setup phase.
This PR makes those configurable and adds a makefile rule that only runs tests.

### DIFF
--- a/functests/performance/performance.go
+++ b/functests/performance/performance.go
@@ -2,6 +2,7 @@ package performance
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"strings"
 	"time"
@@ -21,6 +22,15 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 )
+
+var profileName string
+
+func init() {
+	profileName = os.Getenv("PERF_TEST_PROFILE")
+	if profileName == "" {
+		profileName = "ci"
+	}
+}
 
 const (
 	testTimeout      = 480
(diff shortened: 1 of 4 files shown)


functests/performance/performance.go
```diff
@@ -22,6 +23,15 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
+var profileName string
+
+func init() {
+	profileName = os.Getenv("PERF_TEST_PROFILE")
```
slintes: So we would still have the profilename at 2 places, in the manifest, and in the env var for tests.
Can we put the env var into the Makefile, and use it in the manifest as well? Should work because we pipe manifests through envsubst before applying them.

fedepaol: So if I understood correctly, you mean getting rid completely of the hardcoded one ("ci") and always relying on the env variable one right?

fedepaol: Also, does the same apply to `RoleWorkerRT` ? Because I am seeing that in many places and it ends up being a lot less readable if we set it from an env variable.

fedepaol: Another point: the envsubst is done at `ci-cluster/performance` level, not on base : https://github.com/openshift-kni/performance-addon-operators/blob/ae7d00b169ae271d25aff28071117e5dc108d4bc/hack/deploy.sh#L30

I don't see an easy way to pipe the envsubst here because `base` is read from kustomize. Any ideas? Can the duplication of `ci` be handed later as it was already in place? 
I am only adding a way to override its default value here.

slintes: forget what I wrote. I missed the point that you want to run tests against a cluster which was not prepared by this repo.

PR #47, opened 2020-01-23
Remove RT kernel workaround and use new MC config instead

slintes:
With https://github.com/openshift/machine-config-operator/pull/1330 being merged, we can remove our workaround for installing the RT kernel and use the new config option of MachineConfig.

Not sure when this will be available in CI, so it might fail yet.

Edit: accidentally pushed a change on `go vet`, let's see what CI thinks about it. The original version did not work for me with Go 1.13.5


slintes:
/hold

e2e test isn't required but should be green for this

slintes:
cluster bootstrap failed

/retest

slintes:
The RT kernel test passed! I only forgot to remove another test which checks existence of the removed rt kernel script.

slintes:
And we are green :)
@dvossel do you mind putting a new lgmt on it? Thanks!

/hold cancel

slintes:
> And we are green :)
> @davidvossel do you mind putting a new lgmt on it? Thanks!
> 
> /hold cancel

LOL, you were a bit faster than me :)

PR #49, opened 2020-01-24
Exclusively select worker node

gsr-shanks:
Some clusters could have master nodes labeled as 'worker'.
This patch only selects worker node and excludes nodes with
'master' label.

### DIFF
--- a/hack/label-worker-rt.sh
+++ b/hack/label-worker-rt.sh
@@ -7,5 +7,7 @@ OC_TOOL="${OC_TOOL:-oc}"
 
 # Label 1 worker node
 echo "[INFO]: Labeling 1 worker node with worker-rt"
-node=$(${OC_TOOL} get nodes --selector='node-role.kubernetes.io/worker' -o name | head -1)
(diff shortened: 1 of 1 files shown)


slintes:
I still do not understand why we need this tbh.
If someone wants to use a node as master and worker, then our patches should work and be tested on such a node as well, or not? Which leads me to the question: does it not work? If so, what is the failure?

cynepco3hahue:
@slintes At least from what I remember, currently we should apply RT kernel features only on workers because masters can run critical for the cluster workloads, but by default, KNI deploy scripts, applies worker role to masters(to reduce the number of needed nodes, but it does not default view of the cluster). IMHO the change makes sense.

slintes:
> because masters can run critical for the cluster workloads

Sure, but masters run in HA setup, and MCO will only apply new configs and reboot one after another. It should just work. There might short times where you can't connect to API server during updates, but that's just caused by the proxy in front of them, when it doesn't exclude the current unavailable node fast enough.

However, I won't block on this, just want to understand the reason for this.

/lgtm

PR #48, opened 2020-01-23
Custom CSV Generation logic

davidvossel:
This tool allows us to inject custom information related to the package into the CSV, like description, keywords, maintainers, replaces version, etc...

This logic works by generating a temporary CSV using the operator-sdk which the tool then reads in and injects all our custom data into. 

## Building Custom CSV

A custom CSV entry for the registry container can be generated using the
`make generate-csv` make target.

First export your CSV details as environment variables.

__required options__
export IMAGE_REGISTRY="quay.io"
export REGISTRY_NAMESPACE="some-operator=repo-namespace"
export IMAGE_TAG="some-operator-image-tag"
export CSV_VERSION="0.0.3"

__optional options__
export REPLACES_CSV_VERSION="0.0.2"
export CSV_SKIP_RANGE=">=0.0.1 <0.0.2"

Then run `make generate-csv`

The result will be stored in the deploy/olm-catalog/performance-addon-operator
directory within a directory that matches the CSV_VERSION set.

Running `make registry-container` after creating a new custom CSV will result
in a registry bundle that includes the new CSV version and all other CSV
versions in the deploy/olm-catalog directory.

PR #50, opened 2020-01-27
Updated alm-exaple in CSV

slintes:


deploy/olm-catalog/performance-addon-operator/0.0.1/performance-addon-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -12,24 +12,24 @@ metadata:
           },
           "spec": {
             "cpu": {
-              "isolated": "2-3",
+              "isolated": "1-3",
```
MarSik: Isolated and reserved must not overlap. Also for now, making nonIsolated the same as reserved is probably the best approach. In the future nonIsolated will be the same as all cpus - isolated and reserved will be subset of nonIsolated.

slintes: ok, thx, fixed

fromanirh:
Looks like issue is in CI:
```could not wait for build: the build root failed after 55s with reason DockerBuildFailed: Docker build strategy has failed.

Fedora 31 - x86_64                               11 MB/s |  71 MB     00:06    
No match for argument: golang-1.13.5-1.fc31
Package tar-2:1.32-2.fc31.x86_64 is already installed.
Error: Unable to find a match: golang-1.13.5-1.fc31
error: build error: running 'dnf -y install     $GOVERSION...ext     which     && dnf clean all' failed with exit code 1```

slintes:
> Looks like issue is in CI:
> 
> ```
> 
> Fedora 31 - x86_64                               11 MB/s |  71 MB     00:06    
> No match for argument: golang-1.13.5-1.fc31
> Package tar-2:1.32-2.fc31.x86_64 is already installed.
> Error: Unable to find a match: golang-1.13.5-1.fc31
> error: build error: running 'dnf -y install     $GOVERSION...ext     which     && dnf clean all' failed with exit code 1```
> ```

yes, fix is on its way already in #51

davidvossel:
/hold

this logic needs to be codified and go into #48 once it is merged. 

We can inject whatever alm examples we want inline within that generation logic.

davidvossel:
also, this won't pass CI. There's logic in the `make ci-job` make target to ensure we never make manual changes to a CSV file. It will be detected that these changes aren't auto generated and fail.

fromanirh:
> /hold
> 
> this logic needs to be codified and go into #48 once it is merged.
> 
> We can inject whatever alm examples we want inline within that generation logic.

Ack, I didn't realize this.

slintes:
Ouch, yes, I missed that the example is coming from `deploy/crds/performance.openshift.io_v1alpha1_performanceprofile_cr.yaml`

Closing for now, will create new PR if needed after #48 is merged

/close

PR #55, opened 2020-01-28
Don't ignore existing image vars

slintes:
Might be useful for testing existing images


davidvossel:
/approve

fyi, the same thing can be achieved using the existing env-vars, you'd just have to set more

IMAGE_REGISTRY="quay.io"
REGISTRY_NAMESPACE="my-quay-account"
OPERATOR_IMAGE_NAME="performance-operator"
REGISTRY_IMAGE_NAME="performance-operator-registry"
IMAGE_TAG="latest"

but i can see why it's useful to be able to set the image strings directly as well.

PR #56, opened 2020-01-28
Remove worker-rt MCP with deploy clean-up

gsr-shanks:
Remove worker-rt MCP while performing 'make cluster-clean'


slintes:
hm, exit code 56 probably is curl (failure in receiving network data) with a temporary error. Let's try again....

/retest

slintes:
hm, works locally and on other PRs. The change can't cause the problem. Let's retry later...

slintes:
@gsr-shanks sorry, you need to rebase

gsr-shanks:
Looks like I missed something.

PR #58, opened 2020-01-28
Fix csv dir permissions

davidvossel:


davidvossel:
/retest

ci is failing before we can even get to our tests

PR #60, opened 2020-01-29
Add possibility to specify NUMA node for hugepages allocation

cynepco3hahue:
Once you will specify the NUMA node where you want to allocate hugepages,
the code instead of adding required hugepages to kernel boot parameters, will
create systemd unit that will call simple bash script, that will write the specified value
to `/sys/devices/system/node/node<NUMA_NODE_ID>/hugepages/<HUGEPAGE_SIZE>/nr_hugepages`.

It should give a more flexible way to configure a node for the performance-sensitive workloads.


build/assets/scripts/hugepages-allocation.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+nodes_path="/sys/devices/system/node"
+hugepages_file="${nodes_path}/node${NUMA_NODE}/hugepages/hugepages-${HUGEPAGES_SIZE}kB/nr_hugepages"
+
+if [ ! -d  ${hugepages_file} ]; then
```
slintes: I think this needs to be `-f`

cynepco3hahue: nice catch, copy-paste mistake

cynepco3hahue:
> Looks almost good to me.
> Any chance to get a functest as well? :)

Will do, probably I will need to add additional performance profile, that will include hugepages with specified NUMA node.

functests/performance/hugepages.go
```diff
@@ -0,0 +1,47 @@
...
+		Expect(workerRTNodes).ToNot(BeEmpty())
+		workerRTNode = &workerRTNodes[0]
+	})
+
+	Context("when NUMA node specified", func() {
+		It("should be allocated on the specifed NUMA node ", func() {
+			command := []string{"cat", pathHugepages1048576kBNumaNode0}
+			nrHugepages, err := nodes.ExecCommandOnMachineConfigDaemon(testclient.Client, workerRTNode, command)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(string(nrHugepages)).To(Equal("1"))
```
davidvossel: is there anyway to verify the system picked up this change that goes beyond checking the file?

like maybe inspecting ```cat /proc/meminfo  | grep Huge``` ?

also, can we test that the systemd unit didn't fail and return non-zero exit code?

functests/performance/hugepages.go
```diff
@@ -0,0 +1,47 @@
...
+		var err error
+		workerRTNodes, err := nodes.GetByRole(testclient.Client, testutils.RoleWorkerRT)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(workerRTNodes).ToNot(BeEmpty())
+		workerRTNode = &workerRTNodes[0]
+	})
+
+	Context("when NUMA node specified", func() {
+		It("should be allocated on the specifed NUMA node ", func() {
+			command := []string{"cat", pathHugepages1048576kBNumaNode0}
```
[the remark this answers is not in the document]
cynepco3hahue: will do, thanks for comment

pkg/apis/performance/v1alpha1/performanceprofile_types.go
```diff
@@ -66,6 +66,9 @@ type HugePage struct {
 	Size HugePageSize `json:"size,omitempty"`
 	// Count defines amount of huge pages, maps to the 'hugepages' kernel boot parameter.
 	Count int32 `json:"count,omitempty"`
+	// Node defines the NUMA node where hugepages will be allocated,
+	// if not specified, pages will be allocated equally between NUMA nodes
```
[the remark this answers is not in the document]
cynepco3hahue: done

functests/performance/hugepages.go
cynepco3hahue: `meminfo` shows only information regarding default hugepages, in our case, it will be 1GB, and IIRC it looks exactly at the same file that used under the test

functests/performance/hugepages.go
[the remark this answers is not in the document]
cynepco3hahue: oh actually we do not have clientset for the PerformanceProfile, so I believe we should leave it for another PR, @davidvossel WDT?

functests/performance/hugepages.go
MarSik: cat /sys/devices/system/node/node*/meminfo | grep Huge

cynepco3hahue: @MarSik but does it different from `/sys/devices/system/node/node0/hugepages/hugepages-1048576kB/nr_hugepages`?

cynepco3hahue: @MarSik and again meminfo just show information regarding default hugepages size

cynepco3hahue:
@MarSik @slintes @davidvossel I added basic validation, and currently, disabled one of functests, because it needs or definition of multiple hugepages under the performance profile resource, that currently impossible because of https://github.com/openshift-kni/performance-addon-operators/issues/63 or specification of the additional performance profile, currently I want to avoid it because it will need an update of some hacks scripts, I will do it in the following PR, the same true for more dynamic tests(get hugepages values from the performance profile).

davidvossel:
> that currently impossible because of #63 

use the controller-runtime pkg for the client

here's an example, https://github.com/openshift/ocs-operator/blob/master/functests/storageclusterinitialization_test.go#L64

cynepco3hahue:
> use the controller-runtime pkg for the client

will refactor our functional tests to use controller-runtime client, can it be done under another PR?

cynepco3hahue:
@davidvossel Relevant PR - https://github.com/openshift-kni/performance-addon-operators/pull/69

pkg/controller/performanceprofile/components/profile/profile.go
```diff
@@ -95,3 +110,25 @@ func IsPaused(profile *v1alpha1.PerformanceProfile) bool {
...
+	}
+	hugepagesSizes := map[v1alpha1.HugePageSize]string{}
+	for _, page := range hugepages.Pages {
+		hugepagesSizes[page.Size] = ""
+	}
+
+	// TODO: this validation should be removed, once https://github.com/kubernetes/kubernetes/pull/84051
+	// is available under the openshift
+	// validate that we do not have allocations of hugepages of different sizes
+	if len(hugepagesSizes) > 1 {
```
davidvossel: this is clever :+1:

cynepco3hahue:
> Can we update the func test to validate the hugepages off of the posted performance profile now that the client has been updated. I'd just like to see the functional test pass on a node with multiple numa nodes before we ack this. It's okay if that verification is manual for the time being even. "manual" meaning the funct test is run manually on a custom setup, not that we inspect the cluster node manually.

Sure will do!

PR #62, opened 2020-01-29
Remove worker-rt MCP with deploy clean-up

gsr-shanks:
Remove worker-rt MCP while performing 'make cluster-clean'

### DIFF
--- a/hack/clean-deploy.sh
+++ b/hack/clean-deploy.sh
@@ -8,3 +8,6 @@ $OC_TOOL delete performanceprofile --all
 $OC_TOOL delete ns openshift-performance-addon
 
 $OC_TOOL -n openshift-marketplace delete catalogsource performance-addon-operator-catalogsource
+
+$OC_TOOL delete mcp worker-rt
+


hack/clean-deploy.sh
```diff
@@ -7,4 +7,6 @@ $OC_TOOL delete performanceprofile --all
 
 $OC_TOOL delete ns openshift-performance-addon
 
-$OC_TOOL -n openshift-marketplace delete catalogsource performance-addon-operator-catalogsource
\ No newline at end of file
+$OC_TOOL -n openshift-marketplace delete catalogsource performance-addon-operator-catalogsource
+
+$OC_TOOL delete mcp worker-rt
```
slintes: missing new line at the end

gsr-shanks: Done.

cynepco3hahue:
@gsr-shanks Removing the MCP before you remove the role from the node will drop the node to the degraded state.
Once you remove MCP, the system will remove rendered MC, so machine config controller will fail to find the current configuration and will fail to rollback node to default worker MCP.

IMHO the right way for clean up:
- remove `worker-rt` role from the node
- wait until the node appear back under the worker MCP
- wait for worker MCP updated status
- remove the worker-rt MCP

slintes:
> @gsr-shanks Removing the MCP before you remove the role from the node will drop the node to the degraded state.
> Once you remove MCP, the system will remove rendered MC, so machine config controller will fail to find the current configuration and will fail to rollback node to default worker MCP.
> 
> IMHO the right way for clean up:
> 
> * remove `worker-rt` role from the node
> * wait until the node appear back under the worker MCP
> * wait for worker MCP updated status
> * remove the worker-rt MCP

oh wow, didn't know that.
sounds like a MCO bug to me
At least I would expect that MCO just switches the node to worker MCP when worker-rt is removed

cynepco3hahue:
@slintes Yep, I still need to open one, just wanted to have the appropriate error message from logs, will do it today.

slintes:
superseded by #67 

/close

PR #65, opened 2020-01-30
Generate client set

### LINKED ISSUE #63: Generate client set for our API
It possible situation when we want to get a performance profile via API, for example, to make our functional test less dependent on the hard-coded configuration, but currently, it does not possible, because that we do not have client set for performance profile resources.

NOTE:
I believe we can get performance profile resources via the REST client, but I prefer to avoid it and work with the API.

cynepco3hahue:
This PR provides scripts to generate an updated client set for performance profile API and generated client set.

Fixes #63


slintes:
Nice! But go vet seems to be unhappy...

slintes:
> added cherry-pick-approved and removed cherry-pick-approved labels 34 seconds ago

just ignore this, was testing if we can just set that label. So we can use it for cherry-pick into release branches later

davidvossel:
@cynepco3hahue @slintes 

How are we planning to use this generated client? I think the ```sigs.k8s.io/controller-runtime/pkg/client``` mostly makes generating our own clients irrelevant now.

cynepco3hahue:
@davidvossel Good point, so probably we will need to refactor all functests(under the operator and under the cnf-features-deploy) to use controller-runtime client.

slintes:
superseded by #69 

/close

PR #67, opened 2020-01-31
Improved cleanup script

slintes:
Improved cleanup script
Successfully tested on my cluster

@gsr-shanks fyi

### DIFF
--- a/hack/clean-deploy.sh
+++ b/hack/clean-deploy.sh
@@ -3,8 +3,61 @@
 # expect oc to be in PATH by default
 OC_TOOL="${OC_TOOL:-oc}"
 
+# Remove node label
+echo "[INFO]: Unlabeling worker nodes"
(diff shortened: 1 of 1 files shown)


hack/clean-deploy.sh
```diff
@@ -3,8 +3,61 @@
...
+# Remove node label
+echo "[INFO]: Unlabeling worker nodes"
+nodes=$(${OC_TOOL} get nodes --selector='node-role.kubernetes.io/worker-rt' -o name)
+for node in $nodes
+do
+    ${OC_TOOL} label $node node-role.kubernetes.io/worker-rt-
+done
+
+# Give MCO some time to notice change
+sleep 10
```
yanirq: is that enough ? or maybe we want some retries here so the cleanup/following tests wont be flaky perhaps?

PR #68, opened 2020-02-01
Added test verifying basic hugepages usage by pods

dshchedr:


functests/performance/hugepages.go
```diff
@@ -0,0 +1,120 @@
...
+		testpod = getCentosPod(podCommand)
+		testpod, err := testclient.Client.Pods(testutils.NamespaceTesting).Create(testpod)
+		Expect(err).ToNot(HaveOccurred())
+		Eventually(func() string {
+			logs, err := pods.GetLogs(testclient.Client, testpod)
+			if err != nil {
+				return ""
+			}
+			return logs
+		}, 5*time.Minute, 10*time.Second).Should(ContainSubstring("Complete!"))
```
vladikr: can we pre-install these tools somehow? 5 mins seem long :)

vladikr:
It looks good to me, I'm not sure about the 5 minutes timeout.

functests/performance/hugepages.go
dshchedr: It's just timeout, in real this test takes around 1 minute. 
Unfortunately, I didn't find any containers with pre-installed libhugetlbfs tool..

cynepco3hahue:
looks like some command fails under the pod, can you divide the creation of the pod and execution of the command, it will make debug easier

functests/performance/hugepages.go
```diff
@@ -0,0 +1,124 @@
+package performance
+
+import (
+	"context"
```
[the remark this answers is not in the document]
dshchedr: done

dshchedr:
> looks like some command fails under the pod, can you divide the creation of the pod and execution of the command, it will make debug easier

Downloading centos image takes more than 60 seconds. Definitely, we need to think about pre-installed test image.. but i think it should be another PR

functests/performance/hugepages.go
```diff
@@ -77,4 +84,100 @@ var _ = Describe("[performance]Hugepages", func() {
...
+
+		AfterEach(func() {
+			err := testclient.Client.Delete(context.TODO(), testpod)
+			Expect(err).ToNot(HaveOccurred())
+
+			err = pods.WaitForDeletion(testclient.Client, testpod, 60*time.Second)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("Pod should be able to run process with using hugepages", func() {
```
cynepco3hahue: can you adapt this test to take values from the performance profile, instead of using hardcoded paths for hugepages?

functests/performance/hugepages.go
```diff
@@ -77,4 +84,100 @@ var _ = Describe("[performance]Hugepages", func() {
 	// 		Expect(string(nrHugepages)).To(Equal("128"))
 	// 	})
 	// })
+
+	Context("Basic hugepages usage verification", func() {
```
cynepco3hahue: also, can you add polarion test details, similar to CNV format - https://github.com/kubevirt/kubevirt/blob/9dd17aae95b027e8c7948d570d06aac173caea53/tests/console_test.go#L38

functests/performance/hugepages.go
dshchedr: I thought about using profile for it, but in this test I'm creating the pod with 1G hugepages only, so i don't need to have mutable links..
but if you think we can have cluster with 2M hugepages only - so yeah, need to upgrade test..

cynepco3hahue: Do you want to do it under the current PR or under the separate PR?

dshchedr: I've updated code - now it works for 1G and 2M hugepages

PR #69, opened 2020-02-02
functests: use controller-runtime client

cynepco3hahue:
NOTE: we still have one place where we must to use k8s
clientset, because controller-runtime client does not support
subresources except the status one, so it imposible to get logs
from the pod via controller-runtime client.


slintes:
lgtm, but giving others a chance to review as well

PR #71, opened 2020-02-04
Added test verifying CPU affinity and reservation

dshchedr:


functests/performance/cpu_management.go
```diff
@@ -0,0 +1,125 @@
...
+		workerRTNodes, err := nodes.GetByRole(testclient.Client, testutils.RoleWorkerRT)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(workerRTNodes).ToNot(BeEmpty())
+		workerRTNode = &workerRTNodes[0]
+	})
+
+	It("Verify CPU reservation on the node (allocatable cpu should be less then capacity)", func() {
+		capacityCpu, _ := workerRTNode.Status.Capacity.Cpu().AsInt64()
+		allocatableCpu, _ := workerRTNode.Status.Allocatable.Cpu().AsInt64()
+		Expect(capacityCpu - allocatableCpu).To(Equal(int64(1)))
```
vladikr: we should probably make it visible somehow that we've reserved 1 CPU

vladikr:
Looks great over all! I think we should only make it more visible which configuration values are we testing against.

functests/performance/cpu_management.go
```diff
@@ -0,0 +1,192 @@
...
+				}
+			}
+			err := testclient.Client.Create(context.TODO(), testpod)
+			Expect(err).ToNot(HaveOccurred())
+
+			err = pods.WaitForCondition(testclient.Client, testpod, corev1.PodReady, corev1.ConditionTrue, 60*time.Second)
+			Expect(err).ToNot(HaveOccurred())
+
+			cmd := []string{"/bin/bash", "-c", "ps -o psr $(pgrep -n stress) | tail -1"}
+			Expect(strings.Trim(execCommandOnWorker(cmd, workerRTNode), " ")).To(BeElementOf(listCPU))
```
cynepco3hahue: you already `Trim` under the `execCommandOnWorker`

cynepco3hahue: I can see that you are checking vs `listReservedCPU` for guaranteed and non-guaranteed pods, does it by intent?

functests/performance/cpu_management.go
```diff
@@ -0,0 +1,192 @@
...
+	var listReservedCPU []string
+	var listIsolatedCPU []string
+
+	BeforeEach(func() {
+		workerRTNodes, err := nodes.GetByRole(testclient.Client, testutils.RoleWorkerRT)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(workerRTNodes).ToNot(BeEmpty())
+		workerRTNode = &workerRTNodes[0]
+
+		isolatedCPU = getIsolatedCpuFromMachineConfig()
```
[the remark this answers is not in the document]
dshchedr: Agree, will do that. Initially we didn't have performance profile

functests/performance/cpu_management.go
dshchedr: In execCommandOnWorker we are trimming "\n", buy i need to remove _spaces_ since the output of "ps" command looks like _"   0 "_

dshchedr: hm, not clear what do you mean. If it is non-guaranteed pod **listCPU := listReservedCPU**. If pod is guaranteed **listCPU = listIsolatedCPU**

cynepco3hahue: missed this variable under the `if` condition

PR #73, opened 2020-02-05
API Isolated and NonIsolated CPU changes

davidvossel:
## Overview

The spec.CPU.NonIsolated option is redundant. It can be represented by the same cpu set that spec.CPU.Reserved represents now. If we find we need fine granular control over how the CPUs in the reserved set are used, we can re-approach making an addition to the API at that point. 

The spec.CPU.Isolated option needs two modes, one that just uses the tuned options for cpu partitioning, and another that includes both the tuned settings plus setting the isolcpu kernel option.

The reasoning here is that using the isolcpus kernel option is not always appropriate depending on the use case. We might want to partition a set of isolated cpus and still allow those cpus to load balance. With the addition of the spec.CPU.EnableIsolcpus boolean value, we can toggle on usage of isolcpus kernel args independently of the tuned partitioning.

## Changes

- Removed NonIsolated API option. This now defaults to Reserved
- Added EnableIsolcpus API option for toggling between using kernel isolcpus arg or not for the isolated cpuset. This option defaults to false.

### DIFF
--- a/build/assets/scripts/pre-boot-tuning.sh
+++ b/build/assets/scripts/pre-boot-tuning.sh
@@ -76,7 +76,6 @@ else
 
     type getargs >/dev/null 2>&1 || . /lib/dracut-lib.sh
 
-    #cpumask="$(getargs reserved_cpumask)"
     cpumask='$reserved_cpumask'
 
     log()
--- a/build/assets/scripts/pre-boot-tuning.sh
+++ b/build/assets/scripts/pre-boot-tuning.sh
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
-non_iso_cpumask=""
+reserved_cpumask=""
 cpu_affinity=""
 
 get_reserved_cores() {
@@ -15,10 +15,10 @@ get_reserved_cores() {
         else
             cores+=($part)
         fi
-    done < <( echo ${NON_ISOLATED_CPUS} | tr ',' '\n' )
+    done < <( echo ${RESERVED_CPUS} | tr ',' '\n' )
 }
 
-# $1 - 0 for irq balance banned cpus masking , 1 for non isolated cpus masking
+# $1 - 0 for irq balance banned cpus masking , 1 for reserved cpus masking
 get_cpu_mask() {
     if [ "$1" = "1" ]; then
         mask=( 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 )
@@ -31,7 +31,7 @@ get_cpu_mask() {
     done
     cpumaskBinary=`echo ${mask[@]}| rev`
     cpumaskBinary=${cpumaskBinary//[[:space:]]/}
-    non_iso_cpumask=`printf '%08x\n' "$((2#$cpumaskBinary))"`
+    reserved_cpumask=`printf '%08x\n' "$((2#$cpumaskBinary))"`
 }
 
 get_cpu_affinity() {
@@ -56,7 +56,7 @@ if [ -f ${INITRD_NEW_IMAGE} ] && grep -qs "iso_initrd.img" ${CURRENT_ENTRY_FILE}
     echo "Pre boot tuning configuration already applied"
     echo "Setting kernel rcuo* threads to the housekeeping cpus"
     get_cpu_mask 1
-    pgrep rcuo* | while read line; do taskset -pc ${NON_ISOLATED_CPUS} $line || true; done
+    pgrep rcuo* | while read line; do taskset -pc ${RESERVED_CPUS} $line || true; done
 else
     # Clean up
     rm -rf ${INITRD_GENERATION_DIR}
@@ -76,8 +76,8 @@ else
 
     type getargs >/dev/null 2>&1 || . /lib/dracut-lib.sh
 
-    #cpumask="$(getargs non_iso_cpumask)"
-    cpumask='$non_iso_cpumask'
+    #cpumask="$(getargs reserved_cpumask)"
+    cpumask='$reserved_cpumask'
 
     log()
     {
@@ -93,14 +93,14 @@ else
     done
     fi' > ${INITRD_GENERATION_DIR}/usr/lib/dracut/hooks/pre-udev/00-tuned-pre-udev.sh
 
-    # Set CPU affinity according to NON_ISOLATED_CPUS
+    # Set CPU affinity according to RESERVED_CPUS
     get_cpu_affinity
     echo "[Manager]" >> ${INITRD_GENERATION_DIR}/etc/systemd/system.conf
     echo "CPUAffinity=$cpu_affinity" >> ${INITRD_GENERATION_DIR}/etc/systemd/system.conf
 
-    # Set IRQ banned cpu according to NON_ISOLATED_CPUS
+    # Set IRQ banned cpu according to RESERVED_CPUS
     get_cpu_mask 0
-    echo "IRQBALANCE_BANNED_CPUS=$non_iso_cpumask" >> ${INITRD_GENERATION_DIR}/etc/sysconfig/irqbalance
+    echo "IRQBALANCE_BANNED_CPUS=$reserved_cpumask" >> ${INITRD_GENERATION_DIR}/etc/sysconfig/irqbalance
     
     cd ${INITRD_GENERATION_DIR}
     find . | cpio -co >${INITRD_NEW_IMAGE}
--- a/build/assets/tuned/openshift-node-real-time-kernel
+++ b/build/assets/tuned/openshift-node-real-time-kernel
@@ -18,5 +18,7 @@ kernel.timer_migration = 0
 [sysfs]
 /sys/devices/system/machinecheck/machinecheck*/ignore_ce = 1   
 
+{{if .IsolatedCpus}}
 [scheduler]
-{{if .IsolatedCpus}}isolated_cores={{.IsolatedCpus}}{{end}}
+isolated_cores={{.IsolatedCpus}}
+{{end}}
--- a/build/assets/tuned/openshift-node-real-time-kernel
+++ b/build/assets/tuned/openshift-node-real-time-kernel
@@ -19,4 +19,4 @@ kernel.timer_migration = 0
 /sys/devices/system/machinecheck/machinecheck*/ignore_ce = 1   
 
 [scheduler]
-isolated_cores={{.IsolatedCpus}}
+{{if .IsolatedCpus}}isolated_cores={{.IsolatedCpus}}{{end}}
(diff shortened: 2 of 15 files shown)


deploy/crds/performance.openshift.io_performanceprofiles_crd.yaml
```diff
@@ -34,15 +34,23 @@ spec:
             cpu:
               description: CPU defines set of CPU related parameters.
               properties:
+                enableIsolcpus:
```
MarSik: Do we really need the boolean? What about just checking the isolated string is nil or empty?

davidvossel: the issue from what I could see is there's tuned partitioning that uses the Isolated CPUSet independently of whether or not we set the isolcpu kernel arg.

davidvossel: @MarSik the Isolated cpu set is used in two different configs.

1. In a tuned config to ste teh isolated_cores cpu set.
```
[scheduler]
isolated_cores="4-7"
```

2. In a MachineConfig to set the IRQBALANCE_BANNED_CPUS=11110000 bitmask for irqbalance.

```
echo "IRQBALANCE_BANNED_CPUS=11110000" >> ${INITRD_GENERATION_DIR}/etc/sysconfig/irqbalance
```

Since item [2] here lets us execute arbitrary bash logic on the actually host, the inverse of `Reserved` is possible. However for item [1], since the tuned profile is generated at the cluster level, the only way to compute the inverse of `Reserved` is to know how many cpus are on the matching hosts, which isn't something we can get from the node objects themselves by default.

So, basically the Isolated option is required because there isn't a reasonable way (that i'm aware of) to calculate the inverse of `Reserved` at the cluster level. Because of this, we need the `EnableIsolcpu` boolean to indicate if that kernel arg should be set or not independently of the `Isolated` cpu set

make sense?

davidvossel:
updated to replace all references of nonIsolated with Reserved.

build/assets/scripts/pre-boot-tuning.sh
```diff
@@ -74,8 +74,8 @@ else
 
     type getargs >/dev/null 2>&1 || . /lib/dracut-lib.sh
 
-    #cpumask="$(getargs non_iso_cpumask)"
-    cpumask='$non_iso_cpumask'
+    #cpumask="$(getargs reserved_cpumask)"
```
cynepco3hahue: I think you can remove this line

build/assets/tuned/openshift-node-real-time-kernel
```diff
@@ -19,4 +19,4 @@ kernel.timer_migration = 0
 /sys/devices/system/machinecheck/machinecheck*/ignore_ce = 1   
 
 [scheduler]
-isolated_cores={{.IsolatedCpus}}
+{{if .IsolatedCpus}}isolated_cores={{.IsolatedCpus}}{{end}}
```
cynepco3hahue: I think we should put the whole `[scheduler]` section under `if` condition

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -163,9 +168,9 @@ func getIgnitionConfig(assetsDir string, profile *performancev1alpha1.Performanc
 		})
 	}
 
-	nonIsolatedCpus := profile.Spec.CPU.NonIsolated
+	ReservedCpus := profile.Spec.CPU.Reserved
```
cynepco3hahue: ReservedCpus -> reservedCpus
true for all occurences

deploy/crds/performance.openshift.io_performanceprofiles_crd.yaml
```diff
@@ -34,15 +34,23 @@ spec:
...
+                  description: EnableIsolcpus results in the Isolated CPUset being
+                    listed in the isolcpu kernel arg on the host machines. When this
+                    option is set to "false", the Isolated cpus will be partitioned
+                    by tuned, but load balancing across the Isolated cpus will still
+                    be possible. When this option is set to "true", the Isolated CPU
```
yanirq: @davidvossel   In case we had the full cpu_partitioning profile support from NTO which also includes: no_balance_cores that disable the kernel load balancing in certain isolated CPUs, could we then say that this boolean will becomes redundant? (see [here](https://github.com/redhat-performance/tuned/blob/bfe0eb91949d4e305de5c22bec913fc8318b3a33/man/tuned-profiles-cpu-partitioning.7#L55))

deploy/crds/performance.openshift.io_performanceprofiles_crd.yaml
yanirq: @davidvossel is right regarding the difference of isolated_cores used in kargs and isolated_cores passed to the [tuned profie](https://github.com/openshift-kni/performance-addon-operators/blob/832b615823bd28807d0eef2748d808739f7b266e/build/assets/tuned/openshift-node-real-time-kernel#L22). 

> 2. In a MachineConfig to set the IRQBALANCE_BANNED_CPUS=11110000 bitmask for irqbalance.

the pre tuning script we have right now takes the reserved cpus list and inverts it to an isolated_cpumask.
On the other hand we [set](https://github.com/openshift-kni/performance-addon-operators/blob/832b615823bd28807d0eef2748d808739f7b266e/build/assets/scripts/pre-boot-tuning.sh#L37) CPUAffinity from the reserved cpu list.

Just as a reference, Tuned cpu-partitioning profile takes a list of isolated cpus and inverts them to non_isolated_cpus [list](https://github.com/redhat-performance/tuned/blob/bfe0eb91949d4e305de5c22bec913fc8318b3a33/profiles/cpu-partitioning/tuned.conf#L25) (or [mask](https://github.com/redhat-performance/tuned/blob/bfe0eb91949d4e305de5c22bec913fc8318b3a33/profiles/cpu-partitioning/tuned.conf#L28)) for setting CPUAffinity.  Supported py scripts for that can be found [here](https://github.com/redhat-performance/tuned/blob/master/tuned/profiles/functions/function_cpulist_invert.py).

deploy/crds/performance.openshift.io_performanceprofiles_crd.yaml
MarSik: Actually, this is pretty good reason to keep the argument. We can then always pass `isolated` to isolated_cores=<CPU\-LIST> and iff the boolean is set then set the same list to no_balance_cores as well (isolcpus for now).

build/assets/tuned/openshift-node-real-time-kernel
MarSik: And add the no_balance_cores={{. IsolatedCpus}} conditional of .enableIsolcpus (which we should rename)

MarSik: @davidvossel This was basically my only comment now. Lets keep the boolean, just change the name to something like `balance_isolated_cores` that defaults to true. And add no_balance_cores={{. IsolatedCpus}} when the balancing is disabled.

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
MarSik: I think I fixed this one for David.

build/assets/scripts/pre-boot-tuning.sh
davidvossel: fixed

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
davidvossel: yup, it's fixed, thanks @MarSik

build/assets/tuned/openshift-node-real-time-kernel
davidvossel: I fixed this by calling the boolean `balanceIsolated=[true|false]`. The no_balanace_cores is something I think we should follow up on in another pr as there's probably some discussion around how this option related to isolcpu.

davidvossel:
@MarSik @yanirq @cynepco3hahue @fromanirh 

I updated the PR. The big change here is the "enableIsolcpu" boolean has changed to "balanceIsolated". The default for balanceIsolated=true. Functionally it works just like the enableIsolcpu=false.

slintes:
As far as I understand this... ;)
/lgtm

Not sure if others want to review again, so putting a hold on it. Feel free to remove.
/hold

pkg/controller/performanceprofile/components/machineconfig/machineconfig_test.go
```diff
@@ -93,13 +93,37 @@ const expectedBootArguments = `
...
+
 var _ = Describe("Machine Config", func() {
 	It("should generate yaml with expected parameters", func() {
 		profile := testutils.NewPerformanceProfile("test")
 		profile.Spec.HugePages.Pages = append(profile.Spec.HugePages.Pages, performancev1alpha1.HugePage{
 			Count: 1024,
 			Size:  "2M",
 		})
+		f := false
+		profile.Spec.CPU.BalanceIsolated = &f
```
cynepco3hahue: please use `pointer` module for it

```go
import "k8s.io/utils/pointer"
...
profile.Spec.CPU.BalanceIsolated = pointer.BoolPtr(false)
```

deploy/crds/performance.openshift.io_performanceprofiles_crd.yaml
```diff
@@ -34,15 +34,22 @@ spec:
...
+                balanceIsolated:
+                  description: BalanceIsolated toggles whether or not the Isolated
+                    CPU set is eligible for load balancing work loads. When this option
+                    is set to "false", the Isolated CPU set will be static, meaning
+                    workloads have to explicitly assign each thread to a specific
+                    cpu in order to work across multiple CPUs. Setting this to "true"
+                    allows workloads to be balanced across CPUs. Setting this to "false"
+                    offers the most predictable performance for guaranteed workloads,
+                    but it offloads the complexity of cpu load balancing to the application.
+                    Defaults to "true"
```
cynepco3hahue: Should it default to false, to preserve the previous behavior?

MarSik: Nobody uses it yet and the previous behavior was actually one of the issues. I would prefer enabled balancing by default myself.

MarSik:
> I still unsure if we need `balanceIsolated` parameter needed at all, but I will not block this PR because of it.

Yep, we do. Tuned profiles support that option as well.

cynepco3hahue:
> Yep, we do. Tuned profiles support that option as well.

@MarSik Can you please give the use case, because, for the DPDK workload you want to have isolated CPU, and you do not want that some process will be balanced on top of it.

PR #74, opened 2020-02-06
csvgen: extend to accomodate OCP needs

fromanirh:
This patch extends the CSV generator tooling to allow
it to be used to generate the OCP manifests seamlessly.
Few changes are needed:
- allow to inject arbitrary key/value annotations from
  an external file

the changes are meant to be optional, and they are meant to
not disrupt the existing flows.


slintes:
The only question I have is about the CSV's major.minor only version, as discussed in chat. Feel free to remove the hold after confirming that this is the way to go for Openshift operators.

/hold
/lgtm

fromanirh:
> The only question I have is about the CSV's major.minor only version, as discussed in chat. Feel free to remove the hold after confirming that this is the way to go for Openshift operators.
> 
> /hold
> /lgtm

ack, doublechecking

fromanirh:
turns out that it is possible to use the full version ("4.4.0") and we prefer it that way. PR updated and simplfied accordingly.

fromanirh:
this PR should not disrupt the existing flows, and the ocp changes are moving forward, hance I think we can merge it.

PR #77, opened 2020-02-10
Fix pre-boot tuning to check latest boot entry

yanirq:
**What was the problem**
Pre boot tuning checked if the custom initrd image was applied on all of the bootloader entries instead of only checking the current(latest) version.

**Fix**
Pre boot tuning was fixed to check if the latest boot-loader entry has the custom initrd image.
Applying new entries was adjust as well to match the current entry file.


cynepco3hahue:
lgtm, let's wait for CI

cynepco3hahue:
@yanirq CI is failing :(

PR #79, opened 2020-02-12
Added unit tests for profile updates

slintes:
Added unit tests for profile updates.


slintes:
wait for #73 

/hold

slintes:
rebased and improved test for updating hugepages with numa node

slintes:
modified after recent API changes, ready for review :)

/hold cancel

pkg/controller/performanceprofile/performanceprofile_controller_test.go
```diff
@@ -167,24 +160,14 @@ var _ = Describe("Controller", func() {
...
 			Expect(err).ToNot(HaveOccurred())
 		})
 
-		It("should do nothing on third reconcile loop", func() {
+		It("should create event on third reconcile loop", func() {
 			r := newFakeReconciler(profile)
 
-			result, err := r.Reconcile(request)
-			Expect(err).ToNot(HaveOccurred())
-			Expect(result).To(Equal(reconcile.Result{RequeueAfter: 10 * time.Second}))
```
davidvossel: right here we're checking that the first reconcile loop results in a requeue, and that the others do not. How is that handled now with the new "reconcileTimes()" function?

slintes: the requeue should always happen on the 1st reconcile loop, and that one is tested in line 112

pkg/controller/performanceprofile/performanceprofile_controller_test.go
```diff
@@ -101,16 +102,14 @@ var _ = Describe("Controller", func() {
...
 			Expect(errors.IsNotFound(err)).To(Equal(true))
 		})
 
 		It("should create all resources except KubeletConfig on first reconcile loop", func() {
 			r := newFakeReconciler(profile)
 
-			result, err := r.Reconcile(request)
-			Expect(err).ToNot(HaveOccurred())
-			Expect(result).To(Equal(reconcile.Result{RequeueAfter: 10 * time.Second}))
+			Expect(reconcileTimes(r, request, 1)).To(Equal(reconcile.Result{RequeueAfter: 10 * time.Second}))
```
slintes: @davidvossel see here

PR #80, opened 2020-02-12
Adjust pre boot tunings

yanirq:
Pre boot tuning was changed to operate as following:
- Compute non_isolated_cpus_mask and its invert in the operator code
- Set kernel argument for tuned.non_isolcpus=<non isolated cpus mask> to set `/sys/devices/virtual/workqueue/cpumask` using RHCOS hooks in early boot stage
- openshift-node-real-time-kernel will set cpumask for `/sys/bus/workqueue/devices/writeback/cpumask`. (Tuned)
- Add to ignition file a custom CPUAffinity file : `/etc/systemd/system.conf.d/setAffinity.conf`
- Set systemd with the custom file:
  `rpm-ostree initramfs --enable --arg=-I --arg="/etc/systemd/system.conf /etc/systemd/system.conf.d/setAffinity.conf"`
- Set irqbalance in the script (no need to set it in initramfs)

**Additional Info**
The operator takes the NON_ISOLATED_CPUS (=reserved cpus) parameter and creates from that:
- CPUAffinity (encapsulated in the setAffinity.conf file passed to early tuning script
- NON_ISOLATED_CPU_MASK_INVERT environment variable for the pre-boot script to consume (for irqbalance setting)
- Calculates NotIsolatedCpumask passed to tuned (openshift-node-real-time-kernel profile)

PR #84, opened 2020-02-17
Remove nonIsolated from test config

davidvossel:
The nonIsolated option no longer exists and shouldn't be referenced in the deployment config


yanirq:
@davidvossel  @cynepco3hahue  added that fix on https://github.com/openshift-kni/performance-addon-operators/pull/80 as well + balanceIsolated

slintes:
done in https://github.com/openshift-kni/performance-addon-operators/pull/80

/close

PR #87, opened 2020-02-18
Don't use hardcoded CI registry URL

slintes:
Registry URL is about to change, use env vars for it.
See https://github.com/openshift/ci-tools/blob/master/TEMPLATES.md#image_format

This, unfortunately, means we can't use the registry image in CI, because we don't have
a good env var available in the image build step, so we can't set the operator image
in the registry image. Alternatively, we deploy the CRDs and CSV directly now in CI.

For manual tests with a registry image, you can override FEATURES_ENVIRONMENT to
"other-clusters" when deploying.


slintes:
didn't work
debugging env now

/hold

slintes:
So... the registry image in `deploy.sh` works now.
But the CSV in that image does not have the correct operator image. AFAIU The `IMAGE_FORMAT` env var is not available during the image build step (in `Dockerfile.registry.build`).
Looking through all CI jobs, there are only very few projects besides us using a registry image (still with the hardcoded URL). I looked at another project, they install the CSV directly (without registry image), that way they can set the correct image during test time where the env var is available.
I'm wondering if we should do the same for the operator's e2e test?
@davidvossel @cynepco3hahue wdyt?

cynepco3hahue:
> So... the registry image in `deploy.sh` works now.
> But the CSV in that image does not have the correct operator image. AFAIU The `IMAGE_FORMAT` env var is not available during the image build step (in `Dockerfile.registry.build`).
> Looking through all CI jobs, there are only very few projects besides us using a registry image (still with the hardcoded URL). I looked at another project, they install the CSV directly (without registry image), that way they can set the correct image during test time where the env var is available.
> I'm wondering if we should do the same for the operator's e2e test?
> @davidvossel @cynepco3hahue wdyt?

By directly do you mean, just applying CSV without catalog and subscription?

slintes:
> By directly do you mean, just applying CSV without catalog and subscription?

yes

cynepco3hahue:
I think we can change it only for CI deployment.

slintes:
This is escalating a bit... pushed a new version and updated description.
Let's see if it works in CI...

slintes:
reverted the change for upstream image for being prepared for branches, there is no OPENSHIFT_BUILD_REFERENCE. Will figure out in a follow up how to dealt with it, don't want to block this one for it.

cluster-setup/other-clusters/performance/performance_profile.patch.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: performance.openshift.io/v1alpha1
+kind: PerformanceProfile
+metadata:
+  name: ci
+spec:
+  cpu:
+    isolated: "1-3"
```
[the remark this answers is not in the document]
slintes: damn, I probably copied it before rebasing, thanks

davidvossel:
> This, unfortunately, means we can't use the registry image in CI, because we don't have
a good env var available in the image build step, so we can't set the operator image
in the registry image. Alternatively, we deploy the CRDs and CSV directly now in CI.

@slintes Will the OPENSHIFT_BUILD_NAMESPACE env var no longer be supported?

hack/deploy.sh
```diff
@@ -5,19 +5,24 @@ set -e
 # expect oc to be in PATH by default
 OC_TOOL="${OC_TOOL:-oc}"
 
-# Override the image name when this is invoked from openshift ci
-if [ -n "${OPENSHIFT_BUILD_NAMESPACE}" ]; then
-  FULL_REGISTRY_IMAGE="registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:performance-addon-operator-registry"
+# Override the image name in the CSV when this is invoked from openshift ci
+# See https://github.com/openshift/ci-tools/blob/master/TEMPLATES.md#image_format
+# Fallback to string "REPLACE_IMAGE" what we use at other places as well
+CI_OPERATOR_IMAGE=${IMAGE_FORMAT/'${component}'/performance-addon-operator}
```
davidvossel: can't we check if image format exists, and make that the deciding factor for how we deploy or what image to use?

```
if [ -n "${IMAGE_FORMAT}" ]; then
    echo "Openshift CI detected, using 'ci-cluster' environment"
    CI_OPERATOR_IMAGE=${IMAGE_FORMAT/'${component}'/performance-addon-operator}
    FEATURES_ENVIRONMENT="ci-cluster"
fi

```

davidvossel:
> Can we get some more clarification on why this change is necessary?

I get the context now. 

Our easiest path forward is going to be to react to the registry change once openshift ci makes the change. We'll need to work with them to pass in the registry as an env var we can key off of in the future.

slintes:
> > This, unfortunately, means we can't use the registry image in CI, because we don't have
> > a good env var available in the image build step, so we can't set the operator image
> > in the registry image. Alternatively, we deploy the CRDs and CSV directly now in CI.
> 
> @slintes Will the OPENSHIFT_BUILD_NAMESPACE env var no longer be supported?

that's unclear to me. At least for just detecting CI they will introduce a new var, see https://issues.redhat.com/browse/DPTP-893

slintes:
> I'm really hoping we can avoid segmenting installation methods. Openshift CI needs to match how we're deploying this locally and how end users will consume this.

In a first step I was about just removing catalogsource etc. from the manifests. For being able to do quick tests manually I went with the 2nd override dir then. I'm also not very happy with this, but I don't have a better idea.
End users should look at the deploy repo. There we run periodic CI with registry image and catalogsource / image, so we are fine IMHO?

hack/deploy.sh
slintes: done

slintes:
> > Can we get some more clarification on why this change is necessary?
> 
> I get the context now.
> 
> Our easiest path forward is going to be to react to the registry change once openshift ci makes the change. We'll need to work with them to pass in the registry as an env var we can key off of in the future.

Are you suggesting to postpone this PR until CI actually changes and we have a better picture on what to expect? Works for me.

cynepco3hahue:
@slintes looks like we need to process with your PR because we can not rely on `registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:performance-addon-operator-registry`

cynepco3hahue:
@slintes I can not see the performance profile operator pod at all under the CI, and it hard to say why, because we do not have CSV dump under the CI :(

slintes:
@cynepco3hahue oh indeed, I see the internal image registry changed. Will investigate how to get the correct URL in the registry build step, it wasn't possible when I started with this PR.

However, because of the latter I already switched to installing the CSV directly now. In the events log I see it's complaining about missing operatorgroup. Will fix this first.

slintes:
Deployment was succesfull after last push.
But re-reading the relevant mail thread about this issue, I understand using the new CI image registry URL hardcoded is fine for now (done in #113), and there will be a new method of testing optional operators later. Since using the registry image is also way easier than installing everything manually, I'm closing this one.

/close

PR #88, opened 2020-02-19
Fix for CPU management test

dshchedr:


vladikr:
Looks good to me. Let's wait for the CI

cluster-setup/ci-cluster/performance/performance_profile.patch.yaml
```diff
@@ -4,7 +4,6 @@ metadata:
   name: ci
 spec:
   cpu:
-    balanceIsolated: false
```
slintes: Is this on purpose? Changing the profile for fixing a test feels wrong, at least without looking into the details ;)

dshchedr: Yeah, that is the reason why i updated test - @yanirq mentioned that balanceIsolated should be True by default

vladikr: Well, we have two scenarios to support now...

dshchedr: yes, it should work in both scenarios (when balance is true and false), but i'm not sure we really need to run both of them in our ci, since it may take a lot of time on mco updating.

yanirq: when and if we have multiple profiles support we might be able to test both scenarios. 
make sure you also revert https://github.com/openshift-kni/cnf-features-deploy/pull/86 when this is merged. we can also keep the profile in cnf-features-deploy as it is as well.

dshchedr: good point, will do

PR #89, opened 2020-02-20
Fix Cpumask for workqueues

yanirq:
Fixed:
- Workqueues are set with cpumasks using delimiter
- Workqueues mask upper bound changed from 64 to 256 cpus.


fromanirh:
/hold
remove when others reviews are done

cynepco3hahue:
let's wait for CI

PR #90, opened 2020-02-21
new validation tests for CR/CRD schema

jeremyary:
main commit:
- add test to validate CRD file against PerformanceProfile struct
- add test to validate CR against CRD schema

other assist commits (separated for readability):
- add .idea (IntelliJ)
- add new dependencies for main commit, also includes results from make deps-update (master has outstanding without any new changes)


pkg/utils/validation/validation_suite_test.go
```diff
@@ -1,13 +1,13 @@
...
+	"testing"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-
-	"testing"
 )
 
-func TestTable(t *testing.T) {
+func TestFeatureGate(t *testing.T) {
```
slintes: Something is strange here. It looks like you moved a file from vendor to here.
Also, please use a better name for this function, best practice IIRC is Test<PackageName>.

jeremyary: No problem! I'll get on suggestions, agree across the board and just needed some nudges on what maintainers would prefer here and there. I'll also get that git history confusion cleaned up, not sure what caused that in a new file addition :\

davidvossel:
great contribution :+1: thanks!

jeremyary:
I've tried a few different means of fixing how git is viewing that odd history on what's a newly created file and that's worrying to me, so I'm just going to take a fresh copy and try again with a new PR.

PR #91, opened 2020-02-21
[release-4.4] Use correct operator tag in upstream registry image

slintes:
No way to detect branch when building the image, so we need to hardcode the tag for now


slintes:
/hold

retrigger tests when https://github.com/openshift/release/pull/7277 is merged (running rehearsal right now)

slintes:
CI cluster seems to be broken... but this change is unrelated to any tests, so let's get this in.

/override e2e-gcp

Issue #92, opened 2020-02-21
Operator pod in CrashLoopBackOff with segfault in runtime.go


cwilkers:
Output follows:

```I0221 22:59:14.535851       1 main.go:47] Operator Version: 0.0.1
I0221 22:59:14.535926       1 main.go:48] Go Version: go1.13.5
I0221 22:59:14.535931       1 main.go:49] Go OS/Arch: linux/amd64
I0221 22:59:14.535945       1 main.go:50] Version of operator-sdk: v0.13.0
I0221 22:59:16.643037       1 main.go:98] Registering Components.
E0221 22:59:23.005157       1 main.go:142] Could not create ServiceMonitor object: servicemonitors.monitoring.coreos.com "performance-operator-metrics" already exists
I0221 22:59:23.005223       1 main.go:150] Starting the Cmd.
I0221 22:59:23.005333       1 leaderelection.go:242] attempting to acquire leader lease  openshift-performance-addon/performance-addon-operators...
I0221 22:59:40.625216       1 leaderelection.go:252] successfully acquired lease openshift-performance-addon/performance-addon-operators
I0221 22:59:41.829084       1 performanceprofile_controller.go:199] Reconciling PerformanceProfile
E0221 22:59:41.829373       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 1225 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x14bf9a0, 0x23b78f0)
	/go/src/github.com/openshift-kni/performance-addon-operators/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0xa3
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/src/github.com/openshift-kni/performance-addon-operators/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x82
panic(0x14bf9a0, 0x23b78f0)
	/root/.gimme/versions/go1.13.5.linux.amd64/src/runtime/panic.go:679 +0x1b2
github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/machineconfig.getIgnitionConfig(0x1689e3a, 0x7, 0xc00094edc0, 0x1687e56, 0x1, 0x168804e)
[...]

PR #93, opened 2020-02-23
Add nil check before creating preboot ignition

yanirq:
**What is the Probelm?**
When cpu.reserved field is missing in perfomance profile, a nil panic will occur:
`bserved a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)`

**What was fixed**
Added a nil check before creating ignition content for pre boot tuning that relies on cpu.reserved value

**Note**
Another way to ensure proper behaviour here is to validate that the field exists if its defined as a mandatory field


cynepco3hahue:
@yanirq Please unhold once the CI passed.

slintes:
Fixes #92 

/cherrypick release-4.4

Issue #82, opened 2020-02-14
Event creation fails


slintes:
The operator fails to create events.

```
E0214 13:38:27.451724       1 event.go:263] Server rejected event [...]
events is forbidden: User "system:serviceaccount:openshift-performance-addon:performance-operator" cannot create resource "events" in API group "" in the namespace "default"
```

I think this happens because we create the event for the profile resource, which isn't namespaced, which results in the event being created in the `default` namespace. We probably need to create events in the operator's namespace.

slintes:
fixed by #94 

/close

PR #94, opened 2020-02-23
Fix events permissions

cynepco3hahue:
Performance profile is cluster scope resource, so events creation should be done on the cluster level.

### DIFF
--- a/deploy/olm-catalog/performance-addon-operator/0.0.1/performance-addon-operator.v0.0.1.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/performance-addon-operator/0.0.1/performance-addon-operator.v0.0.1.clusterserviceversion.yaml
@@ -35,6 +35,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
+    containerImage: REPLACE_IMAGE
   name: performance-addon-operator.v0.0.1
   namespace: placeholder
 spec:
@@ -54,6 +55,12 @@ spec:
     spec:
       clusterPermissions:
       - rules:
+        - apiGroups:
+          - ""
+          resources:
+          - events
+          verbs:
+          - '*'
         - apiGroups:
           - performance.openshift.io
           resources:
@@ -124,7 +131,6 @@ spec:
           - services
           - services/finalizers
           - configmaps
-          - events
           verbs:
           - '*'
         - apiGroups:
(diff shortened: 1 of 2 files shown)


slintes:
Fixes #82 

tbh I'm not sure if we want to fix this issue this way. I would prefer to write events into the operator's namespace, so we don't need cluster permission for it. Wdyt @cynepco3hahue @davidvossel ?

cynepco3hahue:
@slintes I do not think you have a choice, because performance profile is a cluster-wide resource and you pass the instance of performance profile to the recorder, maybe another option is to write events under the operator deployment, but I unsure if it will have enough visibility for a user.

cynepco3hahue:
hm, maybe an additional option is to copy instance, set operator namespace and pass it to the recorder, but to me, it looks like a hacky way

davidvossel:
> bh I'm not sure if we want to fix this issue this way. I would prefer to write events into the operator's namespace, so we don't need cluster permission for it. Wdyt @cynepco3hahue @davidvossel ?

Since the resource is cluster scoped, this might be unexpected.

FOr example, nodes are cluster scoped, and the events aren't tied to a resource bound to a namespace.

deploy/olm-catalog/performance-addon-operator/0.0.1/performance-addon-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -35,6 +35,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
+    containerImage: REPLACE_IMAGE
```
davidvossel: weird that this got added. happened to me locally as well when I tried it. There's supposed to be a ci check to ensure we can't merge unless this csv is synced. 

i'll look into it. it's not related to this pr

slintes:
> > bh I'm not sure if we want to fix this issue this way. I would prefer to write events into the operator's namespace, so we don't need cluster permission for it. Wdyt @cynepco3hahue @davidvossel ?
> 
> Since the resource is cluster scoped, this might be unexpected.
> 
> FOr example, nodes are cluster scoped, and the events aren't tied to a resource bound to a namespace.

@davidvossel I don't have a cluster to check, but when events for cluster scoped resources are always created in the default namespace, I'm fine with it of course. However, wouldn't it be enough to use a namespaced permission then, instead of cluster permission like we have now?

slintes:
> wouldn't it be enough to use a namespaced permission then, instead of cluster permission like we have now

probably not even possible with CSVs for other namespaces than you own...

/cherrypick release-4.4

PR #104, opened 2020-02-25
Reboot from rpm-ostree injection in pre-boot

yanirq:
**What is the problem**
There are inconsistent results of rpm-ostree status after pre-boot service have run initramfs injection: `rpm-ostree initramfs --enable --arg=-I --arg="/etc/systemd/system.conf /etc/systemd/system.conf.d/setAffinity.conf" `
The expectation is that the deployment will contain these files:
`Initramfs: -I /etc/systemd/system.conf /etc/systemd/system.conf.d/setAffinity.conf `  (can be viewed by running `rpm-ostree status -b` after a successful deployment)
Its seems that the reason for that might be connected to out-of-sync issues or race conditions.

**What was fixed**
- Pre boot script was changed to initiate a reboot directly from rpm-ostree (using -r flag) to try and be in sync.
- Reboot service was removed since it is not needed anymore (reboot is done in pre-boot script as mentioned above)
- CI test added for checking rpm-ostree status after deployment

### DIFF
--- a/build/assets/scripts/pre-boot-tuning.sh
+++ b/build/assets/scripts/pre-boot-tuning.sh
@@ -5,7 +5,7 @@ set -euo pipefail
 SYSTEM_CONFIG_FILE="/etc/systemd/system.conf"
 SYSTEM_CONFIG_CUSTOM_FILE="/etc/systemd/system.conf.d/setAffinity.conf"
 
-if [ -f /etc/sysconfig/irqbalance ] && [ -f ${SYSTEM_CONFIG_CUSTOM_FILE} ] && egrep -wq "^IRQBALANCE_BANNED_CPUS=${RESERVED_CPU_MASK_INVERT}" /etc/sysconfig/irqbalance; then
+if [ -f /etc/sysconfig/irqbalance ] && [ -f ${SYSTEM_CONFIG_CUSTOM_FILE} ] && rpm-ostree status -b | grep -q -e '-I /etc/systemd/system.conf /etc/systemd/system.conf.d/setAffinity.conf' && egrep -wq "^IRQBALANCE_BANNED_CPUS=${RESERVED_CPU_MASK_INVERT}" /etc/sysconfig/irqbalance; then
     echo "Pre boot tuning configuration already applied"
     echo "Setting kernel rcuo* threads to the housekeeping cpus"
     pgrep rcuo* | while read line; do taskset -pc ${RESERVED_CPUS} $line || true; done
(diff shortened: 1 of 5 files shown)


yanirq:
Added a check in pre-boot script for rpm-ostree status, this way if the injection wasnt complete another try will be made. tested successfully also on a local environment
/cc @MarSik  @cynepco3hahue

cynepco3hahue:
@yanirq Did you have a chance to check how many reboots it initiates?

build/assets/scripts/pre-boot-tuning.sh
```diff
@@ -5,7 +5,7 @@ set -euo pipefail
 SYSTEM_CONFIG_FILE="/etc/systemd/system.conf"
 SYSTEM_CONFIG_CUSTOM_FILE="/etc/systemd/system.conf.d/setAffinity.conf"
 
-if [ -f /etc/sysconfig/irqbalance ] && [ -f ${SYSTEM_CONFIG_CUSTOM_FILE} ] && grep -ls "IRQBALANCE_BANNED_CPUS=${RESERVED_CPU_MASK_INVERT}" /etc/sysconfig/irqbalance; then
+if [ -f /etc/sysconfig/irqbalance ] && [ -f ${SYSTEM_CONFIG_CUSTOM_FILE} ] && rpm-ostree status -b | grep -q -e '-I /etc/systemd/system.conf /etc/systemd/system.conf.d/setAffinity.conf' && egrep -wq "^IRQBALANCE_BANNED_CPUS=${RESERVED_CPU_MASK_INVERT}" /etc/sysconfig/irqbalance; then
```
slintes: Can you use the vars for the filenames here as well?

build/assets/scripts/pre-boot-tuning.sh
```diff
@@ -21,7 +21,10 @@ else
         echo "IRQBALANCE_BANNED_CPUS=${RESERVED_CPU_MASK_INVERT}" >>/etc/sysconfig/irqbalance
     fi
 
-    rpm-ostree initramfs --enable --arg=-I --arg="${SYSTEM_CONFIG_FILE} ${SYSTEM_CONFIG_CUSTOM_FILE}" 
+    if [[ ! -f /var/restarts ]]; then
```
slintes: What is this restart count for?

build/assets/scripts/pre-boot-tuning.sh
yanirq: I could but I think its more readable this way and the files are constant.

build/assets/scripts/pre-boot-tuning.sh
yanirq: Since we need to add a retry mechanism here in case initramfs will be out of sync (can happen with MCDs) I added a retry=restart count for testing purposes and debugging

build/assets/scripts/pre-boot-tuning.sh
slintes: Sorry but I disagree. Why introducing vars if they aren't used everywhere?

build/assets/scripts/pre-boot-tuning.sh
slintes: When this is not used, I prefer to remove it. You can see reboots in nodes' journal if needed

slintes:
Thanks!

/lgtm

Waiting for e2e
/hold

PR #105, opened 2020-02-25
fix lingering mod vendor issues in master via deps-update

jeremyary:
A previous commit included what I suspect to be manual insertion of vendor files, this commit is purely the result of `make deps-update` against `master` to clean up and make way for future PR's.


slintes:
Thanks!

/ok-to-test
/approve

PR #107, opened 2020-02-26
CR(D) Validation tests (Take 2)

jeremyary:
With deps issues in master cleaned up, file history seems correct now. Also addresses the following requested changes from previous [PR #90](https://github.com/openshift-kni/performance-addon-operators/pull/90):

- test assist functions moved into test file instead of separate file
- moved to performance pkg in lieu of a new package containing only unit tests
- renamed suite (and test files) to match package name
- more accurate 'It ___' test names & comments on what validation actually means
  - also raised [operator-utils issue](https://github.com/RHsyseng/operator-utils/issues/43) to address this


slintes:
Oh, github approve is lgtm + approved for prow, interesting.
Wanted to have a look at e2e test before merging

/hold

jeremyary:
> Wanted to have a look at e2e test before merging

last PR was an AWS region resource issue, are you suspecting the same here or wanting to look into something else? Just asking so I can learn more on project e2e and if there's more I can do locally to prep commits... I've got a lab, but my Inf guys gets itchy when I mention running functests against it ^_^

slintes:
> > Wanted to have a look at e2e test before merging
> 
> last PR was an AWS region resource issue, are you suspecting the same here or wanting to look into something else? Just asking so I can learn more on project e2e and if there's more I can do locally to prep commits... I've got a lab, but my Inf guy gets itchy when I mention running functests against it ^_^

I'm not really suspecting something, but some dependencies were updated, so I just want to be on the safe side :) And because the e2e lane isn't mandatory (fails too often during cluster creation already), I put the hold to prevent merging before it runs. Btw, we run on GCP, because the RT kernel does not work on (every) EC2 VM.

slintes:
all green :)

/hold cancel

PR #113, opened 2020-03-02
Updated CI image registry URL

slintes:
WIP

Checking if we have the IMAGE_FORMAT env var available in the image build step now.
Otherwise maybe just updated hardcoded URL.
Or get rid of registry image like in #87


slintes:
Deployment with the new hardcoded URL succeeded. But you can also see that there is not env var for it yet. Will merge nevertheless, in order to unblock CI. 
Alternative solution can be further discussed in #87 

/hold cancel
/cherrypick release-4.4

PR #117, opened 2020-03-03
Add options to configure KubeletConfig via PerformanceProfile CR

cwilkers:


cluster-setup/base/performance/performance_profile.yaml
```diff
@@ -10,3 +10,5 @@ spec:
     defaultHugepagesSize: "1G"
   realTimeKernel:
     enabled: true
+  kubeletConfig:
+    topologyManager: "single-numa-node"
```
MarSik: I consider kubeletConfig and topologyManager to be implementation details. So I would change the structure and naming to something like:

```
numa:
  default-policy: "single-numa-node"
```

This would allow us to change the underlying implementation in the future. For example, our plan is to see if per-pod policy can be implemented. My naming (although probably not perfect) would still make sense in that case.

slintes: +1
Please don't expose the complete KubeletConfig. If you need all of it, you can use it directly, the performance operator wouldn't add any value then IMHO

cwilkers: I'm not fully comfortable with "default-policy". What if we still include the word "topology"? 

It could be either

```
numa:
    topology-policy: (none|best-effort|single-numa-node|restricted)
```

or perhaps:

```
topology:
    policy: (none|best-effort|single-numa-node|restricted)
```

WDYT?

cynepco3hahue:
@cwilkers Can you please add unit tests or at least extend some of existing?

pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig_test.go
```diff
@@ -25,7 +25,7 @@ var _ = Describe("Kubelet Config", func() {
 		selectorKey, selectorValue := components.GetFirstKeyAndValue(profile.Spec.MachineConfigPoolSelector)
 		Expect(manifest).To(ContainSubstring(fmt.Sprintf("%s: %s", selectorKey, selectorValue)))
 		Expect(manifest).To(ContainSubstring("reservedSystemCPUs: 0-3"))
-		Expect(manifest).To(ContainSubstring("topologyManagerPolicy: best-effort"))
+		Expect(manifest).To(ContainSubstring("topologyManagerPolicy: single-numa-node"))
```
slintes: I think to need to set it here: https://github.com/openshift-kni/performance-addon-operators/blob/master/pkg/utils/testing/testing.go#L30

pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
```diff
@@ -44,6 +44,12 @@ func New(profile *performancev1alpha1.PerformanceProfile) (*machineconfigv1.Kube
 		kubeletConfig.ReservedSystemCPUs = string(*profile.Spec.CPU.Reserved)
 	}
 
+	if profile.Spec.NUMA != nil {
+		if profile.Spec.NUMA.TopologyPolicy != nil {
+			kubeletConfig.TopologyManagerPolicy = string(*profile.Spec.NUMA.TopologyPolicy)
```
slintes: I understand there are some specific values allowed for this, right? We should validate them in https://github.com/openshift-kni/performance-addon-operators/blob/master/pkg/controller/performanceprofile/components/profile/profile.go#L20

cluster-setup/base/performance/performance_profile.yaml
MarSik: numa.topology-policy: sounds good to me.

pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
cwilkers: Validations added

pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig_test.go
cwilkers: I've added some logic to testing functions.

slintes:
I think we are in good shape now :+1: 
Let's see what CI think about it.

/retest

pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
MarSik: I wonder if kubelet has an enum somewhere. Just to prevent us forgeting to adapt in case they add something new

MarSik:
Apart from the enum question, this looks good to me.

pkg/controller/performanceprofile/components/profile/profile.go
```diff
@@ -10,6 +10,10 @@ import (
 const (
 	hugepagesSize2M = "2M"
 	hugepagesSize1G = "1G"
+	topologyPolicyNone = "none"
```
slintes: Sory, missed it, there are constants already available for these values.
See https://github.com/openshift-kni/performance-addon-operators/blob/master/vendor/k8s.io/kubelet/config/v1beta1/types.go#L59
and example usage in our functests: https://github.com/openshift-kni/performance-addon-operators/blob/master/functests/performance/topology_manager.go#L47

pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
slintes: Golang doesn't have enums :man_shrugging: 
But there are constants available at least, see other comment.

slintes:
@cwilkers can you please run `gofmt -s -w `find . -path ./vendor -prune -o -type f -name '*.go' -print`` and commit changes? There seems to be a formatting issue.

If you like to rebase, you can just run `make generate` instead afterwards, it was modified after you branched off.

That will make `ci` test happy. Thanks :)

cwilkers:
Okay, I've rebased, squashed, and re-run generate and generate-latest-dev-csv.

slintes:
@MarSik @cwilkers is this supposed to be backported?

slintes:
@cwilkers please remove the WIP :)

PR #119, opened 2020-03-04
Finalize tuning parameters and values

yanirq:
- Remove unneeded kernel argument irqaffinity=0
- Added rcu_nocbs=<isolated cpus>, nohz_full=<isolated_cpus>
- Since rcu_nocbs is added , removed tasket -pc <house keeping cpu> on the rcu threads from the pre-boot script.

> Note: Regardless of BalanceIsolated=true/false, we should be consistent in tuning rcu_nocbs, nohz_full, and tuned.non_isolcpus.  These are always the “houskeeping” cpus.  BalanceIsolated=false should only change 1 thing: adding isolcpus = cpus that are not housekeeping.
This way we always have housekeeping cpus, where by default all systemd services will run, and the rest of the cpus are “low latency”, with “extra-low-latency” when BalanceIsolated=false.


pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -130,7 +125,11 @@ func getKernelArgs(hugePages *performancev1alpha1.HugePages, isolatedCPUs *perfo
 	}
 
 	if isolatedCPUs != nil {
-		kargs = append(kargs, fmt.Sprintf("isolcpus=%s", string(*isolatedCPUs)))
+		if balanceIsolated != nil && *balanceIsolated == false {
+			kargs = append(kargs, fmt.Sprintf("isolcpus=%s", string(*isolatedCPUs)))
```
slintes: nit: string(...) is not needed, Sprintf + %s does the trick without explicit cast. But I see it is done at other places as well, so let's do a cleanup after this PR.

yanirq:
Verifying if poll and intel_pstate must be removed or not
/hold

slintes:
Known test failures becuase of MCO bug

/lgtm
Feel free to remove the hold after confirming we have the correct kargs now

yanirq:
> Known test failures becuase of MCO bug
> 
> /lgtm
> Feel free to remove the hold after confirming we have the correct kargs now

@slintes  ack,  i might revert the last removal commit if we are not 100% sure this removal is really necessary

PR #121, opened 2020-03-05
Revert CI registry URL

slintes:
Signed-off-by: Marc Sluiter <msluiter@redhat.com>


cynepco3hahue:
@slintes CI deployment passed, but we still have 3 failures, that pretty strange, because of PR with the fix should be already merged.
Can you please create a PR vs `release-4.4`?

Issue #30, opened 2020-01-07
Add status section to the performance profile resource


cynepco3hahue:
- We will need to decide what fields are actual for the status section.

cynepco3hahue:
Duplicate of #22

PR #124, opened 2020-03-05
Add test case id's to tests

gsr-shanks:


functests/performance/cpu_management.go
```diff
@@ -76,7 +76,7 @@ var _ = Describe("[performance] CPU Management", func() {
 			Expect(capacityCPU - allocatableCPU).To(Equal(int64(len(listReservedCPU))))
 		})
 
-		It("Verify CPU affinity mask, CPU reservation and CPU isolation on worker node", func() {
+		It("[rfe_id:27363][test_id:28026][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Verify CPU affinity mask, CPU reservation and CPU isolation on worker node", func() {
```
dshchedr: How about moving "rfe_id" on line #30 ? 
All tests have one rfe, so we can just add it to main Describe and remove duplicates from

functests/performance/cpu_management.go
```diff
@@ -97,7 +97,7 @@ var _ = Describe("[performance] CPU Management", func() {
 
 	})
 
-	Describe("Verification of cpu manager functionality", func() {
+	Describe("[rfe_id:27363][test_id:27492][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Verification of cpu manager functionality", func() {
```
dshchedr: you can remove [rfe_id:27363] from this line as well

functests/performance/performance.go
```diff
@@ -82,11 +82,11 @@ var _ = Describe("performance", func() {
 
 		// Check /usr/local/bin/pre-boot-tuning.sh existence under worker's rootfs
 		const perfRtKernelPrebootTuningScript = "/usr/local/bin/pre-boot-tuning.sh"
-		It(perfRtKernelPrebootTuningScript+" should exist on the nodes", func() {
+		It("[test_id:28523][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] /usr/local/bin/pre-boot-tuning.sh should exist on the nodes", func() {
 			checkFileExistence(workerRTNodes, perfRtKernelPrebootTuningScript)
 		})
 
-		It("Should inject systemd configuration files into initramfs", func() {
+		It("[test_id:28525][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Should inject systemd configuration files into initramfs", func() {
```
yanirq: @gsr-shanks this was removed due to a revert patch we had to enforce. i will add this test again in a different PR with a stricter approach (to also test rpm-ostree latest fix)

gsr-shanks: Ok, please use the same test_id.

slintes: we only removed the reboot flag, so this test is fine, isn't it? It's green at least ;)

slintes:
/approve
/hold

@dshchedr do you want to do another review?
feel free to `/hold cancel` if not or when done :)

functests/performance/performance.go
yanirq: test is fine. 
@dshchedr  once rpm-ostree fix confirmed (should be u/s already)  we can change the test to check setAffinity.conf and system.conf existence in lsinitrd twice (it more accurate)

dshchedr:
lgtm

PR #127, opened 2020-03-06
Add functional tests - updating performance profile

dshchedr:
In this func test we are verifying that all profile parameters can be changed. Unfortunately, after running it - some other functional tests might be failed. 
 I didn't find good solution how to postpone running this test until all other will be finished, so I would be much appreciated for any advises - what is the best way to manage it.

### DIFF
--- a/functests/performance/updating_profile.go
+++ b/functests/performance/updating_profile.go
@@ -1,0 +1,152 @@
+package performance
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+
+	testutils "github.com/openshift-kni/performance-addon-operators/functests/utils"
+	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/nodes"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/profiles"
+	performancev1alpha1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1alpha1"
+)
+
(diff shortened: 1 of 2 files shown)


functests/performance/performance.go
```diff
@@ -170,8 +173,60 @@ var _ = Describe("performance", func() {
...
+			profile, err = profiles.GetByNodeLabels(
+				testclient.Client,
+				map[string]string{
+					fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerRT): "",
+				},
+			)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("Verifies that hugepages can be updated thru performance profile", func() {
```
gsr-shanks: Let's add the test case id's and other parameters while we are at it.

functests/performance/performance.go
```diff
@@ -170,8 +173,60 @@ var _ = Describe("performance", func() {
 			execSysctlOnWorkers(workerRTNodes, sysctlMap)
 		})
 	})
+
+	Describe("Updating parameters in performance profile", func() {
```
gsr-shanks: Let's add the rfe id while we are at it.

functests/performance/performance.go
dshchedr: Will do, but at the end this test should combine several polarion tests.. do you know how it's better to mark all of them?

slintes:
/ok-to-test

@dshchedr will review on monday! :)

functests/performance/performance.go
dshchedr: added [rfe_id:28761]

functests/performance/updating_profile.go
```diff
@@ -0,0 +1,148 @@
...
+					Size:  hpSize,
+				},
+			},
+		}
+		profile.Spec.CPU = &performancev1alpha1.CPU{
+			BalanceIsolated: &f,
+			Reserved:        &reserved,
+			Isolated:        &isolated,
+		}
+		profile.Spec.NUMA.TopologyPolicy = &policy
```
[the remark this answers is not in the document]
dshchedr: done

functests/performance/updating_profile.go
```diff
@@ -0,0 +1,148 @@
...
+		profile.Spec.RealTimeKernel.Enabled = &f
```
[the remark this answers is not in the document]
dshchedr: done

PR #131, opened 2020-03-06
[release-4.4] Add injection check for rpm-ostree

yanirq:

PR #132, opened 2020-03-06
Sync with release-4.4

slintes:
We had some trouble on release-4.4 which we fixed directly on the release branch.
This should not get best practice, however it most the most effective way to move forward.

With this PR we align master with changed from the release branch.
Changes are:

- timeout commit of #126 
- #130: revert of #120, which was #104 on master
- #131: partial re-apply of #120 / #104

### DIFF
--- a/build/assets/scripts/pre-boot-tuning.sh
+++ b/build/assets/scripts/pre-boot-tuning.sh
@@ -5,7 +5,7 @@ set -euo pipefail
 SYSTEM_CONFIG_FILE="/etc/systemd/system.conf"
 SYSTEM_CONFIG_CUSTOM_FILE="/etc/systemd/system.conf.d/setAffinity.conf"
 
-if [ -f /etc/sysconfig/irqbalance ] && [ -f ${SYSTEM_CONFIG_CUSTOM_FILE} ] && grep -ls "IRQBALANCE_BANNED_CPUS=${RESERVED_CPU_MASK_INVERT}" /etc/sysconfig/irqbalance; then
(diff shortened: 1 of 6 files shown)


build/assets/scripts/pre-boot-tuning.sh
```diff
@@ -5,7 +5,7 @@ set -euo pipefail
 SYSTEM_CONFIG_FILE="/etc/systemd/system.conf"
 SYSTEM_CONFIG_CUSTOM_FILE="/etc/systemd/system.conf.d/setAffinity.conf"
 
-if [ -f /etc/sysconfig/irqbalance ] && [ -f ${SYSTEM_CONFIG_CUSTOM_FILE} ] && rpm-ostree status -b | grep -q -e "-I ${SYSTEM_CONFIG_FILE} ${SYSTEM_CONFIG_CUSTOM_FILE}" && egrep -wq "^IRQBALANCE_BANNED_CPUS=${RESERVED_CPU_MASK_INVERT}" /etc/sysconfig/irqbalance; then
+if [ -f /etc/sysconfig/irqbalance ] && [ -f ${SYSTEM_CONFIG_CUSTOM_FILE} ]  && rpm-ostree status -b | grep -q -e "${SYSTEM_CONFIG_FILE} ${SYSTEM_CONFIG_CUSTOM_FILE}" && egrep -wq "^IRQBALANCE_BANNED_CPUS=${RESERVED_CPU_MASK_INVERT}" /etc/sysconfig/irqbalance; then
```
yanirq: not sure if the double space was taken from 4.4 , but i dont think its should be a problem

slintes: It's what we have on the release branch now, see https://github.com/openshift-kni/performance-addon-operators/blob/release-4.4/build/assets/scripts/pre-boot-tuning.sh#L8

We can fix it in a follow up, this really isn't a problem. More important to align branches with this PR :)

slintes:
/hold

let's wait for e2e
I think we should make e2e mandatory and override as an exception. Having to hold after a lgtm in order to wait is not nice.

slintes:
all green :slightly_smiling_face: 

/hold cancel

PR #133, opened 2020-03-08
Add additional kerenel arguments option in CRD

yanirq:
- Added an additional field to the performance profile CRD: `additionalKernelArgs` which is optional and can contain a string list of additional kernel argument that are not included as base arguments.
This gives the field engineer the capability to add more kargs (idle=poll and etc) via the performance_profile.yaml when they deem it necessary.

- Set the following kargs as base 
`intel_iommu=on iommu=pt default_hugepagesz=1GB hugepagesz=1G hugepages=32 intel_pstate=disable skew_tick=1 nohz=on nohz_full=<isolated list> rcu_nocbs=<isolated list> tuned.non_isolcpus=<reserved mask> nosoftlockup`

### DIFF
--- a/deploy/olm-catalog/performance-addon-operator/0.0.1/performance.openshift.io_performanceprofiles_crd.yaml
+++ b/deploy/olm-catalog/performance-addon-operator/0.0.1/performance.openshift.io_performanceprofiles_crd.yaml
@@ -31,6 +31,11 @@ spec:
         spec:
           description: PerformanceProfileSpec defines the desired state of PerformanceProfile.
           properties:
+            additionalKernelArgs:
+              description: Addional kernel arguments.
+              items:
+                type: string
+              type: array
             cpu:
               description: CPU defines set of CPU related parameters.
               properties:
--- a/functests/performance/performance.go
+++ b/functests/performance/performance.go
@@ -116,12 +116,11 @@ var _ = Describe("performance", func() {
 	Context("Additional kernel arguments added from perfomance profile", func() {
 		It("Should set additional kernel arguments on the machine", func() {
 			if profile.Spec.AdditionalKernelArgs != nil {
+				additionalArgs := strings.Join(profile.Spec.AdditionalKernelArgs, " ")
 				for _, node := range workerRTNodes {
 					cmdline, err := nodes.ExecCommandOnMachineConfigDaemon(testclient.Client, &node, []string{"cat", "/proc/cmdline"})
 					Expect(err).ToNot(HaveOccurred())
-					for _, arg := range profile.Spec.AdditionalKernelArgs {
-						Expect(cmdline).To(ContainSubstring(arg))
-					}
+					Expect(cmdline).To(ContainSubstring(additionalArgs))
 				}
 			}
 		})
--- a/functests/performance/performance.go
+++ b/functests/performance/performance.go
@@ -12,6 +12,8 @@ import (
 	testutils "github.com/openshift-kni/performance-addon-operators/functests/utils"
 	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/nodes"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/profiles"
+	performancev1alpha1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1alpha1"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
 	ocv1 "github.com/openshift/api/config/v1"
 	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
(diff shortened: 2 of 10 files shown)


cynepco3hahue:
@yanirq CI fails because of:
```
{
          "lastTransitionTime": "2020-03-08T10:46:32Z",
          "message": "Node ci-op-r75kt-w-b-cx5tg.c.openshift-gce-devel-ci.internal is reporting: \"No kernel-rt package available in the OSContainer with URL registry.svc.ci.openshift.org/ci-op-l9sctqjb/stable@sha256:84e3cbc499264b663ed9dea874d294920bc27ae92a4519f31307bee4a39fd251\"",
          "reason": "1 nodes are reporting degraded status on sync",
          "status": "True",
          "type": "NodeDegraded"
},
```
Can you please check why it happens with coreos team?

yanirq:
> do we have a e2e test which covers the changes? The unit test is nice but a e2e test would be better. Otherwise looks ok to me

@fromanirh , ack, added a functional test.

functests/performance/performance.go
```diff
@@ -103,6 +113,20 @@ var _ = Describe("performance", func() {
...
+				for _, node := range workerRTNodes {
+					cmdline, err := nodes.ExecCommandOnMachineConfigDaemon(testclient.Client, &node, []string{"cat", "/proc/cmdline"})
+					Expect(err).ToNot(HaveOccurred())
+					for _, arg := range profile.Spec.AdditionalKernelArgs {
+						Expect(cmdline).To(ContainSubstring(arg))
+					}
+				}
+			}
+		})
+	})
```
fromanirh: Nice, almost there. I'd add an addional kernel argument, something innocent but easy to find like "quiet", "norhgb" or anything else fit to the task you can find, to prove it is correctly passed down. WDYT?

MarSik: I think you can add anything (eg. cnf.test=1), kernel should ignore unknown values

MarSik: It might also be a good idea to add two: cnf.testB=1 and cnf.testA=2 and test it was not reordered (because we saw that with hugepages).

yanirq: I added the reasonable values under here https://github.com/openshift-kni/performance-addon-operators/pull/133/files#diff-94b0c90ad9b93901aa923f6058f1336c  for testing CI.

deploy/olm-catalog/performance-addon-operator/0.0.1/performance.openshift.io_performanceprofiles_crd.yaml
```diff
@@ -31,6 +31,11 @@ spec:
         spec:
           description: PerformanceProfileSpec defines the desired state of PerformanceProfile.
           properties:
+            additionalKernelArgs:
+              description: Addional kernel arguments.
+              items:
+                type: string
+              type: array
```
fromanirh: the golang json encoder will reorder map keys. But this is an array, so it should be fine.

functests/performance/performance.go
yanirq: @MarSik  @fromanirh   how about now?  checking all the arguments as one subset that should appear in proc/cmdline ?  if the order is not the same it will fail.

MarSik: Isn't it just testing that all are there (ContainSubstring)?

yanirq: Substring is "arg1 arg2 arg3" .   it will fail if it wont be the exact substring, for example if in proc/cmdline we have "arg2 arg1 arg3" then the test fails.

MarSik: Oh, ok, old page. But maybe move the additionalArgs definition out of the loop

pkg/controller/performanceprofile/components/machineconfig/machineconfig_test.go
```diff
@@ -167,6 +177,28 @@ var _ = Describe("Machine Config", func() {
...
+
+		y, err := yaml.Marshal(mc)
+		Expect(err).ToNot(HaveOccurred())
+
+		manifest := string(y)
+
+		labelKey, labelValue := components.GetFirstKeyAndValue(profile.Spec.MachineConfigLabel)
+		Expect(manifest).To(ContainSubstring(fmt.Sprintf("%s: %s", labelKey, labelValue)))
+		Expect(manifest).To(ContainSubstring(expectedSystemdUnits))
+		Expect(manifest).To(ContainSubstring(expectedBootArgumentsWithoutAdditionalKerenlArgs))
```
slintes: I think we can add another Expect here for the AdditionalKernelArgs?

yanirq: ah, i have a mistake in the name. should be `expectedBootArgumentsWithAdditionalKerenlArgs`  (not without)

yanirq:
/hold
until CI passes not (without rt kernel test only)

fromanirh:
we need a CI lane which detects these issue. I did a review and didn't spot the whitespace thinghy

yanirq:
> we need a CI lane which detects these issue. I did a review and didn't spot the whitespace thinghy

yes, although whitespace is not the issue in cherry picking here

PR #134, opened 2020-03-09
Temporarily disable RT kernel

slintes:
Reason: https://bugzilla.redhat.com/show_bug.cgi?id=1811630

### DIFF
--- a/cluster-setup/ci-cluster/performance/performance_profile.patch.yaml
+++ b/cluster-setup/ci-cluster/performance/performance_profile.patch.yaml
@@ -11,5 +11,7 @@ spec:
     - size: "1G"
       count: 1
       node: 0
+  realTimeKernel:
+    enabled: false
   nodeSelector:
     node-role.kubernetes.io/worker-rt: ""


simon3z:
@slintes @MarSik what are the other options (if any) that we dismissed to address this temporary issue?

MarSik:
> @slintes @MarSik what are the other options (if any) that we dismissed to address this temporary issue?

This is master branch only. 4.4 seems to be fine. There seems to be no workaround since the packages are simply missing afaik.

slintes:
> > @slintes @MarSik what are the other options (if any) that we dismissed to address this temporary issue?
> 
> This is master branch only. 4.4 seems to be fine. There seems to be no workaround since the packages are simply missing afaik.

Correct. No way to install rt kernel on OCP 4.5 atm, and OCP 4.4 is fine.
Because of the failed installation of the rt kernel the Node is getting degraded, MCP update never finishes, and CI fails before our tests run. With this patch we will still see a test failure, but at least tests will run.

yanirq:
/lgtm
/hold
@slintes  feel free to remove hold once you see fit.   only the RT test is failing as expected

slintes:
Only expected `[It] should have RT kernel enabled ` test failure

slintes:
> /lgtm
> /hold
> @slintes feel free to remove hold once you see fit. only the RT test is failing as expected

ha, didn't see this, thx

/hold cancel

PR #137, opened 2020-03-09
Be less verbose, decrease wait times

slintes:
- Suppress some output during deployment (except during last iteration) and waiting for MCP
- Give operator less time for doimg its work, 30s should be enough
- Revert MCP wait time to old value (there always was a real issue when we run into the timeout)

### DIFF
--- a/hack/deploy.sh
+++ b/hack/deploy.sh
@@ -25,18 +25,31 @@ do
 
   echo "[INFO] Deploying performance operator and profile."
   set +e
-  if ! ${OC_TOOL} kustomize $feature_dir | envsubst | ${OC_TOOL} apply -f -
-  then
-    set -e
+
+  # be verbose on last iteration only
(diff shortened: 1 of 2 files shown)


slintes:
All green :)

hack/deploy.sh
```diff
@@ -25,19 +25,32 @@ do
...
+
+  # be verbose on last iteration only
+  if [[ $iterations -eq $((max_iterations - 1)) ]] || [[ -n "${VERBOSE}" ]]; then
+    ${OC_TOOL} kustomize $feature_dir | envsubst | ${OC_TOOL} apply -f -
+  else
+    ${OC_TOOL} kustomize $feature_dir | envsubst | ${OC_TOOL} apply -f - &> /dev/null
+  fi
+
+  # shellcheck disable=SC2181
+  if [[ $? != 0 ]];then
```
cynepco3hahue: nit: space after `;`

PR #138, opened 2020-03-12
fix guaranteed and cpu affinity tests

yuvalk:
in guaranteed mode, pod can land on any NOT reserved cpu (not just isolated one)

Signed-off-by: Yuval Kashtan <yuvalkashtan@gmail.com>


yuvalk:
Realized my previous attempt was bogus
hopefully this is a more correct one

PR #139, opened 2020-03-13
tools: add a simpler tool to update the image ref

fromanirh:
This patch adds a tool to just and only update the image references
in a CSVs.
We want to use this tool to support the upgrade flows when
we will have more than one stable release.
This way, we will be able to update the older CSVs automatically
from a newer release with minimal changes.

Say, the manifest image of version N could update the bundled
manifests of version N-1 with minimal changes, thus simplifying
the upgrade graphs.

The intended overall flow is the following:
When we release version Y, we will include manifest bundles for older
versions (version Y-1 to say the least).
If a bugfix comes in stream Y-1, with this tool will have the *option*
to pull the newer version in the current Y version and release in the
manifest bundles.

Without this patch, we have only two options:
1. NEVER update the Y-1 CSVs once version Y is out. This way, should
   bugfix versions Y-1 be released, we will have no option of pulling
   them from version Y releases. This is a fine choice per se
    and up to us to decide if we want to go this way
2. When composing the Y version metadata images, pull the most recent
   CSVs from most recent metadata image from version Y-1.

This PR enables us to simplify option #2 while keeping option #1 totally
open.
The rationale is that the most frequent change to CSV is update of the
image reference, a change we can anticipate and it is safe to reproduce.
Everything else in the CSV changes very rarely in a stable stream.

### DIFF
--- a/tools/csv-replace-imageref/csv-replace-imageref.go
+++ b/tools/csv-replace-imageref/csv-replace-imageref.go
@@ -1,0 +1,164 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"strings"
+
+	yaml "github.com/ghodss/yaml"
+	csvv1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	appsv1 "k8s.io/api/apps/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type csvClusterPermissions struct {
+	ServiceAccountName string              `json:"serviceAccountName"`
+	Rules              []rbacv1.PolicyRule `json:"rules"`
+}
+
+type csvPermissions struct {
+	ServiceAccountName string              `json:"serviceAccountName"`
+	Rules              []rbacv1.PolicyRule `json:"rules"`
+}
+
+type csvDeployments struct {
+	Name string                `json:"name"`
+	Spec appsv1.DeploymentSpec `json:"spec,omitempty"`
+}
+
+type csvStrategySpec struct {
+	ClusterPermissions []csvClusterPermissions `json:"clusterPermissions"`
+	Permissions        []csvPermissions        `json:"permissions"`
+	Deployments        []csvDeployments        `json:"deployments"`
+}
+
+var (
+	csvInput      = flag.String("csv-input", "", "path to csv to update")
+	operatorImage = flag.String("operator-image", "", "operator container image")
+)
+
+func unmarshalCSV(filePath string) *csvv1.ClusterServiceVersion {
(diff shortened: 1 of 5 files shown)


fromanirh:
/hold
to give time to reviewrs

tools/csv-replace-imageref/csv-replace-imageref.go
```diff
@@ -0,0 +1,164 @@
...
+	ServiceAccountName string              `json:"serviceAccountName"`
+	Rules              []rbacv1.PolicyRule `json:"rules"`
+}
+
+type csvDeployments struct {
+	Name string                `json:"name"`
+	Spec appsv1.DeploymentSpec `json:"spec,omitempty"`
+}
+
+type csvStrategySpec struct {
```
slintes: Can you use types from here?
https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go#L67

fromanirh: Sure. And this snippet should probably be consolidated somewhere, because I borrowed from the csv-generator tool.

fromanirh: Uhm, turns out we can't really upgrade the dep (and thus use the proper types) until we rebase on 1.18. That's probably the reason why David wrote this way in the first place.
I'll consolidate the common code, so we at least need to update just one place.

fromanirh:
Uhm, I don't think this failure is this patch fault:
```
info: Using registry public hostname registry.svc.ci.openshift.org
Saved credentials for registry.svc.ci.openshift.org
info: Found 247 images in image stream
error: unable to read image registry.svc.ci.openshift.org/ci-op-lwhzfqnz/stable@sha256:595142acf537b1b17b8e340c2de6b73364c79034401d58f5b09dfbd6b742e5c8: received unexpected HTTP status: 500 Internal Server Error
2020/03/13 13:13:21 Container release in pod release-latest failed, exit code 1, reason Error
2020/03/13 13:13:24 No custom metadata found and prow metadata already exists. Not updating the metadata.
2020/03/13 13:13:25 Ran for 27s
2020/03/13 13:13:25 Submitted failure event to sentry (id=8c6dd661bd644c02bf371ed7d55393b3)
error: could not run steps: step [release:latest] failed: release "release-latest" failed: the pod ci-op-lwhzfqnz/release-latest failed after 11s (failed containers: release): ContainerFailed one or more containers exited

Container release exited with code 1, reason Error
---
```
@slintes any idea why this is failing?

slintes:
CI seem(s|ed) to have some troubles.... :man_shrugging: 

/retest

tools/csv-replace-imageref/csv-replace-imageref.go
slintes: oh ok, too bad

slintes:
/lgtm

just try to retest later again, we are not alone with this issue

fromanirh:
No worries, if we like this approach this PR can wait for monday.

PR #141, opened 2020-03-17
API: remove default hugepages size

cynepco3hahue:
To avoid k8s issue https://github.com/kubernetes/kubernetes/issues/80452
let not expose the default hugepages size, under the API, once the issue
fixed, we can re-add it and also we have `additionalKernelArguments` in case
if someone really need to add default hugepages size.


cynepco3hahue:
DPDK requires default hugepages size kernel boot parameter.

PR #142, opened 2020-03-17
make topology manager func test more flexible

dshchedr:
This PR slightly improves the logic of topology manager functional test - now it should work with any pre-configured topology manager policies

### DIFF
--- a/functests/performance/topology_manager.go
+++ b/functests/performance/topology_manager.go
@@ -16,6 +16,8 @@ import (
 	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/nodes"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/pods"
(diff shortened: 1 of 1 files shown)


functests/performance/topology_manager.go
```diff
@@ -43,8 +52,8 @@ var _ = Describe("[rfe_id:27350][performance]Topology Manager", func() {
 		Expect(ok).To(BeTrue())
 		Expect(enabled).To(BeTrue())
 
-		// verify topology manager poicy
-		Expect(kubeletConfig.TopologyManagerPolicy).To(Equal(kubeletconfigv1beta1.SingleNumaNodeTopologyManager))
+		// verify topology manager policy
+		Expect(kubeletConfig.TopologyManagerPolicy).To(Equal(*profile.Spec.NUMA.TopologyPolicy))
```
slintes: both NUMA and TopologyPolicy are optional, so better check if both are set, and if not fall back to default (best-effort?), or maybe skip test at all.

dshchedr: good point! fixed

PR #144, opened 2020-03-19
Merge network and realtime profile to one profile

yanirq:
Combine network latency and realtime performance profiles into one which will contain all default settings. this will result in one tuned profile in the system instead of 2.


cynepco3hahue:
@yanirq Looks good to me, but I prefer to have an additional reviewer, @slintes @fromanirh can you please take a look?

yanirq:
> @yanirq The original realtime tuning had `include=openshift-node-network-latency` in the `[main]` section. Don't we need it in the merged tuning?

@slintes  `openshift-node-network-latency` created by us and was included in a canonical way. since we merged the profiles we don't need one to inherit from the other.

slintes:
> > @yanirq The original realtime tuning had `include=openshift-node-network-latency` in the `[main]` section. Don't we need it in the merged tuning?
> 
> @slintes `openshift-node-network-latency` created by us and was included in a canonical way. since we merged the profiles we don't need one to inherit from the other.

oh yes sure, I missed the rename of the merged script :laughing: 

/approve
/hold cancel

PR #145, opened 2020-03-19
add functional test: changing nodeSelector in performance profile

dshchedr:
Running this test leads to reconfiguring performance profile parameters, as result some other functional tests might be failed. That's why we decided to run such kind of tests when other will be completed:
https://github.com/openshift-kni/performance-addon-operators/blob/cf3e15b602ce3f015c2e432bb1c8ebf2eff00606/hack/run-functests.sh#L10-L11

and it works fine for CI, because cluster is killed at the end and we don't care about latest profile values on it. 
But it is not good for real customer who want to keep his cluster after running this test (he needs to revert back all changes to get initial tuning parameters)

### DIFF
--- a/functests/performance/updating_profile.go
+++ b/functests/performance/updating_profile.go
@@ -9,6 +9,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
@@ -18,6 +19,10 @@ import (
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/nodes"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/profiles"
 	performancev1alpha1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1alpha1"
+)
+
+const (
+	mcpUpdateTimeout = 20
 )
 
 var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", func() {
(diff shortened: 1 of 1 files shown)


functests/performance/updating_profile.go
```diff
@@ -113,12 +118,55 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+		var newWorkerNode corev1.Node
+		newRole := "worker-test"
+		newLabel := fmt.Sprintf("%s/%s", testutils.LabelRole, newRole)
+		newNodeSelector := map[string]string{newLabel: ""}
+
+		By("Skipping test if cluster does not have another available worker node")
+		workerNodes, err := nodes.GetByRole(testclient.Client, "worker")
+		Expect(err).ToNot(HaveOccurred())
+
+		for i := range workerNodes {
```
slintes: hm, since you only read from the slice, it's safe to use the value from the range, and not the index (`_, node := ...`). It's a bit nicer.

functests/performance/updating_profile.go
```diff
@@ -113,12 +118,55 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+			if _, lbl := workerNodes[i].Labels[fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerRT)]; lbl {
```
slintes: AFAIk it's best practice to just use `ok` for the boolean.
Something `exists` would work as well for me though.
But `lbl` does not indicate it's a boolean, so please change this.

functests/performance/updating_profile.go
```diff
@@ -113,12 +118,55 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
 			}
 		}
 	})
+
+	It("[test_id:28440]Verifies that nodeSelector can be updated in performance profile", func() {
+		var newWorkerNode corev1.Node
```
slintes: What about using a pointer?
Makes the check if you found a node a bit nicer later on (skip if node is nil, and not if node.name = "")

functests/performance/updating_profile.go
dshchedr: right, thanks!

functests/performance/updating_profile.go
dshchedr: yeah, you are right! thanks!

functests/performance/updating_profile.go
dshchedr: done, can you please take a look on changes

PR #149, opened 2020-03-24
Fix performance test for merged profile

yanirq:
Test runs that have a profile name combined with a prefix are not able to find the new merged tuned profile we create for performance tunings.


yanirq:
/hold
testing also locally to catch possible mismatches

PR #150, opened 2020-03-24
Test RT Kernel not installed on non worker-rt node

gsr-shanks:


yanirq:
@gsr-shanks  can you add a short description in the PR please?  what the test exactly tries to do and accomplish ?

gsr-shanks:
> @gsr-shanks can you add a short description in the PR please? what the test exactly tries to do and accomplish ?

@yanirq I have updated my commit message with more information.

functests/performance/rt-kernel.go
```diff
@@ -59,4 +60,54 @@ var _ = Describe("[performance]RT Kernel", func() {
 
 	})
 
+	It("[test_id:28526][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Non worker-cnf node should not have RT kernel installed", func() {
+		var newWorkerNode *corev1.Node
+		newRole := "worker-test"
+		newLabel := fmt.Sprintf("%s/%s", testutils.LabelRole, newRole)
+		// newNodeSelector := map[string]string{newLabel: ""}
```
yanirq: remove comment

functests/performance/rt-kernel.go
```diff
@@ -59,4 +60,53 @@ var _ = Describe("[performance]RT Kernel", func() {
...
+			}
+
+			logs, err := pods.GetLogs(testclient.K8sClient, testpod)
+			if err != nil {
+				return ""
+			}
+
+			return logs
+
+		}, 15*time.Minute, 30*time.Second).ShouldNot(ContainSubstring("PREEMPT RT"))
```
slintes: Would it make sense to check if the command did run successfully, e.g by looking for "Linux" in the string?

functests/performance/rt-kernel.go
```diff
@@ -59,4 +60,53 @@ var _ = Describe("[performance]RT Kernel", func() {
 
 	})
 
+	It("[test_id:28526][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Non worker-cnf node should not have RT kernel installed", func() {
+		var newWorkerNode *corev1.Node
+		newRole := "worker-test"
+		newLabel := fmt.Sprintf("%s/%s", testutils.LabelRole, newRole)
+
+		By("Skipping test if cluster does not have another available worker node")
+		workerNodes, err := nodes.GetByRole(testclient.Client, "worker")
```
yanirq: maybe try utilizing this syntax here:
` oc get node --selector='!node-role.kubernetes.io/worker-rt , node-role.kubernetes.io/worker'`

If `nodes.GetByRole` is not enough for this purpose you can add another util function under nodes

dshchedr: we have the same code in updating_profile.go:
https://github.com/openshift-kni/performance-addon-operators/blob/a3ff53ae8ce51c5ed07b5143fdded78545512b9f/functests/performance/updating_profile.go#L128

I agree - probably it's better to create function in utils/nodes that will return the list of non-rt workers

functests/performance/rt-kernel.go
```diff
@@ -59,4 +60,53 @@ var _ = Describe("[performance]RT Kernel", func() {
...
+			newWorkerNode = &node
+			break
+		}
+		if newWorkerNode == nil {
+			Skip("Skipping test because there are no additional worker nodes")
+		}
+		newWorkerNode.Labels[newLabel] = ""
+		Expect(testclient.Client.Update(context.TODO(), newWorkerNode)).ToNot(HaveOccurred())
+
+		Eventually(func() string {
```
dshchedr: In my opinion we don't  need to run pod in this test. I hope it's enough to verify kernel directly on the node, in this case this "eventually" block can be changed to:

			cmd = []string{"uname", "-a"}
			Expect(execCommandOnWorker(cmd, &node)).NotTo(ContainSubstring("PREEMPT RT"), "Node should have non-RT kernel")

functests/performance/rt-kernel.go
gsr-shanks: Ack.

functests/performance/rt-kernel.go
gsr-shanks: In that case, we will need to change the previous test case as well.

functests/performance/rt-kernel.go
```diff
@@ -59,4 +60,20 @@ var _ = Describe("[performance]RT Kernel", func() {
...
+		nonRTWorkerNode, err := nodes.GetNonCNFNode(testclient.Client, "worker")
+		Expect(err).ToNot(HaveOccurred())
+
+		if nonRTWorkerNode == nil {
+			Skip("Skipping test because there are no additional non-cnf worker nodes")
+		}
+
+		fmt.Println(nonRTWorkerNode)
+		cmd := []string{"uname", "-a"}
+		Expect(execCommandOnWorker(cmd, nonRTWorkerNode)).To(ContainSubstring("Linux"), "Node should have Linux string")
```
gsr-shanks: I see 
`functests/performance/rt-kernel.go:74:29: cannot use nonRTWorkerNode (type []"k8s.io/api/core/v1".Node) as type *"k8s.io/api/core/v1".Node in argument to execCommandOnWorker`

It looks like in nodes.go I am returning the wrong type.

functests/performance/rt-kernel.go
```diff
@@ -59,4 +60,20 @@ var _ = Describe("[performance]RT Kernel", func() {
...
+
+		By("Skipping test if cluster does not have another available worker node")
+		nonRTWorkerNode, err := nodes.GetNonCNFNode(testclient.Client, "worker")
+		Expect(err).ToNot(HaveOccurred())
+
+		if nonRTWorkerNode == nil {
+			Skip("Skipping test because there are no additional non-cnf worker nodes")
+		}
+
+		fmt.Println(nonRTWorkerNode)
```
gsr-shanks: And I need to remove this print statement.

functests/performance/rt-kernel.go
dshchedr: hey, as i see here you are returning list:
https://github.com/openshift-kni/performance-addon-operators/blob/b017c804fe9df5cd3d42ef2eb56dd45549a2add0/functests/utils/nodes/nodes.go#L37

and this one: as type _*"k8s.io/api/core/v1".Node_  means you need to add pointer.. so I think you need: &nonRTWorkerNode[0]

functests/performance/rt-kernel.go
```diff
@@ -59,4 +60,19 @@ var _ = Describe("[performance]RT Kernel", func() {
 
 	})
 
+	It("[test_id:28526][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Non worker-cnf node should not have RT kernel installed", func() {
+
+		By("Skipping test if cluster does not have another available worker node")
+		nonRTWorkerNode, err := nodes.GetNonCNFNode(testclient.Client, "worker")
+		Expect(err).ToNot(HaveOccurred())
+
+		if nonRTWorkerNode == nil {
```
dshchedr: IIUC in this line you are verifying if the list of non-cnf nodes is empty? I think you need to use this:
if len(nonRTWorkerNode) == 0

functests/utils/nodes/nodes.go
```diff
@@ -32,6 +33,21 @@ func GetByRole(c client.Client, role string) ([]corev1.Node, error) {
 	return nodes.Items, nil
 }
 
+// GetNonCNFNode retuns node with no worker-rt label
```
dshchedr: *returns list of nodes

slintes: "returns"

functests/utils/nodes/nodes.go
```diff
@@ -32,6 +33,21 @@ func GetByRole(c client.Client, role string) ([]corev1.Node, error) {
...
+func GetNonCNFNode(c client.Client, role string) ([]corev1.Node, error) {
+	nonRTWorkerNode := []corev1.Node{}
+
+	workerNodes, err := GetByRole(testclient.Client, role)
+	for _, node := range workerNodes {
+		if _, ok := node.Labels[fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerRT)]; ok {
+			continue
+		}
+		nonRTWorkerNode = append(nonRTWorkerNode, node)
+		break
```
dshchedr: i think it's better to remove **break**
Without it we will have full list of noncnf nodes, it might be more useful then returning only one node

functests/utils/nodes/nodes.go
```diff
@@ -32,6 +33,21 @@ func GetByRole(c client.Client, role string) ([]corev1.Node, error) {
 	return nodes.Items, nil
 }
 
+// GetNonCNFNode retuns node with no worker-rt label
+func GetNonCNFNode(c client.Client, role string) ([]corev1.Node, error) {
```
slintes: - `c` is unused and can be removed (actually it can be removed in all func here and replaced with direct use of `testclient.Client` like you do here as well)
- IMHO we can even remove `role` and use a hardcoded const for "worker" like we do for `testutils.RoleWorkerRT`
- we use workerRT (at least as const name), and return multiple worker nodes, so maybe rename the func `GetNonRTWorkers()`?

functests/performance/rt-kernel.go
slintes: this check won't work (it will never be false): in `GetNonCNFNode` you init the nodes with `nodesWithResource := []corev1.Node{}`, which is an _empty_, but not _nil_ slice. So you need to check `len(nonRTWorkerNode) == 0` here.

gsr-shanks: Done.

functests/utils/nodes/nodes.go
gsr-shanks: Done.

functests/utils/nodes/nodes.go
gsr-shanks: Ack. Done.

functests/performance/rt-kernel.go
gsr-shanks: Yes, I have resolved this to check the len of slice.

functests/performance/rt-kernel.go
```diff
@@ -59,4 +60,19 @@ var _ = Describe("[performance]RT Kernel", func() {
 
 	})
 
+	It("[test_id:28526][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Non worker-cnf node should not have RT kernel installed", func() {
+
+		By("Skipping test if cluster does not have another available worker node")
+		nonRTWorkerNode, err := nodes.GetNonRTWorkers()
```
slintes: nit: please use plural for the var as well, avoids confusion (sorry, missed it in earlier review)

functests/performance/rt-kernel.go
```diff
@@ -59,4 +60,19 @@ var _ = Describe("[performance]RT Kernel", func() {
...
+		Expect(err).ToNot(HaveOccurred())
+
+		if len(nonRTWorkerNode) == 0 {
+			Skip("Skipping test because there are no additional non-cnf worker nodes")
+		}
+
+		cmd := []string{"uname", "-a"}
+		Expect(execCommandOnWorker(cmd, &nonRTWorkerNode[0])).To(ContainSubstring("Linux"), "Node should have Linux string")
```
slintes: you can run `execCommandOnWorker` once only and use the result in both `Expect`s

functests/utils/nodes/nodes.go
```diff
@@ -32,6 +33,20 @@ func GetByRole(c client.Client, role string) ([]corev1.Node, error) {
 	return nodes.Items, nil
 }
 
+// GetNonRTWorkers returns list of nodes with no worker-rt label
+func GetNonRTWorkers() ([]corev1.Node, error) {
+	nonRTWorkerNode := []corev1.Node{}
+
+	workerNodes, err := GetByRole(testclient.Client, "worker")
```
slintes: please use a new const for it, like `testutils.RoleWorkerRT` (just without the need that it can be overridden by an env var, I think `worker` is always `worker` :) )

functests/utils/nodes/nodes.go
```diff
@@ -32,6 +33,20 @@ func GetByRole(c client.Client, role string) ([]corev1.Node, error) {
 	return nodes.Items, nil
 }
 
+// GetNonRTWorkers returns list of nodes with no worker-rt label
+func GetNonRTWorkers() ([]corev1.Node, error) {
+	nonRTWorkerNode := []corev1.Node{}
```
slintes: plural please

functests/performance/rt-kernel.go
```diff
@@ -59,4 +60,20 @@ var _ = Describe("[performance]RT Kernel", func() {
...
+		By("Skipping test if cluster does not have another available worker node")
+		nonRTWorkerNodes, err := nodes.GetNonRTWorkers()
+		Expect(err).ToNot(HaveOccurred())
+
+		if len(nonRTWorkerNodes) == 0 {
+			Skip("Skipping test because there are no additional non-cnf worker nodes")
+		}
+
+		cmd := []string{"uname", "-a"}
+		execOut := Expect(execCommandOnWorker(cmd, &nonRTWorkerNodes[0]))
```
slintes: I expected more like this, I think it's better readable:
```
kernel := execCommandOnWorker(cmd, &nonRTWorkerNodes[0])
Expect(kernel).To...
Expect(kernel).NotTo...
```
But yours will work as well, I leave it up to you if want to change it.

slintes:
/hold

in case you want to address the comment

slintes:
/hold cancel

:+1: :slightly_smiling_face:

PR #154, opened 2020-03-27
Add Informational status reporting on Performance Profile CR against Machine config pools

yanirq:
Add additional status fields to the Performance Profile CR that would provide additional information that would be useful for an admin to debug in case of issue:
When status of machine config pools that are attached to the performance profile are in a degraded state the performance profile status will be degraded and will contain the machine config pool failure message.

### DIFF
--- a/pkg/controller/performanceprofile/performanceprofile_controller.go
+++ b/pkg/controller/performanceprofile/performanceprofile_controller.go
@@ -304,7 +304,6 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
 
 func (r *ReconcilePerformanceProfile) ppRequestsFromMCP(o handler.MapObject) []reconcile.Request {
 
-	klog.Infof("Getting performance profile requests from machine config pool %q", namespacedName(o.Meta).String())
 	mcp := &mcov1.MachineConfigPool{}
 
 	if err := r.client.Get(context.Background(),
@@ -328,8 +327,10 @@ func (r *ReconcilePerformanceProfile) ppRequestsFromMCP(o handler.MapObject) []r
 	for k := range ppList.Items {
 		if hasMatchingLabels(&ppList.Items[k], mcp) {
 			requests = append(requests, reconcile.Request{NamespacedName: namespacedName(&ppList.Items[k])})
+			klog.Infof("Adding performance profile %q request from machine config pool %q", ppList.Items[k].GetName(), namespacedName(o.Meta).String())
 		}
 	}
+
 	return requests
 }
 
--- a/pkg/controller/performanceprofile/performanceprofile_controller.go
+++ b/pkg/controller/performanceprofile/performanceprofile_controller.go
@@ -275,7 +275,10 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
-	conditions := r.getAvailableConditions()
+	conditions := r.getConditionsByMCPStatus(instance)
+	if conditions == nil {
+		conditions = r.getAvailableConditions()
+	}
 	if err := r.updateStatus(instance, conditions); err != nil {
 		klog.Errorf("failed to update performance profile %q status: %v", instance.Name, err)
 		// we still want to requeue after some, also in case of error, to avoid chance of multiple reboots
@@ -288,15 +291,6 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
 
 	if result != nil {
 		return *result, nil
-	}
-
-	conditions = r.getConditionsByMCPStatus(instance)
-	if conditions != nil {
-		err := r.updateStatus(instance, conditions)
-		if err != nil {
-			klog.Errorf("failed to update performance profile %q status: %v", instance.Name, err)
-			return reconcile.Result{}, err
-		}
 	}
 
 	return reconcile.Result{}, nil
--- a/pkg/controller/performanceprofile/performanceprofile_controller.go
+++ b/pkg/controller/performanceprofile/performanceprofile_controller.go
@@ -135,7 +135,6 @@ func add(mgr manager.Manager, r reconcile.Reconciler, mapMCPToPC handler.ToReque
 		return err
 	}
 
-	// we do not want initiate reconcile loop on the configuration or pause fields update
 	mcpPredicates := predicate.Funcs{
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			if e.MetaOld == nil {
@@ -162,9 +161,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler, mapMCPToPC handler.ToReque
 			mcpNewCopy.Spec.Paused = mcpOld.Spec.Paused
 			mcpNewCopy.Spec.Configuration = mcpOld.Spec.Configuration
 
-			return !reflect.DeepEqual(mcpOld.Spec, mcpNewCopy.Spec) ||
-				!reflect.DeepEqual(mcpOld.Status, mcpNewCopy.Status) ||
-				!apiequality.Semantic.DeepEqual(e.MetaNew.GetLabels(), e.MetaOld.GetLabels())
+			return !reflect.DeepEqual(mcpOld.Status.Conditions, mcpNewCopy.Status.Conditions)
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


slintes:
e2e is fixed by #162
/retest

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -294,6 +296,38 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcilePerformanceProfile) ppRequestsFromMCP(o handler.MapObject) []reconcile.Request {
+
+	mcp := &mcov1.MachineConfigPool{}
+
+	if err := r.client.Get(context.Background(),
+		client.ObjectKey{
```
cynepco3hahue, line 304: can you use `namespacedName` here?

pkg/controller/performanceprofile/status.go
```diff
@@ -144,3 +147,39 @@ func (r *ReconcilePerformanceProfile) getProgressingConditions(reason string, me
 		},
 	}
 }
+
+func (r *ReconcilePerformanceProfile) getConditionsByMCPStatus(profile *performancev1alpha1.PerformanceProfile) []conditionsv1.Condition {
```
cynepco3hahue, line 151: I think it more correct to name it, `getMCPConditionsByProfile` or just `getMCPConditions`

pkg/controller/performanceprofile/status.go
```diff
@@ -144,3 +147,39 @@ func (r *ReconcilePerformanceProfile) getProgressingConditions(reason string, me
...
+
+	mcpList := &mcov1.MachineConfigPoolList{}
+	err := r.client.List(context.TODO(), mcpList)
```
cynepco3hahue, line 154: you can use one liner for every such call
```
if err := r.client.List(context.TODO(), mcpList); err != nil {
    return err
}

File: pkg/controller/performanceprofile/performanceprofile_controller.go [hunk 4]
```diff
@@ -294,6 +296,38 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcilePerformanceProfile) ppRequestsFromMCP(o handler.MapObject) []reconcile.Request {
+
+	mcp := &mcov1.MachineConfigPool{}
+
+	if err := r.client.Get(context.Background(),
+		types.NamespacedName{
```
cynepco3hahue, line 304: `namespacedName(o.Meta)` should work

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -276,7 +272,20 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
-	conditions := r.getAvailableConditions()
+	conditions, err := r.getMCPConditionsByProfile(instance)
+	if err != nil {
+		conditions := r.getDegradedConditions(conditionReasonComponentsCreationFailed, err.Error())
```
slintes, line 277: please use a new reason, components are created successfully at this point.

pkg/controller/performanceprofile/performanceprofile_controller.go
```diff
@@ -294,6 +303,37 @@ func (r *ReconcilePerformanceProfile) Reconcile(request reconcile.Request) (reco
...
+		},
+		mcp,
+	); err != nil {
+		klog.Errorf("No-op: Unable to retrieve mcp %q from store: %v", namespacedName(o.Meta).String(), err)
+		return nil
+	}
+
+	ppList := &performancev1alpha1.PerformanceProfileList{}
+	if err := r.client.List(context.TODO(), ppList); err != nil {
+		klog.Errorf("No-op: Unable to list performance profiles: %v", err)
```
slintes, line 323: why the "no-op" error message prefix?

pkg/controller/performanceprofile/status.go
```diff
@@ -144,3 +148,67 @@ func (r *ReconcilePerformanceProfile) getProgressingConditions(reason string, me
...
+		klog.Errorf("Cannot list Machine config pools to match with profile %q : %v", profile.Name, err)
+		return nil, err
+	}
+
+	mcpItems := removeMCPDuplicateEntries(mcpList.Items)
+
+	performanceProfileLabels := labels.Set(profileutil.GetMachineConfigPoolSelector(profile))
+
+	reason := bytes.Buffer{}
+	reason.WriteString("Matching Machine Config Pools are in a degraded state")
```
slintes, line 166: `reason` should be a short, predefined, machine readable string IMHO.
I see that the existing reasons (see consts on top of this file) also do not fully comply with this...
I'd suggest to add a new reason and modify existing one like this:
```
	conditionReasonValidationFailed         = "ValidationFailed"
	conditionReasonComponentsCreationFailed = "ComponentCreationFailed"
	conditionReasonMCPDegraded = "MCPDegraded"
 ```
In the message you can give the detailed dynamic info.

File: pkg/controller/performanceprofile/performanceprofile_controller.go [hunk 8]
```diff
@@ -44,11 +46,12 @@ const finalizer = "foreground-deletion"
 // Add creates a new PerformanceProfile Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
-	return add(mgr, newReconciler(mgr))
+	r := newReconciler(mgr)
+	return add(mgr, r, r.ppRequestsFromMCP)
```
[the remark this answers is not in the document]
yanirq: @slintes  The problem is that` r reconcile.Reconciler` which add gets:
https://github.com/openshift-kni/performance-addon-operators/blob/20b946decfb3e4d8f14b28e326efe38bd3b14c7c/pkg/controller/performanceprofile/performanceprofile_controller.go#L64
Is not a `ReconcilePerformanceProfile` and I'm not eager to change the type here because I assume it should stay an interface here. 
WDYF ?

pkg/controller/performanceprofile/performanceprofile_controller.go
yanirq: We can still get an MCP degraded state after components were successfully created.
I tried keeping this as idempotent as I can.
Basically we will get new requests (new reconcile loops) on every mcp status change (mcp that relates to an existing Performance Profile) 
So if an existing on is re-passed here due to a MCP status change: the components shouldn't re-apply  and the available conditions will remain in case no degraded state is detected

pkg/controller/performanceprofile/performanceprofile_controller.go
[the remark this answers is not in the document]
slintes: Hm, I see. But changing type isn't a problem IMHO, `add()` isn't public and only called by `Add()`

pkg/controller/performanceprofile/performanceprofile_controller.go
slintes: it's just about the reason in the condition (`conditionReasonComponentsCreationFailed`). It should be something like `FailedGettingMCPStatus`). Or do I misunderstand something?

yanirq: right. missed that extra redundant line. it shouldn't have been there.
good catch.

pkg/controller/performanceprofile/status.go
```diff
@@ -1,19 +1,25 @@
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/klog"
 )
 
 const (
 	conditionReasonValidationFailed         = "validation failed"
 	conditionReasonComponentsCreationFailed = "failed to create components"
+	conditionReasonMCPDegraded              = "MCP degraded"
+	conditionFailedGettingMCPStatus         = "failed getting MCP status"
```
slintes, line 22: please also change the values to e.g. 
```
        conditionReasonValidationFailed         = "ValidationFailed"
	conditionReasonComponentsCreationFailed = "ComponentCreationFailed"
	conditionReasonMCPDegraded              = "MCPDegraded"
        conditionReasonFailedGettingMCPStatus   = "GettingMCPStatusFailed"
```
```

PR #158, opened 2020-03-30
Divide updating profile tests for better interraction with polarion

dshchedr:
For better interaction with Polarion it would be good to divide one big test on small parts

### DIFF
--- a/functests/performance/updating_profile.go
+++ b/functests/performance/updating_profile.go
@@ -3,10 +3,10 @@ package performance
 import (
 	"context"
 	"fmt"
-	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -30,6 +30,10 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
 	var profile *performancev1alpha1.PerformanceProfile
 	var err error
 
+	chkKernel := []string{"uname", "-a"}
+	chkCmdLine := []string{"cat", "/proc/cmdline"}
+	chkKubeletConfig := []string{"cat", "/rootfs/etc/kubernetes/kubelet.conf"}
+
 	BeforeEach(func() {
 		workerRTNodes, err = nodes.GetByRole(testclient.Client, testutils.RoleWorkerRT)
 		Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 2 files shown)


slintes:
e2e is fixed by #162 
/retest

gsr-shanks:
lgtm

functests/performance/updating_profile.go
```diff
@@ -43,80 +47,91 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+				for _, param := range parameter {
+					if shouldContain {
+						Expect(execCommandOnWorker(cmd, &node)).To(ContainSubstring(param))
+					} else {
+						Expect(execCommandOnWorker(cmd, &node)).NotTo(ContainSubstring(param))
+					}
+				}
+			}
+		},
+			table.Entry("[test_id:28024] verify that hugepages size and count updated", chkCmdLine, []string{"default_hugepagesz=2M", "hugepagesz=2M", "hugepages=5"}, true),
```
cynepco3hahue, line 115: do you assume that you always will have the same order of tests under the spec?

dshchedr: it's a list of arguments, so I don't care about order. Here I'm going thru the list:

https://github.com/openshift-kni/performance-addon-operators/blob/3187fb61509fc378e66237be23338d4d89742883/functests/performance/updating_profile.go#L106

dshchedr: hm, I'm not 100% sure, but as I see tests under one "Describe" goes one by one.. at least i didn't see any issues with order after several running..

cynepco3hahue: In general yes, see - https://onsi.github.io/ginkgo/#spec-permutation, but it bad practice to assume it, and ginkgo gives the way to run it fully randomize, and it also recommended to run this way under the CI.

functests/performance/updating_profile.go
```diff
@@ -43,80 +47,91 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+			profile.Spec.NUMA = &performancev1alpha1.NUMA{
+				TopologyPolicy: &policy,
+			}
+			profile.Spec.RealTimeKernel = &performancev1alpha1.RealTimeKernel{
+				Enabled: &f,
+			}
 
-		for _, node := range workerRTNodes {
-			checkCmdLineArgs(node, newCmdLineArgs)
+			if profile.Spec.AdditionalKernelArgs == nil {
```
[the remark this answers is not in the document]
cynepco3hahue: hm for me it looks like different test cases:
```golang
It("should succeed to add new kernel argument", func() {})
It("should succeed to add and remove kernel argument in the same time", func() {})
```
and the first one you can run on every environment, for the second I assume that you will update the profile under the `BeforeEach` to have at least one kernel argument, in case if it does not have any and after you will run the test.

I know that it is very time-consuming, so maybe we can introduce `BeforeAll` and make preparation once for all tests, WDT?

functests/performance/updating_profile.go
dshchedr: I tried to find the way how to do all "updating" stuff before running any It() in this file, but unfortunately i didn't... 

Do you have any idea?

dshchedr:
> and the first one you can run on every environment, for the second I assume that you will update the profile under the BeforeEach to have at least one kernel argument, in case if it does not have any and after you will run the test.
> I know that it is very time-consuming, so maybe we can introduce BeforeAll and make preparation once for all tests, WDT?

Yeah, it's two different testcases.. but in our ci we have pre-configured kernelArguments, so we have coverage for that every run. 
I just didn't want to expand test running time. In d/s repo we have 2 worker nodes and additional modifications may lead to 40+ minutes for updating nodes.. I think we are good to go with what we have now.

Regarding to BeforeAll - you mean this **BeforeSuite**? Hm, it looks promising, I want to try that.

dshchedr:
> Regarding to BeforeAll - you mean this **BeforeSuite**? Hm, it looks promising, I want to try that.

em.. BeforeSuite does not work for me :(

_panic: You may only call BeforeSuite once!_

dshchedr:
@cynepco3hahue added BeforeAll, thanks for help

slintes:
fyi: no need for triggering retests before https://github.com/openshift-kni/performance-addon-operators/pull/164 is sorted out and merged :/

slintes:
nice :+1: 

/lgtm

cynepco3hahue:
I think to make things more robust we should replace https://github.com/openshift-kni/performance-addon-operators/blob/779ab90a9ffab7bf12f3e4fa0270ba088c1df5bb/functests/performance/updating_profile.go#L174

with something like
```
if err := testclient.Client.Get(context.TODO(), key, mcp); err != nil {
    return corev1.ConditionUnknown
}
```

PR #164, opened 2020-04-02
Fix CI (several issues)

slintes:
1. CI cluster was reverted to old one, so change registry URL image... again
2. OCP was rebased on k8s 1.18, which resulted in the following issues:
a) `ocp apply -f` fails to deploy ANY resource if ONE of them has an unknown kind. So apply all one by one, else our loop-until-succeed deployment doesn't work (profile can't be deployed because CRD doesn't exist in 1st iteration(s)
b) functests: `oc rsh` aka `kubectl exec` fails with `kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl kubectl exec [POD] -- [COMMAND] instead.`. So let's add some "--". Update: `oc rsh` still does not work, but `oc exec -i` does, so using that on everywhere.

### DIFF
--- a/hack/deploy.sh
+++ b/hack/deploy.sh
@@ -28,9 +28,15 @@ do
 
   # be verbose on last iteration only
   if [[ $iterations -eq $((max_iterations - 1)) ]] || [[ -n "${VERBOSE}" ]]; then
-    ${OC_TOOL} kustomize $feature_dir | envsubst | ${OC_TOOL} apply -f -
+    # WORKAROUND for https://github.com/kubernetes/kubernetes/pull/89539:
+    # oc / kubectl reject multiple manifests atm as soon as one "kind" in them does not exist yet
+    # so we need to apply one manifest by one
+    # since xargs' delimiter is limited to one char only or a control code, we replace the manifest delimiter "---"
+    # with a "vertical tab (\v)", which should never be used in (at least our) manifests.
+    # revert the sed | xargs steps when the fix landed in oc
+    ${OC_TOOL} kustomize $feature_dir | envsubst | sed "s|---|\v|g" | xargs -d '\v' -I {} bash -c "echo '{}' | ${OC_TOOL} apply -f -"
   else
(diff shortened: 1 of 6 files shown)


MarSik:
@slintes Is there a way to set out own env variable when the pipeline can't do that? We could then just change the env var somewhere (in Travis there was nice UI for that :) without touching our sources.

slintes:
> @slintes Is there a way to set out own env variable when the pipeline can't do that? We could then just change the env var somewhere (in Travis there was nice UI for that :) without touching our sources.

No, there isn't such a UI.
In a test you could add env vars to tests by changing the Job spec (would also need a PR), or use custom secrets which you could modify.
But: we are not in a test step here, but in the image build step. AFAIK we don't have any option here.

slintes:
/hold cancel
Let's unblock CI and discuss my fix for 2a) (2nd commit) in a follow up, at least it worked in an earlier CI run on this PR.

hack/deploy.sh
```diff
@@ -28,9 +28,13 @@ do
 
   # be verbose on last iteration only
   if [[ $iterations -eq $((max_iterations - 1)) ]] || [[ -n "${VERBOSE}" ]]; then
-    ${OC_TOOL} kustomize $feature_dir | envsubst | ${OC_TOOL} apply -f -
+    # oc / kubectl reject multiple manifests now as soon as one "kind" in them does not exist yet
+    # so we need to apply one manifest by one
+    # since xargs' delimiter is limited to one char only or a control code, we replace the manifest delimiter "---"
+    # with a "vertical tab (\v)", which should never be used in (at least our) manifests.
```
MarSik, line 34: Please mention the kubernetes issue here to make it clear this is a workaround for a bug only.

slintes: done

slintes:
one remaining test failure, will be handled in a follow up

/override ci/prow/e2e-gcp

PR #165, opened 2020-04-03
tests: better logs

fromanirh:
borrow/import a tiny part of the k8s e2e test framework
to log messages in our tests, and use it.

We want to test failures to be as easy as possible to troubleshoot,
and for this purpose we add logging to the most critical and common
operations (more to be added in future patches)

The logs will be visible only when/if the test fail.

### DIFF
--- a/functests/performance/performance.go
+++ b/functests/performance/performance.go
@@ -52,8 +52,8 @@ var _ = Describe("[rfe_id:27368]performance", func() {
 	BeforeEach(func() {
 		var err error
 		workerRTNodes, err = nodes.GetByRole(testclient.Client, testutils.RoleWorkerRT)
-		Expect(err).ToNot(HaveOccurred())
-		Expect(workerRTNodes).ToNot(BeEmpty())
+		Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("error looking for node with role %q: %v", testutils.RoleWorkerRT, err))
+		Expect(workerRTNodes).ToNot(BeEmpty(), fmt.Sprintf("no nodes with role %q found", testutils.RoleWorkerRT))
 		profile, err = profiles.GetByNodeLabels(
 			testclient.Client,
 			map[string]string{
--- a/functests/utils/nodes/nodes.go
+++ b/functests/utils/nodes/nodes.go
@@ -96,7 +96,9 @@ func ExecCommandOnMachineConfigDaemon(c client.Client, node *corev1.Node, comman
 		"--",
 	}
 	initialArgs = append(initialArgs, command...)
-	return exec.Command("oc", initialArgs...).CombinedOutput()
+	res, err := exec.Command("oc", initialArgs...).CombinedOutput()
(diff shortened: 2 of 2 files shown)


cynepco3hahue:
@fromanirh What the difference between this package and the klog, in general in case of failures ginkgo will show you all klog and fmt messages.
So adding something like `klog.Errorf()` should be enough.

yuvalk:
> @fromanirh What the difference between this package and the klog, in general in case of failures ginkgo will show you all klog and fmt messages.
> So adding something like `klog.Errorf()` should be enough.

Either way, having our own log util is the best way to ensure we can change it easily in the future.

yuvalk:
I'm all for better readability of our test reports

cynepco3hahue:
> Either way, having our own log util is the best way to ensure we can change it easily in the future.

I prefer to have the same format as k8s logs, just for standardization and to avoid to maintain our own log library. Again I will not the PR once I will see some real advantage of using our own library.

yuvalk:
> > Either way, having our own log util is the best way to ensure we can change it easily in the future.
> 
> I prefer to have the same format as k8s logs, just for standardization and to avoid to maintain our own log library. I will not block the PR once I will see some real advantage of using our own library.

yep, agree.
I thought this is in par with k8s
@fromanirh says it is taken from k8s...
Did I miss understood?

fromanirh:
> @fromanirh What the difference between this package and the klog, in general in case of failures ginkgo will show you all klog and fmt messages.
> So adding something like `klog.Errorf()` should be enough.

Uhm, perhaps I was too fast in importing e2e framework code. That could work as well, let me try it.

slintes:
> > @fromanirh What the difference between this package and the klog, in general in case of failures ginkgo will show you all klog and fmt messages.
> > So adding something like `klog.Errorf()` should be enough.
> 
> Uhm, perhaps I was too fast in importing e2e framework code. That could work as well, let me try it.

If it's just for logging, the framework might be too much overhead indeed.
But maybe there are more useful things inside. Didn't have the time yet to have a deeper look.

fromanirh:
> > > @fromanirh What the difference between this package and the klog, in general in case of failures ginkgo will show you all klog and fmt messages.
> > > So adding something like `klog.Errorf()` should be enough.
> > 
> > 
> > Uhm, perhaps I was too fast in importing e2e framework code. That could work as well, let me try it.
> 
> If it's just for logging, the framework might be too much overhead indeed.
> But maybe there are more useful things inside. Didn't have the time yet to have a deeper look.

No worries, I'll re-evaluate the PR soon(ish). Let's bear in mind I just want to add more logs to make test failures easier to understand, I don't want to (pseudo-)vendor k8s test framework code unless there is no other choice. So the current state of things is not really deliberate. The purpose is just to have more log messages.

fromanirh:
ok folks, that's the final form of this patch, ready for review.

yuvalk:
/lgtm

probably need to review all tests and add such msgs (ie - wherever we expect something, add appropriate msg when it doesnt, with values that aid debugging)

fromanirh:
| /test e2e-gcp
-- | --

PR #168, opened 2020-04-07
Added test - removing rt label

dshchedr:


slintes:
- you need to run `make generate` and commit changes
- can you also check that hugepages config is reverted?

gsr-shanks:
lgtm

gsr-shanks:
Thinking if we could move this to use as _AfterEach_ block.

dshchedr:
> Thinking if we could move this to use as _AfterEach_ block.

I don't think it's good idea.. AfterEach works after every It(), so we don't really want to run it so often

If you are talking about reverting all changes back after running tests from entire file (updating_profile.go) - i think it would be better to add new function AfterAll that will be running when all these tests are completed. Let's do that in another PR

dshchedr:
> * you need to run `make generate` and commit changes
thx, will do

> * can you also check that hugepages config is reverted?
hm, in this case I need to know what hugepages config was initially (before installing performance-addon-operator). As i see in my cluster - worker nodes does not have any hugepages by default, but can we be sure that this is always the case?

slintes:
> > * you need to run `make generate` and commit changes
> > * can you also check that hugepages config is reverted?
> 
> hm, in this case I need to know what hugepages config was initially (before installing performance-addon-operator). As i see in my cluster - worker nodes does not have any hugepages by default, but can we be sure that this is always the case?

good question, I would think so, maybe @cynepco3hahue can confirm?

slintes:
> > > * you need to run `make generate` and commit changes
> > > * can you also check that hugepages config is reverted?
> > 
> > 
> > hm, in this case I need to know what hugepages config was initially (before installing performance-addon-operator). As i see in my cluster - worker nodes does not have any hugepages by default, but can we be sure that this is always the case?
> 
> good question, I would think so, maybe @cynepco3hahue can confirm?

we handle this if it makes sense in a follow up

/lgtm
/approve

PR #170, opened 2020-04-08
RT Kernel is not installed when PerformanceProfile has no realTimeKernel field

gsr-shanks:
This test ensures that RT Kernel is not installed when the
realTimeKernel field does not exist in performance profile

### DIFF
--- a/functests/2_performance_update/updating_profile.go
+++ b/functests/2_performance_update/updating_profile.go
@@ -144,6 +144,28 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			}
 		})
 
+		It("[test_id:22764] verify that realTimeKernel is default when realTimeKernel field does not exist", func() {
+			conditionUpdating := machineconfigv1.MachineConfigPoolUpdating
+
+			if profile.Spec.RealTimeKernel == nil || *profile.Spec.RealTimeKernel.Enabled == true {
+				Skip("Skipping test - This test expects RT Kernel to be disabled. Found it to be enabled or nil.")
+			}
+
+			By("Applying changes in performance profile")
+			profile.Spec.RealTimeKernel = nil
+			Expect(testclient.Client.Update(context.TODO(), profile)).ToNot(HaveOccurred())
+
+			Expect(profile.Spec.RealTimeKernel).To(BeNil())
+			By("Checking that the updating MCP status will consistently stay false")
+			Consistently(func() corev1.ConditionStatus {
+				return getMcpStatus(testutils.RoleWorkerRT, conditionUpdating)
+			}, 30, 5).Should(Equal(corev1.ConditionFalse))
+
+			for _, node := range workerRTNodes {
+				Expect(nodes.ExecCommandOnNode(chkKernel, &node)).NotTo(ContainSubstring("PREEMPT RT"))
+			}
+		})
+
 		It("Reverts back all profile configuration", func() {
 			// BUG: CNF-385. Workaround - we have to remove hugepages first
 			profile.Spec.HugePages = nil
@@ -276,21 +298,26 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 	})
 })
 
-func waitForMcpCondition(mcpName string, conditionType machineconfigv1.MachineConfigPoolConditionType, conditionStatus corev1.ConditionStatus, timeout time.Duration) {
+func getMcpStatus(mcpName string, conditionType machineconfigv1.MachineConfigPoolConditionType) corev1.ConditionStatus {
 	mcp := &machineconfigv1.MachineConfigPool{}
 	key := types.NamespacedName{
 		Name:      mcpName,
 		Namespace: "",
 	}
-	Eventually(func() corev1.ConditionStatus {
-		err := testclient.Client.Get(context.TODO(), key, mcp)
-		Expect(err).ToNot(HaveOccurred())
-		for i := range mcp.Status.Conditions {
-			if mcp.Status.Conditions[i].Type == conditionType {
-				return mcp.Status.Conditions[i].Status
-			}
+	err := testclient.Client.Get(context.TODO(), key, mcp)
+	Expect(err).ToNot(HaveOccurred())
+
+	for i := range mcp.Status.Conditions {
+		if mcp.Status.Conditions[i].Type == conditionType {
+			return mcp.Status.Conditions[i].Status
 		}
-		return corev1.ConditionUnknown
+	}
+	return corev1.ConditionUnknown
+}
+
+func waitForMcpCondition(mcpName string, conditionType machineconfigv1.MachineConfigPoolConditionType, conditionStatus corev1.ConditionStatus, timeout time.Duration) {
+	EventuallyWithOffset(1, func() corev1.ConditionStatus {
+		return getMcpStatus(mcpName, conditionType)
 	}, timeout*time.Minute, 30*time.Second).Should(Equal(conditionStatus))
 }
 


functests/performance/updating_profile.go
```diff
@@ -133,6 +133,29 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
 				}
 			}
 		})
+
+		It("[test_id:22764] verify that realTimeKernerl is default when realTimeKernel field does not exist", func() {
```
slintes, line 137: typo: Kernerl

functests/performance/updating_profile.go
```diff
@@ -133,6 +133,29 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+			timeout := time.Duration(len(workerRTNodes) * 10)
+
+			profile.Spec.RealTimeKernel = nil
+
+			By("Applying changes in performance profile")
+			Expect(testclient.Client.Update(context.TODO(), profile)).ToNot(HaveOccurred())
+
+			if profile.Spec.RealTimeKernel == nil {
```
slintes, line 145: I would make this an expectation: if it's not nil, it's a test failure, otherwise go on

functests/performance/updating_profile.go
```diff
@@ -133,6 +133,29 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+				sum := 0
```
slintes, line 146: what is this used for?

functests/performance/updating_profile.go
```diff
@@ -133,6 +133,29 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+				for i := 1; i < 5; i++ {
```
slintes, line 147: I think Gomega's `https://onsi.github.io/gomega/#consistently` is what you might want to use here

functests/performance/updating_profile.go
```diff
@@ -133,6 +133,29 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
 				}
 			}
 		})
+
+		It("[test_id:22764] verify that realTimeKernerl is default when realTimeKernel field does not exist", func() {
+			timeout := time.Duration(len(workerRTNodes) * 10)
```
dshchedr, line 138: hm, your timeout is 10 minutes. If I understand correctly, your goal - check that MCP **does not start updating** after modifying profile, but in your case you are waiting for 10 minutes to get right status (waitForMcpCondition() is "eventually", so it will show you error only if timeout has gone. 10 minutes can be enough for mcp to start update and complete it).

I see 2 options:
Either you reducing timeout (something very small that will not be enough for MCP (1 minute?)) 
Or, you should get MCP status directly from test - without using waitForMcpCondition() function

functests/performance/updating_profile.go
```diff
@@ -171,6 +194,7 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
 		Expect(execCommandOnWorker(chkCmdLine, newWorkerNode)).To(ContainSubstring("tuned.non_isolcpus"))
 		Expect(execCommandOnWorker(chkKernel, newWorkerNode)).ToNot(ContainSubstring("PREEMPT RT"))
 	})
+
```
dshchedr, line 197: how about removing that? :)

functests/performance/updating_profile.go
slintes: no, `waitForMcpCondition()` returns immediately without error, because the `updating` condition is `false` as expected. The timeout isn't relevant (as long as the test passes as expected). But by doing this in a loop we try to make sure that the update doesn't start within 5 iterations * 5 sec sleep = 25 seconds. That's what I understand at least.

However, as mentioned in my earlier review, there is a func for such an expectation: https://onsi.github.io/gomega/#consistently. Works similar to `Eventually`, but the other way around: instead of waiting for a change until timeout, make sure something doesn't change for some time. I never used that before, but the description sounds good for this usecase.

dshchedr: yes, Shanks want to be sure that after modifying profile update was not started in 25 seconds. And if it is so - waitForMcpCondition() returns immediately. But what if updating really started? waitForMcpCondition() will be waiting inside Eventually loop until timeout has reached (10 minutes)

functests/performance/updating_profile.go
gsr-shanks: Fixed.

functests/performance/updating_profile.go
gsr-shanks: Fixed.

functests/performance/updating_profile.go
gsr-shanks: Sure, I will check this out. Thanks.

functests/performance/updating_profile.go
gsr-shanks: The only case where the updating would start is when realTimeKernel is changed to 'true' in any of the previous steps before this test case which is unlikely since the test itself is for updating profile. 

I can put another if condition at the start of the test `if profile.Spec.RealTimeKernel == true` and set it to `False` before running the test, but I think that would be an overkill here. WDYT?

slintes: > But what if updating really started? 

ah right, we might not catch the negative case, when the update starts but finishes within 10 minutes. But this will also be fixed by using `Consistently` instead of loop + `Eventually` (in the wait func)

> I can put another if condition at the start of the test if profile.Spec.RealTimeKernel == true and set it to False

What about failing the test if it's true. Less overkill, but would ensure that we catch if setting changes in test setup / earlier tests.

functests/performance/updating_profile.go
gsr-shanks: Done.

functests/performance/updating_profile.go
gsr-shanks: > > I can put another if condition at the start of the test if profile.Spec.RealTimeKernel == true and set it to False
> 
> What about failing the test if it's true. Less overkill, but would ensure that we catch if setting changes in test setup / earlier tests.

On second thought, don't think it would be overkill since it would be skipped anyway. :-)

dshchedr: I would change _if profile.Spec.RealTimeKernel == true_ to  _Expect(profile.Spec.RealTimeKernel).To(smth)_ 
IMHO, in tests Expect is more readable then If

slintes:
in case you continue while others are on holidays

/ok-to-test

functests/performance/updating_profile.go
```diff
@@ -133,6 +133,32 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
 				}
 			}
 		})
+
+		It("[test_id:22764] verify that realTimeKernel is default when realTimeKernel field does not exist", func() {
+			conditionType := machineconfigv1.MachineConfigPoolUpdating
+
+			// Fail test if realTimeKernel is already installed.
```
dshchedr, line 140: May be it's better to skip test instead of failing? If rt-kernel is enabled initially - it is not an error of this test

functests/performance/updating_profile.go
```diff
@@ -133,6 +133,32 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+			for _, node := range workerRTNodes {
+				Expect(execCommandOnWorker(chkKernel, &node)).NotTo(ContainSubstring("PREEMPT RT"))
+			}
+
+			profile.Spec.RealTimeKernel = nil
+			By("Applying changes in performance profile")
+			Expect(testclient.Client.Update(context.TODO(), profile)).ToNot(HaveOccurred())
+
+			if profile.Spec.RealTimeKernel != nil {
```
dshchedr, line 149: Not sure you really need this If/else, how about using:
Expect(profile.Spec.RealTimeKernel).To(BeNil())

functests/performance/updating_profile.go
```diff
@@ -215,6 +241,23 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
 	})
 })
 
+func getMcpStatus(conditionType machineconfigv1.MachineConfigPoolConditionType) corev1.ConditionStatus {
```
dshchedr, line 244: This function looks good for me, how about modifying waitForMcpCondition() for calling this function inside? It will reduce duplicate code.

functests/performance/updating_profile.go
```diff
@@ -133,6 +133,32 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+
+		It("[test_id:22764] verify that realTimeKernel is default when realTimeKernel field does not exist", func() {
+			conditionType := machineconfigv1.MachineConfigPoolUpdating
+
+			// Fail test if realTimeKernel is already installed.
+			for _, node := range workerRTNodes {
+				Expect(execCommandOnWorker(chkKernel, &node)).NotTo(ContainSubstring("PREEMPT RT"))
+			}
+
+			profile.Spec.RealTimeKernel = nil
```
dshchedr, line 145: IMHO it's a part of "applying changes to proifle", it's better to put it below By()

functests/performance/updating_profile.go
```diff
@@ -133,6 +133,32 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+			By("Applying changes in performance profile")
+			Expect(testclient.Client.Update(context.TODO(), profile)).ToNot(HaveOccurred())
+
+			if profile.Spec.RealTimeKernel != nil {
+				Expect(err).ToNot(HaveOccurred())
+			} else {
+				By("Checking that the updating MCP status will consistently stay false")
+				Consistently(func() corev1.ConditionStatus {
+					return getMcpStatus(conditionType)
```
dshchedr, line 154: Just my thought. 
How about changing variable name "conditionType to something more readable, like "conditionUpdating"?

functests/performance/updating_profile.go
gsr-shanks: Done.

functests/performance/updating_profile.go
gsr-shanks: Done.

functests/performance/updating_profile.go
gsr-shanks: Done

functests/performance/updating_profile.go
gsr-shanks: Done.

functests/performance/updating_profile.go
gsr-shanks: Done.

functests/performance/updating_profile.go
```diff
@@ -133,6 +133,28 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+
+			By("Applying changes in performance profile")
+			profile.Spec.RealTimeKernel = nil
+			Expect(testclient.Client.Update(context.TODO(), profile)).ToNot(HaveOccurred())
+
+			Expect(profile.Spec.RealTimeKernel).To(BeNil())
+			By("Checking that the updating MCP status will consistently stay false")
+			Consistently(func() corev1.ConditionStatus {
+				return getMcpStatus(testutils.RoleWorker, conditionUpdating)
+			}, 5, 5).Should(Equal(corev1.ConditionFalse))
```
slintes, line 152: the 1st time arg is the duration, the second the interval. So I think the 1st should be something like 30?

gsr-shanks: I thought 25 seconds should be good enough. Do we want to change it? Maybe 15, 4 ?

gsr-shanks: Done.

slintes:
cluster setup failed again...

/retest

slintes:
https://github.com/openshift-kni/performance-addon-operators/pull/177

slintes:
@gsr-shanks you need to rebase

slintes:
cluster was getting unresponsive during tests...

/retest

gsr-shanks:
cluster setup failed again

/retest

dshchedr:
> /retest

I think you need to rebase your code:

` vet: functests/2_performance_update/updating_profile.go:165:12: undeclared name: execCommandOnWorker`

slintes:
cluster setup failed....

/retest

gsr-shanks:
Cluster failed setup.

/retest

gsr-shanks:
Cluster failed setup.

/retest

slintes:
For some reason the MCP is not getting into Updated status for the update tests. The wait method was refactored, but I don't see what might go wrong there...

And because the wait fails, in `utils.BeforeAll()` `first` is never set to `false`, that's why we wait for the MCP for _every_ test, although it should do only once (but that is ok IMHO).

And then, because we wait for 20 minutes so often, we run into a prow timeout (4 hours), and the cluster is shut down. That very probably explains the later API call failures. So at least we don't have to care for those...

If this still won't work, I will cherry pick this into my PR and see what happens there...

/retest

dshchedr:
> For some reason the MCP is not getting into Updated status for the update tests. The wait method was refactored, but I don't see what might go wrong there...
> And because the wait fails, in `utils.BeforeAll()` `first` is never set to `false`, that's why we wait for the MCP for _every_ test, although it should do only once (but that is ok IMHO).
> And then, because we wait for 20 minutes so often, we run into a prow timeout (4 hours), and the cluster is shut down. That very probably explains the later API call failures. So at least we don't have to care for those...
> If this still won't work, I will cherry pick this into my PR and see what happens there...

 I tried it yesterday on my local environment and it took 30+ minutes for MCP.. I thought it was the issue with my cluster (I had some other, not related to this PR problems) and wanted to try this PR  again after rebuilding cluster.

gsr-shanks:
I tried it locally again on my cluster and I see what @dshchedr has reported. There is something wrong with the wait method I guess. I will look further.

gsr-shanks:
> For some reason the MCP is not getting into Updated status for the update tests. The wait method was refactored, but I don't see what might go wrong there...

The MCP does get into 'Updated' status, but the func doesn't return it.

gsr-shanks:
waitForMcpCondition is expected to return MCP Condition status. However, with the refactoring, it does not and that is why we hit timeout and by then the Update has already happened.

I will amend to fix this.

gsr-shanks:
If this doesn't work, let's merge #189 and I will rebase to use utils/mcp.

slintes:
> waitForMcpCondition is expected to return MCP Condition status

I don't see that. It didn't before your refactoring, and the return value isn't used anywhere.

> let's merge #189 and I will rebase to use utils/mcp

ok

gsr-shanks:
> > waitForMcpCondition is expected to return MCP Condition status
> 
> I don't see that. It didn't before your refactoring, 

It did here https://github.com/openshift-kni/performance-addon-operators/pull/170/files#diff-382a12f07313e5f630cc3bb0e1ca129aL290

> and the return value isn't used anywhere.
> 

Yeah, I was looking for where it is used. Couldn't find.

slintes:
> > > waitForMcpCondition is expected to return MCP Condition status
> > 
> > 
> > I don't see that. It didn't before your refactoring,
> 
> It did here https://github.com/openshift-kni/performance-addon-operators/pull/170/files#diff-382a12f07313e5f630cc3bb0e1ca129aL290

That is the return of the anonymous func in `Eventually(func(){...}, ...)`, which is evaluated by `Should(Equal(conditionStatus))`. That is what you refactored into the `getMcpStatus` func, which correctly returns the status. It was not returned by the `wait...` func.

slintes:
if you edit it, can you change `Eventually` to `EventuallyWithOffset(1, ...)` please? That way we can better see from where it is called (updating or updated).

gsr-shanks:
> > It did here https://github.com/openshift-kni/performance-addon-operators/pull/170/files#diff-382a12f07313e5f630cc3bb0e1ca129aL290
> 
> That is the return of the anonymous func in `Eventually(func(){...}, ...)`, which is evaluated by `Should(Equal(conditionStatus))`. That is what you refactored into the `getMcpStatus` func, which correctly returns the status. It was not returned by the `wait...` func.

Ah right! My bad :-(

slintes:
I'm running the PR on my cluster. The MCP is Updated, but tests are still waiting. Something does not work here. But I don't see what yet... investigating.

slintes:
Found it!
From the doc of Eventually: `Eventually will poll the passed in function (which must have *zero-arguments* and at least one return value)...` 
`getMcpStatus` has arguments...
This works:
```
	EventuallyWithOffset(1, func() corev1.ConditionStatus {
		return getMcpStatus(mcpName, conditionType)
	}, timeout*time.Minute, 30*time.Second).Should(Equal(conditionStatus))
```
But now I see your new test failing... investigating further

functests/2_performance_update/updating_profile.go
```diff
@@ -144,6 +144,28 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+			}
+
+			By("Applying changes in performance profile")
+			profile.Spec.RealTimeKernel = nil
+			Expect(testclient.Client.Update(context.TODO(), profile)).ToNot(HaveOccurred())
+
+			Expect(profile.Spec.RealTimeKernel).To(BeNil())
+			By("Checking that the updating MCP status will consistently stay false")
+			Consistently(func() corev1.ConditionStatus {
+				return getMcpStatus(testutils.RoleWorker, conditionUpdating)
```
slintes, line 161: I think the should be `testutils.RoleWorkerRT`?
(on my cluster the worker MCP was updating for another reason, which is a funny coincidence, because otherwise I would not have spotted this one)

gsr-shanks: Yes, that should be testutils.RoleWorkerRT

slintes:
it's green! :)

/lgtm
/approve

PR #173, opened 2020-04-12
Update NTO dependencies against oc 4.5

yanirq:


cynepco3hahue:
we should update all our k8s dependencies to 1.18, but it should be done under another PR

yanirq:
retest

PR #175, opened 2020-04-15
csv-generator: add option to read description

fromanirh:
Allow the generator-tool to optionally read the full description
from a separate text file instead of hardcoding it.
The content of the file will be copied verbatim in the
`csv.Spec.Description` field.


fromanirh:
uhm, still self-approval. Not good.

PR #176, opened 2020-04-15
Add missing performance tags.

fedepaol:
This will make it possible to filter tests when consuming as vendored dependencies.


fromanirh:
/lgtm
adding just to double check @dshchedr and @gsr-shanks are OK with this

dshchedr:
> /lgtm
> adding just to double check @dshchedr and @gsr-shanks are OK with this

sure, I'm fine with that. Thanks

PR #178, opened 2020-04-15
Restore performance profile parameters after running tests

dshchedr:
Tests in updating_profile.go make big changes in profile, so we need to restore them back for right functionality of the cluster

### DIFF
--- a/functests/performance/updating_profile.go
+++ b/functests/performance/updating_profile.go
@@ -2,6 +2,7 @@ package performance
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
@@ -11,6 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
@@ -27,7 +29,8 @@ const (
 
 var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", func() {
 	var workerRTNodes []corev1.Node
-	var profile *performancev1alpha1.PerformanceProfile
+	var profile, initialProfile *performancev1alpha1.PerformanceProfile
+	var timeout time.Duration
 	var err error
 
 	chkKernel := []string{"uname", "-a"}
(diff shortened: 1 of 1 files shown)


functests/performance/updating_profile.go
```diff
@@ -58,10 +62,9 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
 
 		// Modify profile and verify that MCO successfully updated the node
 		testutils.BeforeAll(func() {
-			// timeout should be based on the number of worker nodes
-			timeout := time.Duration(len(workerRTNodes) * mcpUpdateTimeout)
-
 			By("Modifying profile")
```
gsr-shanks, line 65: Let's move this By below initialProfile.

functests/performance/updating_profile.go
```diff
@@ -133,6 +136,19 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
 				}
 			}
 		})
+
+		It("Reverts back all profile configuration", func() {
+			// BUG: CNF-385. Workaround - we have to remove hugepages first
```
slintes, line 141: is this still an issue with https://github.com/kubernetes/kubernetes/pull/82820 being merged in k8s 1.18 and OCP being rebased? (did not look into details, just asking)

slintes: @cynepco3hahue do you know? :)

cynepco3hahue: it requires to enable the specific feature gate until this PR(https://github.com/openshift/api/pull/604) is merged the feature still not usable

functests/performance/updating_profile.go
dshchedr: not really agree.. as for me - it's more readable: save initialProfile before making changes

functests/performance/updating_profile.go
```diff
@@ -133,6 +138,26 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+			profile.Spec.HugePages = nil
+			Expect(testclient.Client.Update(context.TODO(), profile)).ToNot(HaveOccurred())
```
cynepco3hahue, line 145: Can you try to fetch the updated profile one via `Get` before the update, I hope it will solve the problem.

slintes: using `Patch` is eben better, see #179 as example

cynepco3hahue: hh I unsure if it more readable

dshchedr: I don't need to get new profile **here** because I already got it from BeforeEach()

But later (starting from line 149) I have 2 options: either I fetch profile and use Update() or I'm just using Patch()

With Update() we will have less code but one additional request to the cluster.

slintes: > I don't need to get new profile **here** because I already got it from BeforeEach()

ah yes :+1: 

> 
> But later (starting from line 149) I have 2 options: either I fetch profile and use Update() or I'm just using Patch()
> 
> I'm not really sure which way is better.
> With Update() we will have less code but one additional request to the cluster.

`Patch()` :+1: 
As discussed in the context of the other PR: while in a test it's not that relevant to reduce API calls, I consider using `Patch()` best practice, and we should use it everywhere now where needed.

/lgtm
/approve

PR #183, opened 2020-04-20
Execute performance and performance update tests in 1 run.

slintes:
- Refactored update tests into their own package.
- Refactored `ExecCommandOnNode` func into the `utils/nodes` package, so it can be reused from multiple other packages.
- Splitted the old TestSuite into two, one for the "normal" performance tests, the other for the updates.
- Run both suites from one ginkgo command in `run-functests.sh`

This prevents confusion about skipped tests in the logs, ~and results in 1 unit report only~. The latter isn't possible unfortunately: each test suite has its own junit reported which always creates a new report file.

Let's see if this works as expected.
/hold

/cc @MarSik @yuvalk @dshchedr @gsr-shanks


yuvalk:
Great that you added all those expect explanations !

yuvalk:
could you split the commits though, between the one dealing with the suites and the one dealing with adding more explainability about failures/skipping/etc?

I think it will help a lot both for readability and future understanding as of reasoning

slintes:
> could you split the commits though, between the one dealing with the suites and the one dealing with adding more explainability about failures/skipping/etc?
> 
> I think it will help a lot both for readability and future understanding as of reasoning

It's all connected... splitting the testsuite into two needed moving the update test into its own package, which needed the refactoring of the `ExecuteOnNode` func, which needed the error checks on the calling side, and there I just used the failure comments... I updated the PR description, I hope that's enough.

yuvalk:
> > could you split the commits though, between the one dealing with the suites and the one dealing with adding more explainability about failures/skipping/etc?
> > I think it will help a lot both for readability and future understanding as of reasoning
> 
> It's all connected... splitting the testsuite into two needed moving the update test into its own package, which needed the refactoring of the `ExecuteOnNode` func, which needed the error checks on the calling side, and there I just used the failure comments... I updated the PR description, I hope that's enough.

If you look at this from the opposite direction, these are still 2 unrelated things, or even 3:
1. the change to ExecuteOnNode
2. the adding of the msgs (can be combined with 1)
3. the splitting of the suites

functests/README.txt
```diff
@@ -0,0 +1,5 @@
+HEADS UP!
+
+We have 2 test suites here, one running the "normal" performance tests, the other one performance profile update tests.
+The latter should run AFTER the former, and tests are executed in order of filenames.
```
gsr-shanks, line 4: Which would mean if I create test_suite_performance_status_test.go it would get executed before test_suite_performance_update_test.go right? The status tests requires the MCP to be in degraded state, hence it would be better to execute these tests at the end.

slintes: Yes. Maybe I should add a number into the name like 
- test_suite_1_performance_test.go
- test_suite_2_performance_update_test.go
- test_suite_3_performance_status_test.go

to make it more explicit?

slintes:
either the logs are messed up, or it doesn't work as expected and tested in a simple experiment... investigating

functests/README.txt
slintes: I had to move the test suites into the subdirectories, so I added numbers to the directory name

slintes:
rebased and refactored: having multiple test suites in the same package didn't work, so I had to move them into the subdirectories.

slintes:
besides the test failures, this looks as expected now:
- clean log file running the performance test suite first, and the update test suite second, without skips of the other one.
- 2 unit reports with meaningful name (just fixed the trailing dot in the filename; unfortunately each suite creates its own unit report)

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift-kni_performance-addon-operators/183/pull-ci-openshift-kni-performance-addon-operators-master-e2e-gcp/711

/hold cancel

fromanirh:
very nice work. lgtm

PR #184, opened 2020-04-21
Remove nohz_full from kargs

yanirq:
Remove nohz_full kernel argument that was placed as a default.


yanirq:
/hold cancel
the request is to remove the arg from default. removing hold

PR #185, opened 2020-04-21
 Removing leftovers after running nodeSelector and label removing tests

dshchedr:
In this PR I wanted to combine nodeSelector and label removing tests. First of all, it will save us from one additional MCP update during cleaning leftovers. And second - I don't think it's good idea to remove worker-rt label in tests (at the end we can get cluster without any worker-rt nodes, so any follow-up tests will be failed). If we are want to return all cluster stuff back to initial state after running our tests - it's better to "play" with worker-test label and skip this test if we don't have it.

### DIFF
--- a/functests/2_performance_update/updating_profile.go
+++ b/functests/2_performance_update/updating_profile.go
@@ -37,6 +37,8 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 	chkCmdLine := []string{"cat", "/proc/cmdline"}
 	chkKubeletConfig := []string{"cat", "/rootfs/etc/kubernetes/kubelet.conf"}
 
+	nodeLabel := map[string]string{fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerRT): ""}
+
 	BeforeEach(func() {
 		workerRTNodes, err = nodes.GetByRole(testclient.Client, testutils.RoleWorkerRT)
 		Expect(err).ToNot(HaveOccurred())
@@ -46,9 +48,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 
 		profile, err = profiles.GetByNodeLabels(
 			testclient.Client,
-			map[string]string{
-				fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerRT): "",
-			},
+			nodeLabel,
 		)
 		Expect(err).ToNot(HaveOccurred())
 	})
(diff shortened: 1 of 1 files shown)


functests/performance/updating_profile.go
```diff
@@ -204,43 +200,60 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
+		It("Reverts back nodeSelector and cleaning up leftovers", func() {
+			nodeSelector := fmt.Sprintf(`"%s/%s": ""`, testutils.LabelRole, testutils.RoleWorkerRT)
+			Expect(testclient.Client.Patch(context.TODO(), profile,
+				client.ConstantPatch(
+					types.JSONPatchType,
+					[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec/nodeSelector", "value": {%s} }]`, nodeSelector)),
+				),
+			)).ToNot(HaveOccurred())
+			Expect(testclient.Client.Delete(context.TODO(), mcp)).ToNot(HaveOccurred())
+			waitForMcpCondition(testutils.RoleWorkerRT, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue, timeout)
```
slintes, line 255: don't we need to wait for `Updating` first?
maybe even factor this out, it is used at several places:
1. wait for updating
2. wait for updated
3. check node readiness (not sure if this needed at all? when a node isn't ready, the MCP can't be not updated, can it?)

functests/performance/updating_profile.go
```diff
@@ -204,43 +200,60 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
...
-		Expect(testclient.Client.Update(context.TODO(), &node)).ToNot(HaveOccurred())
-		waitForMcpCondition(testutils.RoleWorker, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue, mcpUpdateTimeout)
+		It("[test_id:27484]Verifies that node is reverted to plain worker when the extra labels are removed", func() {
+			By("Deleting cnf labels from the node")
+			for l := range profile.Spec.NodeSelector {
+				delete(newCnfNode.Labels, l)
+			}
+			label, err := json.Marshal(newCnfNode.Labels)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(testclient.Client.Patch(context.TODO(), newCnfNode,
```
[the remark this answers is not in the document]
dshchedr: :) I thought so too.. but got an error.

Here I'm updating **node**, not the profile.. I don't get new version of node from BeforeEach(), so either I need to get it here or just use patch

functests/performance/updating_profile.go
```diff
@@ -204,43 +200,60 @@ var _ = Describe("[rfe_id:28761] Updating parameters in performance profile", fu
 			By("Waiting when MCP finishes updates and verifying new node has updated configuration")
 			waitForMcpCondition(newRole, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue, mcpUpdateTimeout)
 
-			_, err = nodes.ExecCommandOnMachineConfigDaemon(testclient.Client, newWorkerNode, []string{"ls", "/rootfs/" + testutils.PerfRtKernelPrebootTuningScript})
+			_, err = nodes.ExecCommandOnMachineConfigDaemon(testclient.Client, newCnfNode, []string{"ls", "/rootfs/" + testutils.PerfRtKernelPrebootTuningScript})
 			Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("cannot find the file %s", testutils.PerfRtKernelPrebootTuningScript))
-			Expect(execCommandOnWorker(chkKubeletConfig, newWorkerNode)).To(ContainSubstring("topologyManagerPolicy"))
-			Expect(execCommandOnWorker(chkCmdLine, newWorkerNode)).To(ContainSubstring("tuned.non_isolcpus"))
+			Expect(execCommandOnWorker(chkKubeletConfig, newCnfNode)).To(ContainSubstring("topologyManagerPolicy"))
+			Expect(execCommandOnWorker(chkCmdLine, newCnfNode)).To(ContainSubstring("tuned.non_isolcpus"))
```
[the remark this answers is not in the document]
dshchedr: We can check it, but in this case we need to be sure we should have it in any scenario. I tried to avoid verifications that depend on profile settings (customer can have RTKernel=False), I wanted to check only 100% existing parameters...

functests/performance/updating_profile.go
dshchedr: I'm not sure we really need this waitForMcpCondition() here at all..

As I saw, when I changed nodeSelector in profile - worker-rt node (that initially had been updated) was still remained updated. So when I'm returning nodeSelector back - mcp does not start any updates since the node is already updated.
  
Actually I'm not sure may be it is not expected behavior that node is keeping updated after changing nodeSelector and it should be reverted back to **worker** state.. (or may be it takes some time for operator to realize that the node should be reverted back)

functests/performance/updating_profile.go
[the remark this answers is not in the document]
slintes: ok, makes sense :+1:

functests/performance/updating_profile.go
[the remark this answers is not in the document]
slintes: > Here I'm updating node, not the profile

ups... :)

functests/performance/updating_profile.go
slintes: > when I changed nodeSelector in profile - worker-rt node (that initially had been updated) was still remained updated

Oh... I think that would be a bug then?
Should we check in `[test_id:28440]Verifies that nodeSelector can be updated in performance profile` that the original mcp updates and reverts it's nodes to normal worker...? :thinking:

dshchedr: Verified manually: It took more than 10 minutes for MCP to start reverting back worker-rt node after changing NodeSelector from worker-rt to worker-test..

slintes: wow
good to know it does (and we should test it)
but strange that it is so slow
thx for testing :+1:

dshchedr: looks like it was enough time for completing tests before returning nodeSelector to initial state.. that's why i didn't have additinal MCP updates

slintes: not sure if we miss something here... but it can be handled in a follow up if needed

/lgtm

slintes:
and
/approve

functests/performance/updating_profile.go
dshchedr: Thanks, I'll keep an eye on it.

PR #186, opened 2020-04-22
Better logs for command executions

slintes:
Refactored all execs to use the new `ExecAndLogCommand()` func, and cleaned up some related code.


slintes:
@gsr-shanks please also have a look :)

gsr-shanks:
@slintes Should we should also enable '-trace' to the ginkgo command

slintes:
> @slintes Should we should also enable '-trace' to the ginkgo command

I tried it shortly in the context of #183 , but I didn't found it that useful.
Did you try it? If you find it useful I won't block the idea :slightly_smiling_face: 

/retest

PR #187, opened 2020-04-23
Add support for running tests in a disconnected environment.

fedepaol:
Refactoring the images we use for tests, and provide a way to override the registry
they are fetched from.

The images are built from the dockerfiles being added in https://github.com/openshift-kni/cnf-features-deploy/pull/151 
They are currently being pushed to quay.io/fpaoline , a followup will mirror them on quay.io/openshift-kni

### DIFF
--- a/functests/utils/images/images.go
+++ b/functests/utils/images/images.go
@@ -1,0 +1,52 @@
+package images
+
+import (
+	"fmt"
+	"os"
+
+	. "github.com/onsi/gomega"
+)
+
+var registry string
+var images map[string]imageLocation
+
+const (
+	// TestUtils is the image name to be used to retrieve the test utils image
+	TestUtils = "testutils"
+	// Stresser is the image name to be used to retrieve the stresser image
+	Stresser = "stresser"
+)
+
+func init() {
+	registry = os.Getenv("IMAGE_REGISTRY")
+
+	images = map[string]imageLocation{
+		TestUtils: {
+			name:    "cnftest-utils",
+			registy: "quay.io/fpaoline/",
+			version: "v1.0",
+		},
+		Stresser: {
+			name:    "stresser",
+			registy: "quay.io/fpaoline/",
+			version: "v1.0",
+		},
+	}
+}
+
+type imageLocation struct {
+	name    string
+	registy string
+	version string
+}
+
+// For returns the image to be used for the given key
+func For(name string) string {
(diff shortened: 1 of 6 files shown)


functests/utils/images/images.go
```diff
@@ -0,0 +1,50 @@
...
+	name    string
+	registy string
+	version string
+}
+
+// For returns the image to be used for the given key
+func For(name string) string {
+	img, ok := images[name]
+	if !ok {
+		log.Printf("Image for %s not found", name)
```
MarSik, line 44: Should the if return as well? Otherwise it prints the log, but continues to the Sprintfs, right?

fedepaol: Aw right, it was a last second addition for better feedback.
Thanks!

fedepaol: Well technically it would have worked in any case, because `img` would contain zero-valued strings, but it was not intentional :-)

fromanirh:
lgtm

functests/utils/images/images.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+var registry string
+var images map[string]imageLocation
+
+func init() {
+	registry = os.Getenv("IMAGE_REGISTRY")
+
+	images = map[string]imageLocation{
+		"sctp": {
```
slintes, line 16: I don't think we need this one?

functests/utils/images/images.go
```diff
@@ -0,0 +1,51 @@
...
+			name:    "sctptester",
+			registy: "quay.io/fpaoline/",
+			version: "v1.0",
+		},
+		"testutils": {
```
slintes, line 21: maybe worth making these keys constants?

functests/utils/images/images.go
slintes: Now, if an image is not found, I assume we will run into funny errors when trying to create a pod with an empty image field. I would prefer to catch this before creating the pod, by failing the test immediately.

functests/utils/images/images.go
fedepaol: No, we don't.
I wasn't sure to maintain the same functions equal across the different repos, but they already drifted so I assume it's better to clear.

functests/utils/images/images.go
fedepaol: will do.

PR #189, opened 2020-04-25
Create PerformanceProfile and wait for MCP updated in test code

slintes:
In order to be able to provide a test binary for CNF compliance tests, we want to configure the cluster in test code, before running the actual tests.

- renamed `worker-rt` to `worker-cnf` everywhere: this reflects that the realtime kernel is only 1 of many features of the operator, and aligns with the feature deploy repository. The actual role name is still configurable, but heads up, the env var name changed to `ROLE_WORKER_CNF`!
- also made the default profile name "performance" (still configurable with `PERF_TEST_PROFILE`)
- left an example profile manifest in a new kustomize dir for easier local tests and deployments
- some code cleanup and refactorings

@fedepaol: do we want to create the MCP in test code as well? If so, here or in the test suite in the feature deploy repo? (My opinion: leave it to the cluster admin, same for labelling the nodes)

/cc @dshchedr


slintes:
/hold

let #170 merge first, I don't want it to need to be rebased again

fedepaol:
> @fedepaol: do we want to create the MCP in test code as well? If so, here or in the test suite in the feature deploy repo? (My opinion: leave it to the cluster admin, same for labelling the nodes)
> 
> /cc @dshchedr

It makes sense to leave it to the cluster admin, also because other tests also depend on it.

functests/0_config/test_suite_performance_config_test.go
```diff
@@ -0,0 +1,24 @@
...
+package __performance_config_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"testing"
+
+	ginkgo_reporters "kubevirt.io/qe-tools/pkg/ginkgo-reporters"
+
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/junit"
```
fedepaol, line 12: You did not import the ` __performance_config` here, so it won't be executed.

functests/utils/mcps/mcps.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	return mcps.Items, nil
+}
+
+// GetByName returns the MCP with the specified name
+func GetByName(name string) (*machineconfigv1.MachineConfigPool, error) {
+	mcp := &machineconfigv1.MachineConfigPool{}
+	key := types.NamespacedName{
+		Name:      name,
+		Namespace: "",
```
fedepaol, line 49: Just a nit, but no need to specify Namespace if it's empty string.

fedepaol:
Another questions:
here we run tests with 

`GOFLAGS=-mod=vendor ginkgo $NO_COLOR --v -r --keepGoing -requireSuite functests -- -junitDir /tmp/artifacts`

Will it catch also the new suite?

functests/0_config/test_suite_performance_config_test.go
slintes: Seems it is not needed, everything is executed as expected. I guess `*_test` packages import their `*` package implicitly?

slintes:
> Another questions:
> here we run tests with
> 
> `GOFLAGS=-mod=vendor ginkgo $NO_COLOR --v -r --keepGoing -requireSuite functests -- -junitDir /tmp/artifacts`
> 
> Will it catch also the new suite?

Thanks for the quick review!
Yes, the `-r` will make ginkgo executing all suites in `functests` recursively.

functests/utils/mcps/mcps.go
slintes: Fair point, but actually I like stating explicitly that no namespace is needed. But there is a nicer way to do that, will change it to `metav1.NamespaceNone` (which is `""`, but nicer), which we use at other places as well.

functests/0_config/test_suite_performance_config_test.go
fedepaol: Ah yes, that's it. I did not notice they are in the same folder.

fedepaol:
/lgtm
@slintes this needs rebase, will re-approve after

PR #193, opened 2020-04-27
Fix for manual profile and cleaning script

dshchedr:
added nodeSelector to manual profile

 updated cleaning script - now it should go thru all existing profiles and clean cluster accordingly

### DIFF
--- a/hack/clean-deploy.sh
+++ b/hack/clean-deploy.sh
@@ -2,15 +2,21 @@
 
 # expect oc to be in PATH by default
 OC_TOOL="${OC_TOOL:-oc}"
-nodeSelector="$(${OC_TOOL} get performanceprofile ci -o=jsonpath='{.spec.nodeSelector}'  | awk -F'[/:]' '{print $2}')"
-mcp="$(${OC_TOOL} get mcp -l machineconfiguration.openshift.io/role=$nodeSelector -o name | awk -F "/" '{print $2}')"
 
-# Remove node label
-echo "[INFO]: Unlabeling worker nodes"
-nodes=$(${OC_TOOL} get nodes --selector="node-role.kubernetes.io/${nodeSelector}" -o name)
-for node in $nodes
+profiles=$(${OC_TOOL} get performanceprofile -o name)
+for profileName in $profiles
 do
-    ${OC_TOOL} label $node node-role.kubernetes.io/${nodeSelector}-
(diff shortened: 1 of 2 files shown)


hack/clean-deploy.sh
```diff
@@ -2,15 +2,19 @@
...
-# Remove node label
-echo "[INFO]: Unlabeling worker nodes"
-nodes=$(${OC_TOOL} get nodes --selector="node-role.kubernetes.io/${nodeSelector}" -o name)
-for node in $nodes
+profiles=$(${OC_TOOL} get profiles -o name)
+for profileName in $profiles
 do
-    ${OC_TOOL} label $node node-role.kubernetes.io/${nodeSelector}-
+    nodeSelector="$(${OC_TOOL} get performanceprofile $profileName -o=jsonpath='{.spec.nodeSelector}'  | awk -F'[/:]' '{print $2}')"
+    mcps+=($(${OC_TOOL} get mcp -l machineconfiguration.openshift.io/role=$nodeSelector -o name | awk -F "/" '{print $2}'))
```
slintes, line 10: can we `continue` here in case we have found the `worker` role? I'm a bit afraid that we might unlabel normal worker nodes and delete their MCP

dshchedr: hm.. we are doing that inside loop `for profileName in $profiles`, so we should not touch any node that does not have label equal to `nodeSelector`..

slintes: yes, but you can have a profile targeting the `worker` role, if want to tune just all worker nodes.

dshchedr: ah, right.. but I just moved this part into the loop, so we had this issue previously.. :) 

We still need to delete extra labels somehow, how about adding 
`if [[ $nodeSelector -ne "worker" ]]`

dshchedr: pls, see changes

slintes: > we had this issue previously

previously we didn't delete all profiles but `ci` only ;) It's easier now to mess things up.

> if [[ $nodeSelector -ne "worker" ]]

works for me :+1: 

/lgtm
/approve

dshchedr: > previously we didn't delete all profiles but `ci` only ;) It's easier now to mess things up.
 Yes, but we could delete worker label... :)

dshchedr:
hm, found an issue
/hold

PR #194, opened 2020-04-28
Retry getting MCP

slintes:
Getting the MCP might fail especially when waiting for Updated condition during config when master nodes are rebooted, so retry it a few times.

### DIFF
--- a/functests/utils/client/clients.go
+++ b/functests/utils/client/clients.go
@@ -1,17 +1,25 @@
 package client
 
 import (
-	"github.com/openshift-kni/performance-addon-operators/pkg/apis"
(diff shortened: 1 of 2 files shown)


functests/utils/client/clients.go
```diff
@@ -71,3 +79,30 @@ func NewK8s() *kubernetes.Clientset {
 	}
 	return clientset
 }
+
+func GetWithRetry(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
+	var err error
```
fedepaol, line 84: Just a nit, but why not initing with assignement?

fedepaol: Ah, never mind. You are returning it

PR #196, opened 2020-04-29
Implement release process

slintes:
TL/DR: using drone.io for building and pushing upstream release images to quay.io and creating github releases based on git tags.

Results for pushing a `7.0.12-4` tag (on my repo):
- drone.io logs: https://cloud.drone.io/slintes/performance-addon-operators/33
- images pushed (note that both `7.0.12-4` and `7.0.12` were pushed, ignore all others, just testing a lot):
	https://quay.io/repository/slintes/performance-addon-operator?tab=tags
- github release:
	https://github.com/slintes/performance-addon-operators/releases/tag/7.0.12-4

For more info also see the added `docs/releasing.md` doc.
And of course the the `.drone.yml` config file.

- Cleanup of existing Dockerfiles
- Updated OLM manifest generation:
	- The old method with the temporary CSV in the /deploy/olm-catalog directory, and using the same directory for the registry image build, was leading to having the temporary 0.0.1 CSV in the image also when building versions != 0.0.1. Now that directory is completely recreated each time with operator-sdk, and the final manifests for the registry image are copied to / enriched in /build/_output/olm-catalog. The enriched dev CSV is copied back to /deploy/olm-catalog (for tracking it in git).
	- Additionally the package channel now matches the CSV version, in preparation for having 2 versions in it for testing updates.
- Added drone config:
	- create build tools image (Openshift CI uses the same)
	- build artifacts
	- build and push images
	- validate images
	- create github release

TODO: enable drone.io and configure secrets for the `openshift-kni` org


slintes:
> overall looks very nice. There are only two things that caught my eye on the first pass

Thanks for the review! 

> 
> 1. image validation: is the sqlite db structure (tables etc.) stable and documented? can these queries break? maybe not very important since the format will change in 4.5/4.6 timeframe

Good question. Since it is no public API they might change indeed. But since the registry image will change anyway, and looking at the db is easier than parsing files or querying the grpc server, I think this is fine for now.
 
> 2. the manifest building changes may have impacts d/s. Just at glance it seems OK, but I'll have a deeper look later.

I looked at the ds builds, and I understood that you have your own CSV generation script there. So yes, I hope it's fine :)

fromanirh:
agree.d Let's move forward.
/lgtm

PR #204, opened 2020-04-30
Added CSV and subscription channels

dshchedr:
In order to test operator upgrades we need to provide both CSV version - pointed to image v4.4 and v4.5. 

In this PR I added CSV with previous version
Added subscription channels for switching between versions


deploy/crds-old/4.4.0_performanceprofiles_crd.yaml
```diff
@@ -0,0 +1,162 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
slintes, line 1: Do we need this manifest here, isn't it enough to have it in `deploy/olm-catalog/performance-addon-operator/4.4.0`?

deploy/olm-catalog/performance-addon-operator/4.4.0/performance-addon-operator.v4.4.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,190 @@
...
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.annotations['olm.targetNamespaces']
+                - name: POD_NAME
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.name
+                - name: OPERATOR_NAME
+                  value: performance-operator
+                image: quay.io/openshift-kni/performance-addon-operator:v4.4
```
slintes, line 129: image tag should be `4.4.0`

hack/csv-generate.sh
```diff
@@ -4,18 +4,27 @@ set -e
...
 
 PACKAGE_DIR="deploy/olm-catalog/${PACKAGE_NAME}"
 CSV_DIR="${PACKAGE_DIR}/${CSV_VERSION}"
-CSV_FILE="$CSV_DIR/${PACKAGE_NAME}.v${CSV_VERSION}.clusterserviceversion.yaml"
 
 OUT_DIR="build/_output/olm-catalog"
 OUT_CSV_DIR="${OUT_DIR}/${PACKAGE_NAME}/${CSV_VERSION}"
-OUT_CSV_FILE="${OUT_CSV_DIR}/${PACKAGE_NAME}.v${CSV_VERSION}.clusterserviceversion.yaml"
+
+TMP_CSV_VERSION="999.999.999"
```
slintes, line 26: I do not understand why we need this temporary CSV :thinking:

hack/csv-generate.sh
```diff
@@ -43,32 +51,47 @@ fi
 # clean up all old data first
 clean_package
 
+# do not generate new CRD for old versions
+if [[ "$CSV_VERSION" != "$LATEST" ]]; then
```
slintes, line 55: when we have the 4.4.0 CRD in deploy/olm-catalog/performance-addon-operator/4.4.0 already, maybe do like `if version != prev then generate crds`? (otherwise nothing)

version/version.go
```diff
@@ -2,7 +2,7 @@ package version
 
 var (
 	// Version is the operator version
-	Version = "0.0.1"
+	Version = "4.5.0"
```
slintes, line 5: maybe keep this at `0.0.1`. The build process should set the correct version. If we see 0.0.1 then later, we know that something went wrong with the build.

dshchedr: actually, i don't know :) let's keep it 0.0.1

hack/csv-generate.sh
dshchedr: removed tmp folder

deploy/crds-old/4.4.0_performanceprofiles_crd.yaml
dshchedr: done

deploy/olm-catalog/performance-addon-operator/4.4.0/performance-addon-operator.v4.4.0.clusterserviceversion.yaml
dshchedr: thanks!

deploy/olm-catalog/performance-addon-operator/4.4.0/performance.openshift.io_performanceprofiles_crd.yaml
```diff
@@ -0,0 +1,162 @@
...
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: PerformanceProfileSpec defines the desired state of PerformanceProfile.
+          properties:
+            cpu:
```
slintes, line 34: this is missing `additionalKernelArgs`?

hack/csv-generate.sh
```diff
@@ -43,8 +50,12 @@ fi
 # clean up all old data first
 clean_package
 
+# do not generate new CRD for old versions
+if [[ "$CSV_VERSION" != "$PREV" ]]; then
+  $OPERATOR_SDK generate crds
```
slintes, line 55: What do you think about moving the CSV generation here as well?

deploy/olm-catalog/performance-addon-operator/4.4.0/performance.openshift.io_performanceprofiles_crd.yaml
dshchedr: ah, i got this crd from d/s repo - it does not have KernelArgs for some reason..

will add it

hack/csv-generate.sh
dshchedr: yep, moved

slintes:
when this is merged, you can add 4.6m using 4.4 and 4.5 as old versions :)

PR #205, opened 2020-05-04
Fix CI build image, don't use root

slintes:


slintes:
CI on this is useless, will merge

PR #209, opened 2020-05-04
Set permission on $HOME at the end

slintes:


openshift-ci/Dockerfile.tools
```diff
@@ -22,7 +22,6 @@ RUN yum --setopt=install_weak_deps=False -y install \
     && yum clean all
 
 RUN mkdir -p $HOME && \
-    chmod -R 777 $HOME && \
```
yuvalk, line 25: not a big difference, but the default is 755
kinda weird that we need to create it.
and if we do, maybe just as well, not call it $HOME...

PR #212, opened 2020-05-05
Improved release doc.

slintes:
Added release branches, re-ordered some paragraphs, and smaller fixes


slintes:
e2e does not like this doc change :slightly_frowning_face: 
also, the retry feature from #194 does not seem to work on some errors where it should, will have a look in a follow up

/retest

PR #214, opened 2020-05-06
cleanup: streamline and use csv-generator options

fromanirh:
This patch renames for consistency the `-inject-annotations-from`
option to `-annotations-from` - so all the options now look like
`-*-from`. We fix all the usages in the source tree.

Additionally, let's also use in the generator script the option
to replace the CSV description. This is opt-in as the rest of the
replacement options.


fromanirh:
oh dear, still autoapproved.

PR #216, opened 2020-05-07
Do not fail config suite when RT kernel installation fails

gsr-shanks:
The recent RT issue was failing the whole CI. We need to learn from that and make sure an RT kernel failure only fails RT specific tests so we can identify the culprit right away.


slintes:
we discussed this change and came to the conclusion it increases test time significantly for very low benefit:
- it's not very likely that the reason for this change happens again (RT kernel not available in OS image)
- other features have dependencies on other components as well, so in theory we would need to do something similar for them as well
- we would run tests of other features without RT kernel which might have (missing?) side effects

/close

PR #220, opened 2020-05-08
[release-4.4] align the csv processing tools with the improvements in the 4.5 branch

fromanirh:
In this PR we backport all the improvements made in the 4.5 branch, to make 4.4 maintenance easier and more like 4.5.
No one of these patches actually affects the operator, but rather only the release process and the metadata handling, so the risk is very low.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ build/_output
 build/_test
 _cache
 tools/csv-generator/csv-generator
+tools/csv-replace-imageref/csv-replace-imageref
 # Created by https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
 ### Emacs ###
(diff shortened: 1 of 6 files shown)


fromanirh:
@slintes we don't yet _need_ this PR, but I'm preparing it to make it handy when the time comes.

slintes:
> @slintes we don't yet _need_ this PR now, but I'm preparing it to make it handy when the time comes.

I would actually prefer to merge now, instead of having a PR hanging around for a longer time. I verified that it doesn't touch the existing CSV/CRDs, so feel free to remove the hold when you are ready with the PR.

/lgtm

fromanirh:
merging now works for me. Will remove the hold later today.

PR #221, opened 2020-05-08
 Fix names in clean-deploy.sh script

dshchedr:
clean-deploy.sh does not delete all necessary resources after merging  PR #203


slintes:
good catch, thanks, I always forget this script.

/lgtm
/approve

PR #225, opened 2020-05-11
Use the cnf tests image as container image for tests pod.

fedepaol:
We are replacing the various test images packing everything inside the cnf tests image.
This reduces the number of images we need to ship.


fedepaol:
/hold
requires https://github.com/openshift-kni/cnf-features-deploy/pull/196 to land and the image to get mirrored on quay

fedepaol:
Overriding as config timed out but proper tests passed, meaning the new image is good.

PR #228, opened 2020-05-12
[release-4.5] Fixed registry image

slintes:


slintes:
the config phase failed, but all tests were running fine, so there was a flake...

PR #234, opened 2020-05-13
[release-4.4] Add test id's to tests in 4.4 release branch

gsr-shanks:


functests/performance/performance.go
```diff
@@ -92,13 +92,13 @@ var _ = Describe("performance", func() {
 
 		// Check /usr/local/bin/pre-boot-tuning.sh existence under worker's rootfs
 		const perfRtKernelPrebootTuningScript = "/usr/local/bin/pre-boot-tuning.sh"
-		It(perfRtKernelPrebootTuningScript+" should exist on the nodes", func() {
+		It(perfRtKernelPrebootTuningScript+"[test_id:28523][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] should exist on the nodes", func() {
```
dshchedr, line 95: hm. may be shoould be this?
		It("[test_id:28523][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] "+perfRtKernelPrebootTuningScript+" should exist on the nodes", func() {

dshchedr: hm, IMHO even better to use `fmt.Sprintf()`

functests/performance/performance.go
```diff
@@ -92,13 +92,13 @@ var _ = Describe("performance", func() {
 
 		// Check /usr/local/bin/pre-boot-tuning.sh existence under worker's rootfs
 		const perfRtKernelPrebootTuningScript = "/usr/local/bin/pre-boot-tuning.sh"
```
dshchedr, line 94: don't really understand why we need this const, but.. probably it's out of scope of this PR

PR #237, opened 2020-05-13
Increase the waiting timeout for the stress pod, now using the cnf tests image

fedepaol:
We align the waiting timeout to the other pods.

### DIFF
--- a/functests/1_performance/cpu_management.go
+++ b/functests/1_performance/cpu_management.go
@@ -135,7 +135,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 			err = testclient.Client.Create(context.TODO(), testpod)
 			Expect(err).ToNot(HaveOccurred())
 
-			err = pods.WaitForCondition(testpod, corev1.PodReady, corev1.ConditionTrue, 60*time.Second)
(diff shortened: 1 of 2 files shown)


fedepaol:
@slintes I just noticed the other timeout is 180 secs.
Trying to make this longer.

slintes:
hm, I see the hugepages test failing quite often when waiting for the test pod getting ready [0].
I'm wondering if even 180 seconds isn't enough? Because I don't see anything else which could cause this... thoughts? :thinking: 

[0] https://github.com/openshift-kni/performance-addon-operators/blob/master/functests/1_performance/hugepages.go#L114

fedepaol:
> hm, I see the hugepages test failing quite often when waiting for the test pod getting ready [0].
> I'm wondering if even 180 seconds isn't enough? Because I don't see anything else which could cause this... thoughts? thinking
> 
> [0] https://github.com/openshift-kni/performance-addon-operators/blob/master/functests/1_performance/hugepages.go#L114

With the stress one was even more evident, because it's just two tests and only the first one fails. I assume it's the first time it downloads the image. We can try to push the timeout even further, to get to a stable situation.

slintes:
thanks :+1: 

/lgtm

PR #226, opened 2020-05-12
Add performance operator upgrade test 4.4.0->4.5.0

dshchedr:

PR #241, opened 2020-05-17
Better retry for getting resources in tests

slintes:
Just always retry, it didn't catch some errors for which it should have retried (connection refused e.g.)


slintes:
#226 was merged in the meanwhile with unrelated changes...

/override ci/prow/e2e-gcp

PR #250, opened 2020-05-21
[release-4.4] Use same structure of functests as in release-4.5

gsr-shanks:
functests layout have changed in release-4.5 branch and hence cherry-picking tests to release-4.4 is difficult. This change is to align with the refactored code in release-4.5 branch.


functests/0_config/config.go
```diff
@@ -0,0 +1,132 @@
...
+				},
+				AdditionalKernelArgs: []string{
+					"nmi_watchdog=0",
+					"audit=0",
+					"mce=off",
+					"processor.max_cstate=1",
+					"idle=poll",
+					"intel_idle.max_cstate=0",
+				},
+				NUMA: &performancev1alpha1.NUMA{
```
dshchedr, line 79: We don't have `&performancev1alpha1.NUMA` in 4.4, so need to remove this part

dshchedr:
To be honest, I'm not sure we really need such big changes in 4.4. We have very big difference between 4.4 and 4.5 test structures (it's not only functests folder - same goes for internal image, restructure of cluster-setup folder, etc..) so as I think it's better to just manually import all lacking tests (only 4?) and don't touch this branch anymore :))

Let me close this PR, if someone has other thoughts - we can reopen it.

gsr-shanks:
Yes, I and @dshchedr did some assessment on the number of tests that actually should be in 4.4 and the count is 4. Considering both the efforts and given that 4.4 is GA as dev preview, we will just add the missing tests.

If the situation demands a complete refactoring we will revisit this.

PR #251, opened 2020-05-21
e2e: hugepages: add wait on calculation of allocated hugepages

cynepco3hahue:
Possible the race, when we fetch free hugepages before the kernel updated it.

### DIFF
--- a/functests/1_performance/hugepages.go
+++ b/functests/1_performance/hugepages.go
@@ -123,9 +123,10 @@ var _ = Describe("[performance]Hugepages", func() {
 			availableHugepages := checkHugepagesStatus(availableHugepagesFile, workerRTNode)
 
 			freeHugepagesFile := fmt.Sprintf("/sys/devices/system/node/node0/hugepages/hugepages-%skB/free_hugepages", hpSizeKb)
-			freeHugepages := checkHugepagesStatus(freeHugepagesFile, workerRTNode)
-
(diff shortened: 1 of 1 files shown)


functests/1_performance/hugepages.go
```diff
@@ -123,9 +123,10 @@ var _ = Describe("[performance]Hugepages", func() {
 			availableHugepages := checkHugepagesStatus(availableHugepagesFile, workerRTNode)
 
 			freeHugepagesFile := fmt.Sprintf("/sys/devices/system/node/node0/hugepages/hugepages-%skB/free_hugepages", hpSizeKb)
-			freeHugepages := checkHugepagesStatus(freeHugepagesFile, workerRTNode)
-
-			Expect(availableHugepages - freeHugepages).To(Equal(1))
+			Eventually(func() int {
+				freeHugepages := checkHugepagesStatus(freeHugepagesFile, workerRTNode)
+				return availableHugepages - freeHugepages
+			}, 30*time.Second, time.Second).Should(Equal(1))
```
MarSik, line 129: I am not a big fan of hardcoded timing. Is this really the only way?

dshchedr: Hm, it's not hardcoded timing, it will continue running code as soon as it will get right response

MarSik: Ack, I see it now, it tries every second for up to 30.

PR #254, opened 2020-05-22
Make it possible to override the performance profile used in tests.

fedepaol:
In case of a manifest being there, we use it for tests.
This should cover the case of dpdk not requiring a new profile.

### DIFF
--- a/functests/0_config/config.go
+++ b/functests/0_config/config.go
@@ -3,6 +3,8 @@ package __performance_config
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
+	"os"
 	"time"
 
 	. "github.com/onsi/ginkgo"
@@ -11,6 +13,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog"
 	"k8s.io/utils/pointer"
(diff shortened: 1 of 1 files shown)


functests/0_config/config.go
```diff
@@ -31,64 +37,22 @@ var _ = Describe("[performance][config] Performance configuration", func() {
...
-					"processor.max_cstate=1",
-					"idle=poll",
-					"intel_idle.max_cstate=0",
-				},
-				NUMA: &performancev1alpha1.NUMA{
-					TopologyPolicy: pointer.StringPtr("single-numa-node"),
-				},
-			},
+		performanceProfile := testProfile()
+		performanceManifest := os.Getenv("PERFORMANCE_PROFILE_MANIFEST_OVERRIDE")
```
fromanirh, line 41: Next time I'd use LookupEnv. For this case I think is good enough (or maybe add it as followup)

functests/0_config/config.go
```diff
@@ -130,3 +94,73 @@ var _ = Describe("[performance][config] Performance configuration", func() {
 	})
 
 })
+
+func externalPerformanceProfile(performanceManifest string) (*performancev1alpha1.PerformanceProfile, error) {
+	performanceScheme := runtime.NewScheme()
+	apis.AddToScheme(performanceScheme)
```
fromanirh, line 100: is it OK to this multiple times, right? Any side effects?

fedepaol: Why multiple times? This is happening only once.
In any case the scheme we are adding to is local, used only for the decoder.

functests/0_config/config.go
fedepaol: Ah, did not know about it. Let me change it.

fedepaol: Done

functests/0_config/config.go
fromanirh: Nevermind. I was worried about some code calling this code more than once. But this can't happen as things stand today.

functests/0_config/config.go
```diff
@@ -31,64 +37,22 @@ var _ = Describe("[performance][config] Performance configuration", func() {
...
-					"processor.max_cstate=1",
-					"idle=poll",
-					"intel_idle.max_cstate=0",
-				},
-				NUMA: &performancev1alpha1.NUMA{
-					TopologyPolicy: pointer.StringPtr("single-numa-node"),
-				},
-			},
+		performanceProfile := testProfile()
+		performanceManifest, found := os.LookupEnv("PERFORMANCE_PROFILE_MANIFEST_OVERRIDE")
```
fromanirh, line 41: for a followup: all the env variables we use should be documented _somewhere_ or we will forget very fast

PR #257, opened 2020-05-24
Fix realtime kernel argument in base profile

yanirq:
The wrong kernel argument was set in base profile for cpu isolation (done as an equivalent for tuned realtime profile) 
changed from isolated_cpus to isolcpus

**Note**:  This fix is intended for 4.6 and above (currently fixes an u/s change made in https://github.com/openshift-kni/performance-addon-operators/pull/190)

### DIFF
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -54,6 +54,15 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 
 	Context("Pre boot tuning adjusted by tuned ", func() {
 
+		It("Should set CPU affinity kernel argument", func() {
+			for _, node := range workerRTNodes {
+				cmdline, err := nodes.ExecCommandOnMachineConfigDaemon(&node, []string{"cat", "/proc/cmdline"})
+				Expect(err).ToNot(HaveOccurred())
+				// since systemd.cpu_affinity is calculated on node level using tuned we can check only the key in this context.
+				Expect(cmdline).To(ContainSubstring("systemd.cpu_affinity="))
+			}
+		})
+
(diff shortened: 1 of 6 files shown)


dshchedr:
that's strange, why we didn't see this issue in previous PRs...

functests/1_performance/performance.go
```diff
@@ -54,6 +54,15 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 
 	Context("Pre boot tuning adjusted by tuned ", func() {
 
+		It("Should set CPU affinity kernel argument", func() {
```
gsr-shanks, line 57: [test_id:31198] please

yanirq:
> that's strange, why we didn't see this issue in previous PRs...

see the note in PR description

cynepco3hahue:
Hm do we have a test that really checks the affinity and not the only `cmdline`?

yanirq:
> Hm do we have a test that really checks the affinity and not the only `cmdline`?

We have this: https://github.com/openshift-kni/performance-addon-operators/blob/6b1a7cef91089675243709b596b6d0db42e801eb/functests/1_performance/cpu_management.go#L69  but it might be missing something on the kernel level. need to check how we can enhance it

functests/1_performance/performance.go
yanirq: done

cynepco3hahue:
> We have this:
> 
> https://github.com/openshift-kni/performance-addon-operators/blob/6b1a7cef91089675243709b596b6d0db42e801eb/functests/1_performance/cpu_management.go#L69
> 
> but it might be missing something on the kernel level. need to check how we can enhance it

I just wonder how did it pass, when we passed incorrect argument, or do I miss something?

PR #258, opened 2020-05-26
[release-4.5] Make upgrade operator test more flexible

dshchedr:
This PR allows us to run upgrade test on a cluster with pre-installed operator.

Also slightly updated the logic of the test: now it should verify CSV version and
check that Image Tag and CRD were also modified

### DIFF
--- a/cluster-setup/upgrade-test-cluster/performance/operator_subscription.patch.yaml
+++ b/cluster-setup/upgrade-test-cluster/performance/operator_subscription.patch.yaml
@@ -4,5 +4,4 @@ metadata:
   name: performance-addon-operator
   namespace: openshift-performance-addon
 spec:
-  channel: 4.4.0
-
+  channel: REPLACE_CHANNEL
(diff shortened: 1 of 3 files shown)


cynepco3hahue:
@dshchedr hm why do we need to cherry-pick it?

dshchedr:
> @dshchedr hm why do we need to cherry-pick it?
The main reason - for running test in d/s

cynepco3hahue:
Like if you under branch 4.5 the only upgrade that can happen is from 4.4 to 4.5, no? Or d/s releases look different for u/s?

dshchedr:
in u/s we have subscription channels: 4.4.0 and 4.5.0
in d/s we have 4.4 and 4.5

also, previously upgrade script got empty cluster, installed operator, upgraded it and cleaned cluster.  In d/s we can have pre-installed operator.

dshchedr:
I decided to close this PR, I'm planning to modify upgrade script in master branch first and after that we can cherry pick it here, if needed.

PR #264, opened 2020-06-01
Do not update the feature gate anymore

cynepco3hahue:
The topology manager was promoted to beta under the k8s 1.18,
so we do not need to update the default feature gate resource to
enable LatencySensitive feature set.


cynepco3hahue:
@fromanirh @MarSik Please review

MarSik:
@cynepco3hahue Can you explain what will happen on upgrades? Will the feature gate just stay enabled, but cause no issues?

cynepco3hahue:
yes, it the expectation, so we will have two flows

1. clear installation, we will not touch feature gate resource at all
2. update, we will stop to monitor feature gate object

PR #265, opened 2020-06-02
Improvement for operator upgrade test:

dshchedr:
- it runs all other tests after upgrade
- option to run on a cluster with pre-installed profile
- option to run it on d/s

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ cluster-wait-for-mcp:
     # NOTE: for CI this is done in the config suite of the functests!
     # Use this when deploying manifests manually with CLUSTER=manual
 	@echo "Waiting for MCP to be updated"
-	hack/wait-for-mcp.sh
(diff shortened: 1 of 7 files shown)


dshchedr:
hm, need to update code a little bit

/hold

cynepco3hahue:
/lgtm

@dshchedr Do you have intention to add additional checks under the upgrade test, I know you encounter some but with the change of tuned profile, I prefer to catch such bugs with the CI.

dshchedr:
> @dshchedr Do you have intention to add additional checks under the upgrade test, I know you encounter some but with the change of tuned profile, I prefer to catch such bugs with the CI.

@cynepco3hahue  hm, with this PR we will run all tests  **after** upgrading PAO, so if we have test in _functests/_ folder that covers some specific feature in new version - it should be verified.. 

Do you know some uncovered scenario?

dshchedr:
didn't make changes since previous lgtm, so will add it again :)

/lgtm

PR #267, opened 2020-06-09
e2e: conver bytes to string under the tuned test

cynepco3hahue:


cynepco3hahue:
The fix does not really fix anything.

PR #268, opened 2020-06-09
Remove outdated tuned CR owned by a performance profile

yanirq:
This PR comes to remove outdate tuned CR that do not match the expected name that is generated from their performance profile owner.
This arose from upgrade procedures where tuned generate name differs from the previous version and resulted in outdated tuned CR that were not removed from the cluster.

### DIFF
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -69,6 +69,19 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 			err = testclient.Client.Create(context.TODO(), outdatedTuned)
 			Expect(err).NotTo(HaveOccurred())
 
+			outdatedKey := types.NamespacedName{
+				Name:      outdatedTuned.Name,
+				Namespace: components.NamespaceNodeTuningOperator,
+			}
+			tuned = &tunedv1.Tuned{}
+			Eventually(func() bool {
+				if err := testclient.Client.Get(context.TODO(), outdatedKey, tuned); err != nil {
+					return false
+				}
+				return true
+			}, 120*time.Second, testPollInterval*time.Second).Should(BeTrue(),
+				"outdated tuned does not exist in the cluster")
+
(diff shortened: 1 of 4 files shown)


functests/1_performance/performance.go
```diff
@@ -50,6 +50,49 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 		Expect(err).ToNot(HaveOccurred())
 	})
 
+	Context("Tuned CRs generated from profile", func() {
+		It("[test_id:] Should have the expected name for tuned from the profile owner object", func() {
```
gsr-shanks, line 54: test_id: 31754

functests/1_performance/performance.go
```diff
@@ -50,6 +50,49 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+			tuned := &tunedv1.Tuned{}
+			err = testclient.Client.Get(context.TODO(), key, tuned)
+			Expect(err).ToNot(HaveOccurred(), "cannot find the Cluster Node Tuning Operator object "+components.ProfileNamePerformance)
+			outdatedTuned := tuned.DeepCopy()
+			outdatedTuned.Name = "outdatedName"
+			err = testclient.Client.Create(context.TODO(), outdatedTuned)
+			Expect(err).NotTo(HaveOccurred())
+
+			Eventually(func() bool {
+				for t := range tunedList.Items {
```
cynepco3hahue, line 73: you need to list again under the Eventually, otherwise, you always will have old list

cynepco3hahue: I also afraid of the race here:
- tuned created
- the list still does not have it
- you call to get and it does not return the error, because the operator wasn't fast enough to delete it

cynepco3hahue:
I also miss the unittest and it can be great to add a similar e2e test to the upgrade suit.

PR #271, opened 2020-06-11
[release-4.5] Remove outdated tuned CR owned by a performance profile

yanirq:
This is a cherry-pick of #268

This PR comes to remove outdated tuned CR that do not match the expected name that is generated from their performance profile owner.
This arose from upgrade procedures where tuned generate name differs from the previous version and resulted in outdated tuned CR that were not removed from the cluster.


cynepco3hahue:
@yanirq What was the bugzilla for it?

PR #272, opened 2020-06-17
must-gather: remove unneeded volume mount

fromanirh:
Further testing revealed we don't actually need the host mounts
in the current scenario (just run lspci and lscpu).
We already get /proc and /sys out of the box.

Also, remapping the host /proc inside the container was just wrong
and sometime lead to
```
container create failed: time="2020-06-17T09:06:26Z" level=error msg="container_linux.go:349: starting container process caused \"set process label: open /proc/self/task/1/attr/exec: no such file or directory\""
container_linux.go:349: starting container process caused "set process label: open /proc/self/task/1/attr/exec: no such file or directory"
```

It's just simpler and safer to avoid the mount until there is
a clear and obvious need for them.

### DIFF
--- a/must-gather/collection-scripts/gather_nodes
+++ b/must-gather/collection-scripts/gather_nodes
@@ -54,7 +54,7 @@ do
 
     oc exec $pod -n perf-node-gather -- lspci -nvv 2>/dev/null >> $NODE_PATH/lspci
     oc exec $pod -n perf-node-gather -- lscpu -e 2>/dev/null >> $NODE_PATH/lscpu
-    oc exec $pod -n perf-node-gather -- cat /host/proc/cmdline 2>/dev/null >> $NODE_PATH/proc_cmdline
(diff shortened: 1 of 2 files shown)


fromanirh:
> Further testing revealed we don't actually need the host mounts
> in the current scenario (just run lspci and lscpu).
> We already get /proc and /sys out of the box.
> 
> Also, remapping the host /proc inside the container was just wrong
> and sometime lead to
> 
> ```
> container create failed: time="2020-06-17T09:06:26Z" level=error msg="container_linux.go:349: starting container process caused \"set process label: open /proc/self/task/1/attr/exec: no such file or directory\""
> container_linux.go:349: starting container process caused "set process label: open /proc/self/task/1/attr/exec: no such file or directory"
> ```
> 
> It's just simpler and safer to avoid the mount until there is
> a clear and obvious need for them.
> 
> Signed-off-by: Francesco Romani [fromani@redhat.com](mailto:fromani@redhat.com)

worth pointing out: not sure why this error didn't happen before. It happened on a cluster also running kubevirt, but doesn't seem caused or related to kubevirt.

PR #276, opened 2020-06-22
Deploy the custom runtime via the operator

cynepco3hahue:
Functionality to disable/enable the CPU load balancing will be implemented on the CRI-O level, the code under the CRI-O will disable/enable CPU load balancing only when:

- the pod uses `performance-<profile_name>` runtime class
- the pod has `cpu-load-balancing.crio.io: true` annotation

The performance-addon-operator will be responsible for the creation of the high-performance runtime handler config snippet, it will have the same content as default runtime handler, under relevant nodes, and for creation of the high-performance runtime class under the cluster.

A user will be responsible for specifying the relevant runtime class and annotation under the pod.

### DIFF
--- a/build/assets/configs/99-runtimes.conf
+++ b/build/assets/configs/99-runtimes.conf
@@ -1,0 +1,14 @@
+# We should copy paste the default runtime because this snippet will override the whole runtimes section
+[crio.runtime.runtimes.runc]
+runtime_path = ""
+runtime_type = "oci"
+runtime_root = "/run/runc"
+
+# The CRI-O will check the runtime handler name under the code and will activate high-performance features,
+# like CPU load balancing.
+# We should provide the runtime_path because we need to inform that we want to re-use runc binary and we
+# do not have high-performance binary under the $PATH that will point to it.
+[crio.runtime.runtimes.high-performance]
+runtime_path = "/bin/runc"
+runtime_type = "oci"
+runtime_root = "/run/runc"
--- a/deploy/olm-catalog/performance-addon-operator/4.6.0/performance-addon-operator.v4.6.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/performance-addon-operator/4.6.0/performance-addon-operator.v4.6.0.clusterserviceversion.yaml
@@ -97,6 +97,12 @@ spec:
           - tuneds
           verbs:
           - '*'
+        - apiGroups:
+          - node.k8s.io
+          resources:
+          - runtimeclasses
+          verbs:
+          - '*'
         serviceAccountName: performance-operator
       deployments:
       - name: performance-operator
--- a/functests/1_performance/cpu_management.go
+++ b/functests/1_performance/cpu_management.go
@@ -3,13 +3,16 @@ package __performance
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 
 	. "github.com/onsi/ginkgo"
@@ -23,6 +26,7 @@ import (
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/pods"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/profiles"
 	performancev1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
 )
 
 var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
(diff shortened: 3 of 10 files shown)


cynepco3hahue:
@yanirq @fromanirh @MarSik Please start to review, I still want to add e2e tests under the current PR, and the documentation I will add under the separate PR.

build/assets/scripts/cpu-load-balancing-runtime.sh
```diff
@@ -0,0 +1,51 @@
...
+      if [ -n "$CGROUP_CPUSET" ]; then
+        # parse CPU's under the container cpuset.cpus cgroup
+        # explanation of awk magic:
+        # RS=, this tells awk to use , as the record separator.
+        # if, for example, the input is 0,2,4-6,8, then awk will see four records: 0 and 2 and 4-6 and 8
+        # FS=- this tells awk to use - as the field separator
+        # with FS set this way and if, for example, the input record consists of 2-4, then awk will see 2 as the first field and 4 as the second field.
+        # /-/ for loop will be applied for every record that contains -, and continue after to the next record
+        # if the record does not have -, it will just print the line, 1 is awk's shorthand for print-the-line.
+        for cpu in $(awk '/-/{for (i=$1; i<=$2; i++)printf "%s%s",i,ORS;next} 1' RS=, FS=- < "$CGROUP_CPUSET/cpuset.cpus"); do
```
fromanirh, line 23: great doc, thanks. I just wonder if there is a shell tool that nicely prints the list of CPUs from a cpuset. It seems more and more a relatively frequent task for system tools.

cynepco3hahue: We have libraries for python and golang, but I did not aware of some tool that will be available our of the box on the CoreOS

fromanirh: ack, same here. Let's keep the script and the awk magic, then.

deploy/olm-catalog/performance-addon-operator/4.6.0/performance-addon-operator.v4.6.0.clusterserviceversion.yaml
```diff
@@ -92,6 +92,12 @@ spec:
           - tuneds
           verbs:
           - '*'
+        - apiGroups:
```
fromanirh, line 95: we need to make sure this works fine in the update flow

functests/1_performance/cpu_management.go
```diff
@@ -153,6 +157,121 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+				},
+			}
+
+			// use the CPU load balancing runtime class
+			runtimeClassName := components.GetComponentName(profile.Name, components.ComponentNamePrefix)
+			testpod.Spec.RuntimeClassName = &runtimeClassName
+		})
+
+		AfterEach(func() {
+			// it possible that the VM already was deleted as part of the test, in this case we want to skip teardown
```
fromanirh, line 212: VM -> pod? :)

functests/1_performance/cpu_management.go
```diff
@@ -153,6 +157,121 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+		It("should disable CPU load balancing for CPU's used by the pod", func() {
+			By("Starting the pod")
+			err := testclient.Client.Create(context.TODO(), testpod)
+			Expect(err).ToNot(HaveOccurred())
+
+			err = pods.WaitForCondition(testpod, corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Checking what CPU the pod is using")
+			cmd := []string{"/bin/bash", "-c", "ps -o psr $(pgrep -n test) | tail -1"}
```
fromanirh, line 238: In the future we probably should make the test pod just record the pid of the main process somewhere, so tests like this are as trivial (and more robust and future proof) as to read the pid from a pidfile.

functests/1_performance/cpu_management.go
```diff
@@ -153,6 +157,121 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+			err = pods.WaitForDeletion(testpod, 60*time.Second)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Getting the CPU scheduling flags")
+			flags, err = getCPUSchedulingDomainFlags(cpu)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Verifying that the CPU load balancing was enabled back")
+			Expect(len(flags)).To(Equal(len(defaultFlags)))
+			// the CPU flags should be almost the same except the LSB that should be disabled
```
fromanirh, line 269: uhm, same comment as per line 251? Are both correct?

functests/1_performance/cpu_management.go
cynepco3hahue: hh, looks like I thought about the KubeVirt, when I wrote it.

functests/1_performance/cpu_management.go
cynepco3hahue: This check does not verify correctly the CPU pinning(because we do not have test process, only the `/usr/libexec/crio/conmon` processes), I will check the cpu pinning under the container cgroup.

fromanirh: ok, then I need to re-read more carefully the code :)

functests/1_performance/cpu_management.go
cynepco3hahue: done

functests/1_performance/cpu_management.go
```diff
@@ -153,6 +157,130 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+			}
+
+			err = testclient.Client.Delete(context.TODO(), testpod)
+			Expect(err).ToNot(HaveOccurred())
+
+			err = pods.WaitForDeletion(testpod, 60*time.Second)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should disable CPU load balancing for CPU's used by the pod", func() {
```
gsr-shanks, line 229: [test_id:32646] please

cynepco3hahue: done

yanirq:
@cynepco3hahue  please make sure that if some tests are going to work against the generate profile but might not work against a already deployed profile (discovery mode) the you will need to skip it or find a profile that will pass the test (see https://github.com/openshift-kni/performance-addon-operators/pull/290)

build/assets/configs/99-runtimes.conf
```diff
@@ -0,0 +1,10 @@
+# We should copy paste the default runtime because this snippet will override the whole runtimes section
+[crio.runtime.runtimes.runc]
+runtime_path = ""
+runtime_type = "oci"
+runtime_root = "/run/runc"
+
+[crio.runtime.runtimes.high-performance]
```
fromanirh, line 7: do we need to update the commit message?
Could you please document (or link the documentation) about how the crio configuration works now? before using the wrapper was pretty easy to get, but now it feels like something's missing

deploy/olm-catalog/performance-addon-operator/4.6.0/performance-addon-operator.v4.6.0.clusterserviceversion.yaml
fromanirh: still relevant - if covered by existing tests: fine, just state so :)

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -115,24 +116,24 @@ func getIgnitionConfig(assetsDir string, profile *performancev1alpha1.Performanc
 		},
 	}
 
+	// add script files under the node /usr/local/bin directory
+	mode := 0700
 	for _, script := range []string{hugepagesAllocation} {
-		content, err := ioutil.ReadFile(fmt.Sprintf("%s/scripts/%s.sh", assetsDir, script))
-		if err != nil {
+		src := fmt.Sprintf("%s/scripts/%s.sh", assetsDir, script)
```
fromanirh, line 122: this should ne filepath.Join - but better fix in another PR

build/assets/configs/99-runtimes.conf
fromanirh: Questions to clarify why is not very clear (!!!) - if the functionality is baked in crio, and if we just need to provide a new separate runtime class (if so this deserve a comment), why the two runtime classes aren't 1:1 clones?
IOW, why for the default runtime class runtime_path = ""?

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -115,24 +116,24 @@ func getIgnitionConfig(assetsDir string, profile *performancev1alpha1.Performanc
...
-		})
+	}
+
+	// add crio config snippet under the node /etc/crio/configs/ directory
+	crioConfdRuntimesMode := 0644
+	if err := addFile(
+		ignitionConfig,
+		fmt.Sprintf("%s/configs/%s.conf", assetsDir, crioRuntimesConfig),
+		fmt.Sprintf("%s/%s.conf", crioConfd, crioRuntimesConfig),
+		&crioConfdRuntimesMode,
```
fromanirh, line 134: ditto

fromanirh:
Another option: squash the documentation PR in this one, so the documentation need are fulfilled by the doc PR itself.

build/assets/configs/99-runtimes.conf
MarSik: Btw, is the name fixed? Low latency is specifically NOT high performance. Quite the opposite.

cynepco3hahue: done

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
cynepco3hahue: done

cynepco3hahue:
@fromanirh Added comments and updated the PR and commit messages, give me know if I missed something.

build/assets/configs/99-runtimes.conf
cynepco3hahue: @MarSik Yes, it hard-coded under the CRI-O, I was pretty under the expression that providing the low latency will give you a better performance at least for the process that will run under the container configured with it. Like we named our operator as performance-addon-operator.

cynepco3hahue:
@fromanirh The upgrade test is covered by the upgrade lane because under the lane we run upgrade procedure from 4.5 to 4.6 and after we run all e2e tests(including the one that I added), so if the operator will fail to create the runtime class after the upgrade my test will fail.

build/assets/configs/99-runtimes.conf
fromanirh: ok, good enough for me now

PR #277, opened 2020-06-22
fix for test 28026 (rcu_sched|rcu_preempt)

dshchedr:
There is a different kernel scheduler name in RT and non-RT kernels, this small fix verifies both of them.

### DIFF
--- a/functests/1_performance/cpu_management.go
+++ b/functests/1_performance/cpu_management.go
@@ -85,7 +85,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 			Expect(conf).To(MatchRegexp(fmt.Sprintf(`"reservedSystemCPUs": ?"%s"`, reservedCPU)))
 
 			By("checking CPU affinity mask for kernel scheduler")
-			cmd = []string{"/bin/bash", "-c", "taskset -pc $(pgrep rcu_sched)"}
+			cmd = []string{"/bin/bash", "-c", "taskset -pc $(pgrep 'rcu_sched|rcu_preempt')"}
 			sched, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
 			Expect(err).ToNot(HaveOccurred(), "failed to execute taskset")
 			mask := strings.SplitAfter(sched, " ")


functests/1_performance/cpu_management.go
```diff
@@ -85,7 +85,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 			Expect(conf).To(MatchRegexp(fmt.Sprintf(`"reservedSystemCPUs": ?"%s"`, reservedCPU)))
 
 			By("checking CPU affinity mask for kernel scheduler")
-			cmd = []string{"/bin/bash", "-c", "taskset -pc $(pgrep rcu_sched)"}
+			cmd = []string{"/bin/bash", "-c", "taskset -pc $(pgrep 'rcu_sched|rcu_preempt')"}
```
fromanirh, line 88: ok, so we want the CPU affinity mask of a kernel thread. Is any kernel thread ok for this check? Can we pick a kernel thread which is unlikely to change in the future? Or are these good enough?

cynepco3hahue: IIUC it does not all kernel threads but only `rcu_*` threads, I unsure regarding what thread should we check, but `rcu_preempt` thread only available once you compile your kernel with `CONFIG_PREEMPT=y`, from https://lwn.net/Articles/519432/
```
There is one thread for each flavor of RCU, which means three on CONFIG_PREEMPT=y kernels (rcu_bh, rcu_preempt, and rcu_sched) and two on CONFIG_PREEMPT=n kernels (rcu_bh and rcu_sched, with rcu_preempt mapping onto rcu_sched).
In theory, a single thread could handle all the RCU flavors, but when I tried this the code got quite ugly.
``` 
So I assume to check `rcu_sched|rcu_preempt` should be good enough because we cover two cases with `CONFIG_PREEMPT=y` and with `CONFIG_PREEMPT=n`.

I still unsure why `rcu_sched` disappeared under the RT kernel :(

cynepco3hahue: @fromanirh I want to merge it to unblock the CI, do you have some additional objections?

fromanirh: @cynepco3hahue go ahead and merge, we can improve later.

fromanirh: BTW, the thing that bothers me here is: why are we checking the RCU threads to verify the reserved mask? I mean, why RCU and not, say, the softirq threads? or the swap threads? what makes rcu a good fit for this case? I think this deserve a few lines of docs, or a link to the aforementioned docs in case I missed it :)

PR #285, opened 2020-06-26
Add a clean.All method that can be exported and consumed externally.

fedepaol:
By adding this method, we allow external suites consuming these tests to be able to clean all the leftovers all together.

### DIFF
--- a/functests/utils/clean/clean.go
+++ b/functests/utils/clean/clean.go
@@ -1,0 +1,61 @@
+package clean
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/profile"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils"
+	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/mcps"
+	performancev1alpha1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1alpha1"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
+	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog"
+)
+
+var cleanPerformance bool
+
+func init() {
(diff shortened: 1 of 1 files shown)


functests/utils/clean/clean.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
+	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+var dontWaitMcp bool
+
+func init() {
```
cynepco3hahue, line 25: Can you rename it to WAIT_FOR_MCP? Or do you already use somewhere the `DONT_WAIT_MCP_ON_DEL` ENV variable?

cynepco3hahue: I think it just unusual to name ENV variable with a negative connotation, it also will be shorter. What the problem with the true value? You can also use `WAIT_FOR_MCP=${WAIT_FOR_MCP:-true}`

fedepaol: What I would like to do is that when you run the tests without specifying the value, it waits.
When I think to a boolean variable, I expect the default value to be false (or, to put things in another way, to have everything to behave as if the value is false if the variable is not set). But I can live with that, we'll document that you need to set `WAIT_FOR_MCP=false` to disable it.

cynepco3hahue: I just curious, what the use case to run tests without waiting for the MCP update?

fedepaol: If you are running in CI for example, you don't care about clean to finish.
But let me re-think about it. Probably it makes more sense to have a way not to delete the profile at all. The thing is, applying the profile is the most time consuming thing in cnf-tests, so it makes sense to optimize with that regard.

cynepco3hahue: @fedepaol ok, give me know if you want to merge as it is now and optimize it under a different PR.

fedepaol: No, let's change this with a toggle that disables the cleanup completely.
If we wait, is because we care about restoring the cluster in the state it was.
If we don't wait, it's because we either want the iteration to be faster (and in that case we may want to keep the profile) or the cluster is a thrown-away (and in that case we don't care about cleaning it).

fedepaol:
> I have one small comment but except it, lgtm.
> Another question where do you want to use this method?

The idea is to have each dependency of cnf-tests to expose it's own clean.All() method and calling all of them in a more organized way. Right now the cleaning is a bit messy.

fedepaol:
Ok, I think this approach is better for both purposes of not waiting the test to finish (if we don't care about the cluster) and making the following iterations faster.
@cynepco3hahue please take another look

PR #286, opened 2020-06-30
Remove hugepages multiple sizes validation

cynepco3hahue:
Multiple hugepages sizes support feature was promoted to beta under
the Kubernetes 1.19, it gives us possibility to support multiple
hugepages sizes under the PAO, without need to enable additional
feature gate.

### DIFF
--- a/functests/1_performance/hugepages.go
+++ b/functests/1_performance/hugepages.go
@@ -47,6 +47,8 @@ var _ = Describe("[performance]Hugepages", func() {
 		Expect(profile.Spec.HugePages).ToNot(BeNil())
 	})
 
+	// We have multiple hugepages e2e tests under the upstream, so the only thing that we should check, if the PAO configure
+	// correctly number of hugepages that will be available on the node
 	Context("[rfe_id:27369]when NUMA node specified", func() {
 		It("[test_id:27752][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] should be allocated on the specifed NUMA node ", func() {
 			for _, page := range profile.Spec.HugePages.Pages {
@@ -57,14 +59,10 @@ var _ = Describe("[performance]Hugepages", func() {
 				hugepagesSize, err := machineconfig.GetHugepagesSizeKilobytes(page.Size)
 				Expect(err).ToNot(HaveOccurred())
 
-				availableHugepagesFile := fmt.Sprintf("/sys/devices/system/node/node%d/hugepages/hugepages-%skB/nr_hugepages", *page.Node, hugepagesSize)
+				availableHugepagesFile := fmt.Sprintf("/sys/kernel/mm/hugepages/hugepages-%skB/nr_hugepages", hugepagesSize)
 				nrHugepages := checkHugepagesStatus(availableHugepagesFile, workerRTNode)
 
-				if discovery.Enabled() && nrHugepages != 0 {
-					Skip("Skipping test since other guests might reside in the cluster affecting results")
-				}
-
-				freeHugepagesFile := fmt.Sprintf("/sys/devices/system/node/node%d/hugepages/hugepages-%skB/free_hugepages", *page.Node, hugepagesSize)
+				freeHugepagesFile := fmt.Sprintf("/sys/kernel/mm/hugepages/hugepages-%skB/nr_hugepages", hugepagesSize)
 				freeHugepages := checkHugepagesStatus(freeHugepagesFile, workerRTNode)
 
 				Expect(int32(nrHugepages)).To(Equal(page.Count), "The number of available hugepages should be equal to the number in performance profile")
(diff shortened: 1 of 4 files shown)


fromanirh:
Looks reasannable. For the lgtm, please add a couple lines in the commit message documenting which is the minimum version (I guess kube 1.19) which supports the multiple page sizes.

pkg/controller/performanceprofile/components/profile/profile.go
```diff
@@ -3,7 +3,7 @@ package profile
 import (
 	"fmt"
 
-	"github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
+	v1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
```
marcel-apf, line 6: It seems the preferred convention is now performancev1, just saying

pkg/controller/performanceprofile/components/profile/profile_test.go
```diff
@@ -1,7 +1,9 @@
 package profile
 
 import (
-	"github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
+	"fmt"
+
+	v1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
```
marcel-apf, line 6: No need for rename if you use the same v1 prefix

pkg/controller/performanceprofile/components/profile/profile_test.go
```diff
@@ -75,15 +77,15 @@ var _ = Describe("PerformanceProfile", func() {
 			Expect(err.Error()).To(ContainSubstring("hugepages default size should be equal"))
 		})
 
-		It("should reject hugepages allocation with different sizes", func() {
+		It("should reject hugepages allocation with unexpected page size", func() {
 			profile.Spec.HugePages.Pages = append(profile.Spec.HugePages.Pages, v1.HugePage{
 				Count: 128,
 				Node:  pointer.Int32Ptr(0),
-				Size:  v1.HugePageSize("2M"),
+				Size:  v1.HugePageSize("14M"),
```
marcel-apf, line 84: Can you explain this change?

functests/1_performance/hugepages.go
```diff
@@ -57,14 +59,39 @@ var _ = Describe("[performance]Hugepages", func() {
...
+					availableHugepagesFile = fmt.Sprintf("/sys/devices/system/node/node%d/hugepages/hugepages-%skB/nr_hugepages", *page.Node, hugepagesSize)
+				}
 				nrHugepages := checkHugepagesStatus(availableHugepagesFile, workerRTNode)
 
 				if discovery.Enabled() && nrHugepages != 0 {
 					Skip("Skipping test since other guests might reside in the cluster affecting results")
 				}
 
-				freeHugepagesFile := fmt.Sprintf("/sys/devices/system/node/node%d/hugepages/hugepages-%skB/free_hugepages", *page.Node, hugepagesSize)
+				freeHugepagesFile := fmt.Sprintf("/sys/kernel/mm/hugepages/hugepages-%skB/free_hugepages", hugepagesSize)
```
marcel-apf, line 90: Not for now, but it seems MCO is in charge of setting the huge pages, maybe we should only check the MCO profile.

pkg/controller/performanceprofile/components/profile/profile_test.go
cynepco3hahue: I changed the code to drop the validation once we have unsupported hugepages size and the same did for the unittest.

pkg/controller/performanceprofile/components/profile/profile.go
cynepco3hahue: in general if it the API that defined under the same repo as a controller, I think you do not need the `performance` prefix, at least we had such convention under the KubeVirt
If it the API from the different repo you should add the prefix

functests/1_performance/hugepages.go
cynepco3hahue: I think we can leave it for the discussions that you sent via email ;)

cynepco3hahue:
@fromanirh done

pkg/controller/performanceprofile/components/profile/profile.go
fromanirh: my take: the biggest concern is ambiguity. We do renamed imports (can't remember the go-specific name atm) when we need to do something like

```
import (
    "github.com/foo/pkg/great/api/v1"
    "github.com/bar/pkg/another/great/api/v1"
)
```
and we need to disambiguate the import. In this case there is no ambiguity, so from my perspective code is fine.
We can drop the initial "v1" - we are aliasing the old package with its own name, but I don't mind.

marcel-apf: ok

pkg/controller/performanceprofile/components/profile/profile_test.go
marcel-apf: You updated 2M to 14M somwhere?

functests/1_performance/hugepages.go
marcel-apf: definitely

pkg/controller/performanceprofile/components/profile/profile_test.go
fromanirh: Isn't "14M" a random illegal value which should trigger the validation failure (which the tests expects and want to make sure it happens)?

PR #288, opened 2020-07-01
add support for isolate_managed_irq

marcel-apf:
Tuned added support for the managed_irq flag of isolcpus kernel parameter.
Set the flag in order to move kernel managed IRQs out of  isolated cores in case "StaticIsolation" is enabled.
Update the tests to include the new flag.

### DIFF
--- a/build/assets/tuned/openshift-node-performance
+++ b/build/assets/tuned/openshift-node-performance
@@ -85,9 +85,9 @@ initrd_add_dir=
 # overrides cpu-partitioning cmdline
 cmdline_cpu_part=+nohz=on rcu_nocbs=${isolated_cores} tuned.non_isolcpus=${not_isolated_cpumask} intel_pstate=disable nosoftlockup
 {{if .StaticIsolation}}
-cmdline_realtime=+tsc=nowatchdog intel_iommu=on iommu=pt isolcpus=${isolated_cores} systemd.cpu_affinity=${not_isolated_cores_expanded}
+cmdline_realtime=+tsc=nowatchdog intel_iommu=on iommu=pt isolcpus=domain,managed_irq,${isolated_cores} systemd.cpu_affinity=${not_isolated_cores_expanded}
 {{else}}
-cmdline_realtime=+tsc=nowatchdog intel_iommu=on iommu=pt systemd.cpu_affinity=${not_isolated_cores_expanded}
+cmdline_realtime=+tsc=nowatchdog intel_iommu=on iommu=pt isolcpus=managed_irq,${isolated_cores} systemd.cpu_affinity=${not_isolated_cores_expanded}
 {{end}}
 cmdline_hugepages=+{{if .DefaultHugepagesSize}} default_hugepagesz={{.DefaultHugepagesSize}} {{end}} {{if .Hugepages}} {{.Hugepages}} {{end}}
 cmdline_additionalArg=+{{if .AdditionalArgs}} {{.AdditionalArgs}} {{end}}
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -91,6 +91,18 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 			}
 		})
 
+		It("Should set CPU isolcpu's kernel argument managed_irq flag", func() {
+			for _, node := range workerRTNodes {
+				cmdline, err := nodes.ExecCommandOnMachineConfigDaemon(&node, []string{"cat", "/proc/cmdline"})
+				Expect(err).ToNot(HaveOccurred())
+				if profile.Spec.CPU.BalanceIsolated != nil && *profile.Spec.CPU.BalanceIsolated == false {
+					Expect(string(cmdline)).To(ContainSubstring("isolcpus=domain,managed_irq,"))
+				} else {
+					Expect(string(cmdline)).To(ContainSubstring("isolcpus=managed_irq,"))
+				}
+			}
+		})
+
 		It("[test_id:27081][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Should set workqueue CPU mask", func() {
 			for _, node := range workerRTNodes {
 				By("Getting tuned.non_isolcpus kernel argument")
--- a/functests/2_performance_update/updating_profile.go
+++ b/functests/2_performance_update/updating_profile.go
@@ -117,7 +117,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			table.Entry("[test_id:28024] verify that hugepages size and count updated", chkCmdLine, []string{"default_hugepagesz=2M", "hugepagesz=2M", "hugepages=5"}, true, false),
 			table.Entry("[test_id:28070] verify that hugepages updated (NUMA node unspecified)", chkCmdLine, []string{"hugepagesz=2M"}, true, false),
 			table.Entry("[test_id:28025] verify that cpu affinity mask was updated", chkCmdLine, []string{"tuned.non_isolcpus=00000009"}, true, false),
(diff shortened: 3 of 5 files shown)


build/assets/tuned/openshift-node-performance
```diff
@@ -85,7 +85,7 @@ initrd_add_dir=
 # overrides cpu-partitioning cmdline
 cmdline_cpu_part=+nohz=on rcu_nocbs=${isolated_cores} tuned.non_isolcpus=${not_isolated_cpumask} intel_pstate=disable nosoftlockup
 {{if .StaticIsolation}}
-cmdline_realtime=+tsc=nowatchdog intel_iommu=on iommu=pt isolcpus=${isolated_cores} systemd.cpu_affinity=${not_isolated_cores_expanded}
+cmdline_realtime=+tsc=nowatchdog intel_iommu=on iommu=pt isolcpus=managed_irq,${isolated_cores} systemd.cpu_affinity=${not_isolated_cores_expanded}
```
MarSik, line 88: I think I need more context about what isolcpus=managed_irq does. Because old isolcpus disables cpu process balancing for those cores (isolated_cores). However irqs have impact even in cases when cpu balancing is still enabled and we need to redirect them to not_isolated_cores in all cases (even if static isolation is not wanted).

So it is very likely the managed_irq change is a bit more complicated than this.

Also make sure the option is only used in kernel versions that support that argument. Because it is rather new. You will probably need a conditional section for that (see https://github.com/redhat-performance/tuned/commit/8d9cd00387426ed0)

marcel-apf: I think we can use https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/commit/?id=11ea68f553e244851d15793a7fa33a97c46d8271 as context.

The way I understand it (I might be wrong) managed_irq flag works only in combination with a X-Y cpu range of isolcpus and aims to prevent the case where interrupts are routed to the X-Y range. Meaning, no range no case: ```This isolation is best effort and only effective if the automatically
assigned interrupt mask of a device queue contains isolated and
housekeeping CPUs. ```

Another interesting point is the solution is not bullet free:  ```If a queue's affinity mask contains only isolated CPUs then this parameter
has no effect on the interrupt routing decision...```

However I did find an issue, here is housekeeping_isolcpus_setup function:
 `while (isalpha(*str)) {
                if (!strncmp(str, "nohz,", 5)) {
                        str += 5;
                        flags |= HK_FLAG_TICK;
                        continue;
                }

                if (!strncmp(str, "domain,", 7)) {
                        str += 7;
                        flags |= HK_FLAG_DOMAIN;
                        continue;
                }

                if (!strncmp(str, "managed_irq,", 12)) {
                        str += 12;
                        flags |= HK_FLAG_MANAGED_IRQ;
                        continue;
                }

                //[...]
        }

        /* Default behaviour for isolcpus without flags */
        if (!flags)
                flags |= HK_FLAG_DOMAIN;

So if we add managed_irq flag we remove the main isolation logic, so we need to add ```isolcpus=managed_irq,domain,X-Y```

marcel-apf: Regarding kernel versions, I am all for testing everything and failing as early as possible, but my state of mind was the whole point of the way the system is made is to not have to check this stuff, meaning I hunted down the RHCOS kernel used for 4.5 and 4.6 *before* I started to work on PAO and the code was in.
So my questions are:
1. How can we get in a situation we have an older kernel (do we support at all?) ? 
2. Does it worth the check even if would happen for a not supported kernel and the worth case scenario is a dmesg info line? (not even a warning)
`  pr_info("isolcpus: Skipped unknown flag %.*s\n", len, par);`

If you still think is necessary, I'll start working on it.

MarSik:
And in general I agree with @fromanirh Even though I admit we are not perfect either, commit messages should explain the WHY and HOW. The WHAT is the patch itself.

marcel-apf:
@fromanirh thanks for the comments.
Regarding the commit message, I'll do my best to improve it.
In order to answer about the impact on the upgrade flow, I need to understand it, how can I know how it works?
Where can I see and how can I run the 2e2 tests?

build/assets/tuned/openshift-node-performance
fromanirh: great that you mentioned the `domain` part, I was reading about that and I was wondering (and about to ask) why was it left out

fromanirh:
> @fromanirh thanks for the comments.
> Regarding the commit message, I'll do my best to improve it.

No need to be overly verbose. Just please document the rationale of this change: why you are doing this change, why is needed, what's the benefit.
Of course comments here on github provide more context and the discussion/review here is beneficial. It's fine to refer them here. But the commit message (and the git history in general) should be self-sufficient.

> In order to answer about the impact on the upgrade flow, I need to understand it, how can I know how it works?

It's a long discussion, we'll have to bring it offline :) Also, see below about the upgrade jobs (ci/prow/e2e-gcp-operator-upgrade in our case)

> Where can I see and how can I run the e2 tests?

You need a OCP cluster for that, there is a Makefile target. And the CI does that for you (ci/prow/ci + ci/prow/e2e-gcp)

marcel-apf:
@fromanirh @MarSik I pushed again with a modified commit message and using the managed_irq flag in conjunction with the domain flag

build/assets/tuned/openshift-node-performance
marcel-apf: I added the  'domain' flag to the PR now.

cynepco3hahue:
Ok we have one failing test, I added some fix for the error message that will make it easier to understand what happened, probably new parameter messes something with kernel arguments, probably we should change the check https://github.com/openshift-kni/performance-addon-operators/blob/7031048cfa987d4de2215a855ecb53597389d7ae/functests/1_performance/performance.go#L128 to work with regexp.

build/assets/tuned/openshift-node-performance
MarSik: Ack, it makes much more sense now. But that brings me to another question. Should we use isolcpus=managed_irq,X (without domain) for the non-static case?

marcel-apf:
> /retest

Thanks!

build/assets/tuned/openshift-node-performance
marcel-apf: Well, it seems it can't hurt, in non-static case we will have systemd.cpu_affinity=${not_isolated_cores_expanded}  which will kind of avoid the reserved cpus while  isolcpus=managed_irq,X will avoid triggering interrupts on them.

On the other hand in the static case is clear the user wants the reserved CPU completely isolated and it makes sense to not have IRQs triggered on them at the expense of interrupts starvation (e.g. net packet would be lost since we don't have enough CPUs to process the IRQs), however in the non-static case it possible the starvation may be an issue (theoretically).

After a session with Yanir we decided to consult with Andrew Theurer about this.

marcel-apf:
@fromanirh regarding e2e tests the patch addresses changes to them (if will remain in this form) by verifying the command line parameter values for static isolation - this being the only "functional" change other than the kernel working differently.

fromanirh:
> @fromanirh regarding e2e tests the patch addresses changes to them (if will remain in this form) by verifying the command line parameter values for static isolation - this being the only "functional" change other than the kernel working differently.

OK, changes to https://github.com/openshift-kni/performance-addon-operators/pull/288/commits/d6e3cc459c3e9899941c8ee4268ded65765dd372#diff-382a12f07313e5f630cc3bb0e1ca129a seems good enough. Thanks!

build/assets/tuned/openshift-node-performance
marcel-apf: Here is Andrew's answer from a private chat:

> whatever CPUs are in “isolated:” in the profile should be in  the managed_irq list
> basically any cpu that is in the list for rcu_nocbs should also be used for managed_irq

I'll get to it

marcel-apf: Done, repushed

PR #290, opened 2020-07-02
Make the performance tests work in discovery mode

yanirq:
The performance tests should be able to work in "discovery mode", meaning that they rely on the feature they want to test to be configured on the cluster use use the default test profile the performance tests create. 
Discovery mode is be enabled the environment parameter DISCOVERY_MODE=1/ t / T / TRUE / true /True

### DIFF
--- a/functests/0_config/config.go
+++ b/functests/0_config/config.go
@@ -42,23 +42,14 @@ var _ = Describe("[performance][config] Performance configuration", func() {
 		performanceProfile := testProfile()
 		profileAlreadyExists := false
 
-		if discovery.Enabled() {
+		performanceManifest, foundOverride := os.LookupEnv("PERFORMANCE_PROFILE_MANIFEST_OVERRIDE")
+		if foundOverride {
 			var err error
-			performanceProfile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
-			Expect(err).ToNot(HaveOccurred(), "Failed finding a performance profile in discovery mode")
-			profileAlreadyExists = true
-			klog.Info("Discovery mode: consuming a deployed performance profile from the cluster")
-		} else {
-			performanceManifest, found := os.LookupEnv("PERFORMANCE_PROFILE_MANIFEST_OVERRIDE")
-			if found {
-				var err error
-				performanceProfile, err = externalPerformanceProfile(performanceManifest)
-				Expect(err).ToNot(HaveOccurred(), "Failed overriding performance profile", performanceManifest)
-				klog.Warning("Consuming performance profile from ", performanceManifest)
-			}
+			performanceProfile, err = externalPerformanceProfile(performanceManifest)
+			Expect(err).ToNot(HaveOccurred(), "Failed overriding performance profile", performanceManifest)
+			klog.Warning("Consuming performance profile from ", performanceManifest)
 		}
-
-		if !profileAlreadyExists {
+		if !discovery.Enabled() || foundOverride {
 			By("Creating the PerformanceProfile")
 			// this might fail while the operator is still being deployed and the CRD does not exist yet
 			Eventually(func() error {
@@ -70,6 +61,12 @@ var _ = Describe("[performance][config] Performance configuration", func() {
 				}
 				return err
 			}, 15*time.Minute, 15*time.Second).ShouldNot(HaveOccurred(), "Failed creating the performance profile")
+		} else if !foundOverride {
+			var err error
+			performanceProfile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+			Expect(err).ToNot(HaveOccurred(), "Failed finding a performance profile in discovery mode")
+			klog.Info("Discovery mode: consuming a deployed performance profile from the cluster")
+			profileAlreadyExists = true
 		}
 
 		By("Getting MCP for profile")
--- a/functests/0_config/config.go
+++ b/functests/0_config/config.go
@@ -25,7 +25,9 @@ import (
 	"github.com/openshift-kni/performance-addon-operators/functests/utils"
 	testutils "github.com/openshift-kni/performance-addon-operators/functests/utils"
 	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/discovery"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/mcps"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/profiles"
 	"github.com/openshift-kni/performance-addon-operators/pkg/apis"
 	performancev1alpha1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1alpha1"
 
@@ -38,26 +40,41 @@ var _ = Describe("[performance][config] Performance configuration", func() {
 	It("Should successfully deploy the performance profile", func() {
 
 		performanceProfile := testProfile()
-		performanceManifest, found := os.LookupEnv("PERFORMANCE_PROFILE_MANIFEST_OVERRIDE")
-		if found {
+		profileAlreadyExists := false
+
+		if discovery.Enabled() {
 			var err error
-			performanceProfile, err = externalPerformanceProfile(performanceManifest)
-			Expect(err).ToNot(HaveOccurred(), "Failed overriding performance profile", performanceManifest)
-			klog.Warning("Consuming performance profile from ", performanceManifest)
+			performanceProfile, err = profiles.GetByNodeLabels(
+				map[string]string{
+					fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerCNF): "",
+				},
+			)
+			Expect(err).ToNot(HaveOccurred(), "Failed finding a performance profile in discovery mode")
+			profileAlreadyExists = true
+			klog.Info("Discovery mode: consuming a deployed performance profile from the cluster")
+		} else {
+			performanceManifest, found := os.LookupEnv("PERFORMANCE_PROFILE_MANIFEST_OVERRIDE")
+			if found {
+				var err error
+				performanceProfile, err = externalPerformanceProfile(performanceManifest)
+				Expect(err).ToNot(HaveOccurred(), "Failed overriding performance profile", performanceManifest)
+				klog.Warning("Consuming performance profile from ", performanceManifest)
+			}
 		}
 
-		By("Creating the PerformanceProfile")
-		// this might fail while the operator is still being deployed and the CRD does not exist yet
-		profileAlreadyExists := false
-		Eventually(func() error {
-			err := testclient.Client.Create(context.TODO(), performanceProfile)
-			if errors.IsAlreadyExists(err) {
-				klog.Warning(fmt.Sprintf("A PerformanceProfile with name %s already exists! If created externally, tests might have unexpected behaviour", performanceProfile.Name))
-				profileAlreadyExists = true
-				return nil
-			}
-			return err
-		}, 15*time.Minute, 15*time.Second).ShouldNot(HaveOccurred(), "Failed creating the performance profile")
+		if !profileAlreadyExists {
+			By("Creating the PerformanceProfile")
+			// this might fail while the operator is still being deployed and the CRD does not exist yet
+			Eventually(func() error {
+				err := testclient.Client.Create(context.TODO(), performanceProfile)
+				if errors.IsAlreadyExists(err) {
+					klog.Warning(fmt.Sprintf("A PerformanceProfile with name %s already exists! If created externally, tests might have unexpected behaviour", performanceProfile.Name))
+					profileAlreadyExists = true
+					return nil
+				}
+				return err
+			}, 15*time.Minute, 15*time.Second).ShouldNot(HaveOccurred(), "Failed creating the performance profile")
+		}
 
 		By("Getting MCP for profile")
 		mcpLabel := profile.GetMachineConfigLabel(performanceProfile)
--- a/functests/0_config/config.go
+++ b/functests/0_config/config.go
@@ -44,11 +44,7 @@ var _ = Describe("[performance][config] Performance configuration", func() {
 
 		if discovery.Enabled() {
 			var err error
-			performanceProfile, err = profiles.GetByNodeLabels(
-				map[string]string{
-					fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerCNF): "",
-				},
-			)
+			performanceProfile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
 			Expect(err).ToNot(HaveOccurred(), "Failed finding a performance profile in discovery mode")
 			profileAlreadyExists = true
 			klog.Info("Discovery mode: consuming a deployed performance profile from the cluster")
@@ -137,9 +133,6 @@ func testProfile() *performancev1alpha1.PerformanceProfile {
 	isolated := performancev1alpha1.CPUSet("1-3")
 	hugePagesSize := performancev1alpha1.HugePageSize("1G")
 
-	cnfRoleLabel := fmt.Sprintf("%s/%s", testutils.LabelRole, utils.RoleWorkerCNF)
-	nodeSelector := map[string]string{cnfRoleLabel: ""}
-
 	return &performancev1alpha1.PerformanceProfile{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "PerformanceProfile",
@@ -163,7 +156,7 @@ func testProfile() *performancev1alpha1.PerformanceProfile {
 					},
 				},
 			},
-			NodeSelector: nodeSelector,
+			NodeSelector: testutils.NodeSelectorLabels,
 			RealTimeKernel: &performancev1alpha1.RealTimeKernel{
 				Enabled: pointer.BoolPtr(true),
 			},
--- a/functests/1_performance/cpu_management.go
+++ b/functests/1_performance/cpu_management.go
@@ -33,18 +33,14 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 	var reservedCPUSet cpuset.CPUSet
 
 	BeforeEach(func() {
-		workerRTNodes, err := nodes.GetByRole(testutils.RoleWorkerCNF)
+		workerRTNodes, err := nodes.GetByLabels(testutils.NodeSelectorLabels)
 		Expect(err).ToNot(HaveOccurred())
 		workerRTNodes, err = nodes.MatchingOptionalSelector(workerRTNodes)
 		Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("error looking for the optional selector: %v", err))
 		Expect(workerRTNodes).ToNot(BeEmpty())
 		workerRTNode = &workerRTNodes[0]
 
-		profile, err = profiles.GetByNodeLabels(
-			map[string]string{
-				fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerCNF): "",
-			},
-		)
+		profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
 		Expect(err).ToNot(HaveOccurred())
 		Expect(profile.Spec.HugePages).ToNot(BeNil())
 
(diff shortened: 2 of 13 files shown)


functests/0_config/config.go
```diff
@@ -38,26 +40,41 @@ var _ = Describe("[performance][config] Performance configuration", func() {
...
-		performanceManifest, found := os.LookupEnv("PERFORMANCE_PROFILE_MANIFEST_OVERRIDE")
-		if found {
+		profileAlreadyExists := false
+
+		if discovery.Enabled() {
 			var err error
-			performanceProfile, err = externalPerformanceProfile(performanceManifest)
-			Expect(err).ToNot(HaveOccurred(), "Failed overriding performance profile", performanceManifest)
-			klog.Warning("Consuming performance profile from ", performanceManifest)
+			performanceProfile, err = profiles.GetByNodeLabels(
```
fedepaol, line 47: Not sure it's doable, but I'd do another step further, checking for any profile regardless of the machine config pool.
As this is going to be run against an already configured cluster, we don't know what worker pool will it relate to.
In that case, while discovering it, you should also verify that the parameters are compliant with the tests.

functests/0_config/config.go
```diff
@@ -38,26 +40,41 @@ var _ = Describe("[performance][config] Performance configuration", func() {
...
-			err := testclient.Client.Create(context.TODO(), performanceProfile)
-			if errors.IsAlreadyExists(err) {
-				klog.Warning(fmt.Sprintf("A PerformanceProfile with name %s already exists! If created externally, tests might have unexpected behaviour", performanceProfile.Name))
-				profileAlreadyExists = true
-				return nil
-			}
-			return err
-		}, 15*time.Minute, 15*time.Second).ShouldNot(HaveOccurred(), "Failed creating the performance profile")
+		if !profileAlreadyExists {
+			By("Creating the PerformanceProfile")
```
fedepaol, line 66: I would invert the order. Do the creation if needed first (if `!discovery.Enabled`) and do the discovery after.
In this way the discovery logic should be able to find the profile we create.

functests/0_config/config.go
```diff
@@ -38,26 +40,41 @@ var _ = Describe("[performance][config] Performance configuration", func() {
...
+			performanceProfile, err = profiles.GetByNodeLabels(
+				map[string]string{
+					fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerCNF): "",
+				},
+			)
+			Expect(err).ToNot(HaveOccurred(), "Failed finding a performance profile in discovery mode")
+			profileAlreadyExists = true
+			klog.Info("Discovery mode: consuming a deployed performance profile from the cluster")
+		} else {
+			performanceManifest, found := os.LookupEnv("PERFORMANCE_PROFILE_MANIFEST_OVERRIDE")
```
fedepaol, line 56: I would put this part together with the creation.
Also, even if we are now adding this discovery mode, there's value in having a way to create a test profile and use it in subsequent test runs (this is because applying the profile is time consuming). Even when not in discovery mode, I'd try to see if the profile with the given name exists and in that case not do anything, maybe with a more explicit name (something like testprofile)

functests/utils/consts.go
```diff
@@ -20,6 +26,22 @@ func init() {
 	if PerformanceProfileName == "" {
 		PerformanceProfileName = "performance"
 	}
+
+	if discovery.Enabled() {
```
fedepaol, line 30: I'd rather move this logic in the discovery package and use it from here.
It doesn't feel it belongs to `const.go`

functests/1_performance/rt-kernel.go
```diff
@@ -30,33 +32,43 @@ var _ = Describe("[performance]RT Kernel", func() {
 
 	It("[test_id:26861][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] should have RT kernel enabled", func() {
 
-		Eventually(func() string {
+		profile := &performancev1alpha1.PerformanceProfile{}
+		key := types.NamespacedName{
+			Name: testutils.PerformanceProfileName,
+		}
+		err := testclient.Client.Get(context.TODO(), key, profile)
```
marcel-apf, line 38: I think the profile query should not be part of the test, each test may be asking for the profile. If some tests do not need the profile, but is not an issue

functests/1_performance/rt-kernel.go
```diff
@@ -30,33 +32,43 @@ var _ = Describe("[performance]RT Kernel", func() {
...
+				testpod.Spec.RestartPolicy = corev1.RestartPolicyNever
+				testpod.Spec.NodeSelector = map[string]string{
+					fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerCNF): "",
+				}
 
-			if err := pods.WaitForPhase(testpod, corev1.PodSucceeded, 60*time.Second); err != nil {
-				return ""
-			}
+				if err := testclient.Client.Create(context.TODO(), testpod); err != nil {
+					return ""
```
marcel-apf: Is the returning of an empty string some kind of error flow? I find it odd

functests/1_performance/rt-kernel.go
```diff
@@ -30,33 +32,43 @@ var _ = Describe("[performance]RT Kernel", func() {
...
+				}
 
-			logs, err := pods.GetLogs(testclient.K8sClient, testpod)
-			if err != nil {
-				return ""
-			}
+				if err := pods.WaitForPhase(testpod, corev1.PodSucceeded, 60*time.Second); err != nil {
```
marcel-apf: I would move the timeout to a global const, maybe even to an env variable, I kind of don't like literals

functests/1_performance/rt-kernel.go
```diff
@@ -30,33 +32,43 @@ var _ = Describe("[performance]RT Kernel", func() {
...
-			return logs
+				logs, err := pods.GetLogs(testclient.K8sClient, testpod)
+				if err != nil {
+					return ""
+				}
 
-		}, 15*time.Minute, 30*time.Second).Should(ContainSubstring("PREEMPT RT"))
+				return logs
+
+			}, 15*time.Minute, 30*time.Second).Should(ContainSubstring("PREEMPT RT"))
```
marcel-apf: Same here about the literals, but a test waiting 15 min? :)

functests/utils/consts.go
```diff
@@ -24,7 +30,25 @@ func init() {
 		PerformanceProfileName = "performance"
 	}
 
+<<<<<<< HEAD
 	NodesSelector = os.Getenv("NODES_SELECTOR")
+=======
+	if discovery.Enabled() {
+		performanceProfile, err := profiles.GetByNodeLabels(
+			map[string]string{
+				fmt.Sprintf("%s/%s", LabelRole, RoleWorkerCNF): "",
```
marcel-apf: My PR addresses this ad-hoc construct appearing in a lot of places, I'll be sure to add also this occurrence.

functests/utils/consts.go
```diff
@@ -24,7 +30,25 @@ func init() {
...
+			},
+		)
+		if err != nil {
+			klog.Error("Failed finding an existing performance profile in discovery mode", err)
```
marcel-apf: So is an error in logs, not connected to the tests error flow? Is this the standard way to expose tests results? I ask because I didn't see this approach in other places

functests/utils/discovery/discovery.go
```diff
@@ -0,0 +1,12 @@
+package discovery
+
+import (
+	"os"
+	"strconv"
+)
+
+// Enabled indicates whether test discovery mode is enabled.
+func Enabled() bool {
+	discoveryMode, _ := strconv.ParseBool(os.Getenv("DISCOVERY_MODE"))
```
marcel-apf, line 18: I would group all env variables together, but that's only me.

functests/utils/discovery/discovery.go
```diff
@@ -1,12 +1,58 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // Enabled indicates whether test discovery mode is enabled.
 func Enabled() bool {
 	discoveryMode, _ := strconv.ParseBool(os.Getenv("DISCOVERY_MODE"))
 	return discoveryMode
 }
+
+// GetDiscoveryPerformanceProfile returns an existing profile in the cluster with the most nodes using it.
```
marcel-apf, line 22: Why? I don't understand the benefit of using a profile used by most nodes? What is the upside? How does it help?

functests/utils/discovery/discovery.go
```diff
@@ -1,12 +1,58 @@
...
+// In case no profile exists - return nil
+func GetDiscoveryPerformanceProfile(exclude ...string) (*performancev1alpha1.PerformanceProfile, error) {
+	performanceProfiles, err := profiles.GetAllProfiles()
+	if err != nil {
+		return nil, err
```
marcel-apf, line 27: Test wise, would it return a "nil"  error? Would we need to go to the code to see were it failed or would we receive a nice "Error fetching profiles", and in this case, would we have some info on "why"? (is the operator healthy?)

functests/utils/mcps/mcps.go
```diff
@@ -3,9 +3,10 @@ package mcps
 import (
 	"context"
 	"fmt"
-	"k8s.io/klog"
```
marcel-apf, line 6: is the import ...

functests/utils/mcps/mcps.go
```diff
@@ -3,9 +3,10 @@ package mcps
...
 	"time"
 
+	"k8s.io/klog"
```
marcel-apf, line 8: moved here by a tool?

functests/utils/mcps/mcps.go
```diff
@@ -52,6 +55,18 @@ func GetByName(name string) (*machineconfigv1.MachineConfigPool, error) {
 	return mcp, err
 }
 
+// GetByProfile returns the MCP by a given performance profile
+func GetByProfile(performanceProfile *performancev1alpha1.PerformanceProfile) (string, error) {
+	mcpLabel := profile.GetMachineConfigLabel(performanceProfile)
+	key, value := components.GetFirstKeyAndValue(mcpLabel)
```
marcel-apf, line 61: If we have more of this, is it an error?

functests/utils/nodes/nodes.go
```diff
@@ -37,16 +37,24 @@ func GetBySelector(selector labels.Selector) ([]corev1.Node, error) {
 	return nodes.Items, nil
 }
 
-// GetNonRTWorkers returns list of nodes with no worker-cnf label
+// GetByLabels returns all nodes with the specified labels
+func GetByLabels(nodeLabels map[string]string) ([]corev1.Node, error) {
```
marcel-apf, line 41: Nice, we mostly don't need my PR anymore

functests/utils/nodes/nodes.go
```diff
@@ -37,16 +37,24 @@ func GetBySelector(selector labels.Selector) ([]corev1.Node, error) {
...
+
+// GetNonRTWorkers returns list of nodes with non matching perfomance profile labels
 func GetNonRTWorkers() ([]corev1.Node, error) {
 	nonRTWorkerNodes := []corev1.Node{}
 
 	workerNodes, err := GetByRole(testutils.RoleWorker)
 	for _, node := range workerNodes {
-		if _, ok := node.Labels[fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerCNF)]; ok {
-			continue
+		for label := range testutils.NodeSelectorLabels {
```
marcel-apf, line 52: Do you need all the node selectors to be "true" in order for a worker to be "RT"? If yes, the test is not correct

functests/utils/nodes/nodes.go
```diff
@@ -37,16 +37,24 @@ func GetBySelector(selector labels.Selector) ([]corev1.Node, error) {
...
 
-// GetNonRTWorkers returns list of nodes with no worker-cnf label
+// GetByLabels returns all nodes with the specified labels
+func GetByLabels(nodeLabels map[string]string) ([]corev1.Node, error) {
+	selector := labels.SelectorFromSet(nodeLabels)
+	return GetBySelector(selector)
+}
+
+// GetNonRTWorkers returns list of nodes with non matching perfomance profile labels
 func GetNonRTWorkers() ([]corev1.Node, error) {
```
marcel-apf, line 47: is the "RT/nonRT" naming right?

functests/utils/discovery/discovery.go
yanirq: @marcel-apf  we need to device a deterministic (although not super clever here) mechanism to get a profile to test against if we are in discovery mode (we will probably want to test the one with most nodes in CI)
this is ofcourse can get clever later on if we wish to test all profiles (Con - Time consuming)

functests/utils/discovery/discovery.go
yanirq: this is specific to discovery mode. what other variables are you referring to ?

functests/utils/discovery/discovery.go
yanirq: this is the way we move errors up the stack in case we encounter them (kindof throwable) a nicer way would be to wrap this error with another message and throw a new error but i think its redundant here and you will get the full path of the error here in case you want to debug the test code. 
There should be more places in code like this, we can consider a total refactor of all if you see fit

functests/utils/mcps/mcps.go
yanirq: yes, but ill move it to the right place

functests/utils/mcps/mcps.go
yanirq: no error since its getting only the First key and value. (maybe this should be more generic but it should be enough for the tests atm)

functests/utils/nodes/nodes.go
yanirq: good call

functests/utils/consts.go
yanirq: moved previous logic to discovery.go

functests/0_config/config.go
yanirq: posted changes that should answer these requested changes. hope the logic is not too complex now.

functests/1_performance/rt-kernel.go
yanirq: its tested against the clause in the end of eventually `.Should(ContainSubstring("PREEMPT RT")`
this is existing code

functests/1_performance/rt-kernel.go
yanirq: existing code . leaving it for a global re factoring in a different PR

functests/1_performance/rt-kernel.go
yanirq: once its found it stops, guess it was there as an extra time buffer

functests/1_performance/rt-kernel.go
marcel-apf: thanks

functests/1_performance/rt-kernel.go
marcel-apf: legacy already?

functests/1_performance/rt-kernel.go
marcel-apf: So maybe we should use predefined token like return FAILURE (var FAILURE = "") ?

functests/1_performance/rt-kernel.go
marcel-apf: agreed, a different PR is on order

functests/utils/mcps/mcps.go
marcel-apf: Why not fail? I don't like the approach "take the first one" since it looks like wishful thinking (i this case I wish the first one will pass get the test to pass :) )

functests/1_performance/rt-kernel.go
yanirq: can be, but im fine with the current form

yanirq:
@fedepaol  @marcel-apf  refactored nodelabels and mc labels handling to be more generic, fixes added according to reviews, please take a second look

functests/1_performance/rt-kernel.go
```diff
@@ -2,23 +2,51 @@ package __performance
...
+		for profilesCount > 0 {
+			if profile.Spec.RealTimeKernel != nil && *profile.Spec.RealTimeKernel.Enabled == true {
+				break
+			}
+			excludedProfiles = append(excludedProfiles, profile.Name)
+			profile, err = discovery.GetDiscoveryPerformanceProfile(excludedProfiles...)
+			profilesCount--
+		}
+		if profilesCount == 0 {
+			Skip("Skipping RT Kernel tests since no profile found with RT kernel set")
```
fedepaol, line 46: Nit: BeforeAll is "artificial".
If you skip here, you'll only skip the first test of the suite.
A decent compromise I found was to set a discoveryFailed flag in `BeforeAll` and checking it in `BeforeEach`

functests/utils/consts.go
```diff
@@ -25,6 +32,18 @@ func init() {
...
+		fmt.Sprintf("%s/%s", LabelRole, RoleWorkerCNF): "",
+	}
+
+	if discovery.Enabled() {
+		profile, err := discovery.GetDiscoveryPerformanceProfile()
+		if err == nil {
+			PerformanceProfileName = profile.Name
+		}
+		NodeSelectorLabels = profile.Spec.NodeSelector
+	}
```
fedepaol, line 46: I think you should use the content of `NODES_SELECTOR` (if present) and add another entry to the map here.

functests/1_performance/rt-kernel.go
```diff
@@ -2,23 +2,51 @@ package __performance
...
+			Name: testutils.PerformanceProfileName,
+		}
+		err := testclient.Client.Get(context.TODO(), key, profile)
+		Expect(err).ToNot(HaveOccurred(), "failed to get the performance profile")
+
+		profilesCount, err := profiles.GetProfilesCount()
+		Expect(err).ToNot(HaveOccurred())
+		excludedProfiles := []string{}
+
+		for profilesCount > 0 {
```
fedepaol, line 37: Not sure I get the purpouse here. You check the existance of _the_ profile, and look for alternatives.
I guess you can remove the initial `Get` as we you are looking for a specific one.
Also, passing a filter function to GetDiscoveryPerformanceProfile instead of a list would make this more efficient, as you would get the list only once.

functests/utils/consts.go
```diff
@@ -25,6 +32,18 @@ func init() {
 	}
 
 	NodesSelector = os.Getenv("NODES_SELECTOR")
+
+	NodeSelectorLabels = map[string]string{
```
fedepaol, line 36: Suggestion, feel free to ignore (but it's one of my fixations :p) 
How about exposing this in the nodes package and rename it as `SelectorLabels`?
That would read as `nodes.SelectorLabels` when you call it.
The package name is part of the object name and we should use it. By relying on "utils" we gave up part of our naming structure.

functests/utils/profiles/profiles.go
```diff
@@ -47,3 +47,21 @@ func GetConditionMessage(nodeLabels map[string]string, conditionType v1.Conditio
 	}
 	return ""
 }
+
+// GetAllProfiles gets all the exiting profiles in the cluster
+func GetAllProfiles() (*performancev1alpha1.PerformanceProfileList, error) {
```
fedepaol, line 52: The package is already `profiles`, this reads out as `profiles.GetAllProfiles`.
`profiles.GetAll` reads the same and sounds better, or even `profiles.All()`

functests/utils/profiles/profiles.go
```diff
@@ -47,3 +47,21 @@ func GetConditionMessage(nodeLabels map[string]string, conditionType v1.Conditio
...
+	profiles := &performancev1alpha1.PerformanceProfileList{}
+	if err := testclient.Client.List(context.TODO(), profiles); err != nil {
+		return nil, err
+	}
+	return profiles, nil
+}
+
+// GetProfilesCount returns the total number of existing perfomance profiles
+func GetProfilesCount() (int, error) {
```
fedepaol, line 61: Same, `profiles.Count()`

functests/1_performance/cpu_management.go
```diff
@@ -105,6 +102,16 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 	Describe("[test_id:27492][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Verification of cpu manager functionality", func() {
 		var testpod *corev1.Pod
 
+		testutils.BeforeAll(func() {
+			if discovery.Enabled() {
+				profile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+				Expect(err).ToNot(HaveOccurred())
+				if len(*profile.Spec.CPU.Isolated) == 1 {
+					Skip("Skipping tests since there are insufficant isolated cores to create a stress pod")
```
fedepaol, line 110: Sorry did not notice before, this is gonna cause only the first item of the table to skip.
You should the same trick you did in rt, setting a flag and skipping in beforeEach

yanirq: @fedepaol  done

PR #291, opened 2020-07-06
Makefile: add targets for must-gather

fromanirh:
Add targets and rules to create the must-gather image.
The end goal is to have our CI automatically create those images
nightly, not only on releases.


cynepco3hahue:
@fromanirh Any reason why does it on hold?

fromanirh:
> @fromanirh Any reason why does it on hold?

A silly one: I uploaded the PR before the quay repo was created. I think this is no longer needed.

PR #293, opened 2020-07-06
metadata: add metadata image, to be used with OPM

fromanirh:
Not long ago the OLM gained support for the OPM tooling, which changes
the way we pack and deliver the metadata to be consumed by the OLM
itself. The new, recommended flow, makes use of the so-called "operator
bundles" and of the "index images".

See: https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md

For our intents and purposes, the main benefit here is that we don't
need to maintain anymore binary registry images, but rather only
much simpler and less challenging metadata images.

### DIFF
--- a/openshift-ci/Dockerfile.metadata
+++ b/openshift-ci/Dockerfile.metadata
@@ -1,0 +1,17 @@
+# see: https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md
+FROM scratch
+
+ARG CHANNEL=4.6
+ARG FULL_OPERATOR_IMAGE=quay.io/openshift-kni/performance-addon-operator-metadata:4.6-snapshot
+ARG OLM_SOURCE=deploy/olm-catalog
+ARG OLM_METADATA=deploy/metadata
+
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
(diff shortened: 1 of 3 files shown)


openshift-ci/Dockerfile.metadata
```diff
@@ -0,0 +1,17 @@
...
+ARG OLM_METADATA=deploy/metadata
+
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=performance-addon-operator
+LABEL operators.operatorframework.io.bundle.channels.v1=4.6.0
+LABEL operators.operatorframework.io.bundle.channel.default.v1=4.6.0
+
+COPY ${OLM_SOURCE}/performance-addon-operator/${CHANNEL}.0 /manifests/
```
fromanirh, line 16: we need to replace FULL_OPERATOR_IMAGE here in the csv, otherwise the pullspec will be broken. But since this image is FROM SCRATCH we can't do it in the Dockerfile.

openshift-ci/Dockerfile.metadata
```diff
@@ -0,0 +1,17 @@
+# see: https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md
+FROM scratch
+
+ARG CHANNEL=4.6
```
cynepco3hahue, line 4: Will we need to update the dockerfile for each version?

fromanirh: the idea was not to, that's because I used ARG.

PR #294, opened 2020-07-06
functest/nodes: allow nodes matching by an optional selector

marcel-apf:
Currently the nodes for e2e tests are selected by a 'role' label.
Sometimes the granularity is not enough and there is a need to filter
the 'role' nodes using a standard label.

Allow the filtering based on a "NODES_SELECTOR" environment variable.

### DIFF
--- a/functests/utils/nodes/nodes.go
+++ b/functests/utils/nodes/nodes.go
@@ -115,3 +115,39 @@ func GetKubeletConfig(node *corev1.Node) (*kubeletconfigv1beta1.KubeletConfigura
 	}
 	return kubeletConfig, err
 }
+
+// MatchingOptionalSelector filter the given slice with only the nodes matching the optional selector.
+// If no selector is set, it returns the same list.
+// The NODES_SELECTOR must be set with a labelselector expression.
+// For example: NODES_SELECTOR="sctp=true"
+// Inspired from: https://github.com/fedepaol/sriov-network-operator/blob/master/test/util/nodes/nodes.go
+func MatchingOptionalSelector(toFilter []corev1.Node) ([]corev1.Node, error) {
+	if testutils.NodesSelector == "" {
+		return toFilter, nil
+	}
+
+	selector, err := labels.Parse(testutils.NodesSelector)
+	if err != nil {
+		return nil, fmt.Errorf("Error parsing the %s label selector, %v", testutils.NodesSelector, err)
+	}
+
+	toMatch, err := GetBySelector(selector)
+	if err != nil {
+		return nil, fmt.Errorf("Error in getting nodes matching the %s label selector, %v", testutils.NodesSelector, err)
+	}
+	if len(toMatch) == 0 {
+		return nil, fmt.Errorf("Failed to get nodes matching %s label selector", testutils.NodesSelector)
+	}
+
+	res := make([]corev1.Node, 0)
+	for _, n := range toFilter {
+		for _, m := range toMatch {
+			if n.Name == m.Name {
+				res = append(res, n)
+				break
+			}
+		}
+	}
+
+	return res, nil
+}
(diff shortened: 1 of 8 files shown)


marcel-apf:
The current proposal first filters the nodes by role, then filters the results by an optional label.
The alternative would be that if that optional label exists, to not filter by role. I would opt for the second option.

functests/utils/nodes/nodes.go
```diff
@@ -19,6 +20,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// NodesSelector represents the label selector used to filter impacted nodes.
+var NodesSelector string
```
fromanirh, line 24: Any good reason why we use a global? all considered this should work fine, but I dislike globals in general and I'd prefer to actually call getenv() each time MatchingOptionalSelector is invoked.
What do other reviewers think?

marcel-apf: This is an artifact of the old copy-paste technique. I do prefer to avoid global variables and we can always promote it when/if we will need.

marcel-apf: @fromanirh thinking a little bit more, keeping it a global may not be such a bad idea. We already have RoleWorkerCNF and PerformanceProfileName globals that do the same. An alternative is to not use RoleWorkerCNF global directly but getting the env variable as part of nodes.GetByRole every time we need.

However the side effect is that one could *change* the env variable during the tests and some the tests *may* run with different values. So using is as a global can be used in order to emphases we only get the env variable once.

Moving it to utils/consts.go

marcel-apf: Moved and re-pushed

fromanirh:
Good enough for me as well. Let's discuss the globals in another PR :)

fromanirh:
I think we need to check what's happening on our drone.io instance:
```

write /tmp/go-build716463573/b322/vet.cfg: no space left on device
--
```

cynepco3hahue:
@fromanirh I will try to find what the problem, I just afraid it has some limited space under the /tmp and it not open for the configuration.

cynepco3hahue:
@marcel-apf I see some error under the openshift CI
```
hack/verify-generated.sh
 M functests/utils/nodes/nodes.go
```
Can you please fix it?

functests/utils/nodes/nodes.go
```diff
@@ -115,3 +115,39 @@ func GetKubeletConfig(node *corev1.Node) (*kubeletconfigv1beta1.KubeletConfigura
...
+		for _, m := range toMatch {
+			if n.Name == m.Name {
+				res = append(res, n)
+				break
+			}
+		}
+	}
+
+	return res, nil
+}
```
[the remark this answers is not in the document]
marcel-apf: After the function? Will do.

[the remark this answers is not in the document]
marcel-apf: Done

marcel-apf:
> @marcel-apf I see some error under the openshift CI
> 
> ```
> Verifying that all code is committed after updating deps and formatting and generating code
> hack/verify-generated.sh
> uncommitted generated files. run 'make generate' and commit results.
>  M functests/utils/nodes/nodes.go
> ```
> 
> Can you please fix it?
Sure, on it

marcel-apf:
> > @marcel-apf I see some error under the openshift CI
> > ```
> > Verifying that all code is committed after updating deps and formatting and generating code
> > hack/verify-generated.sh
> > uncommitted generated files. run 'make generate' and commit results.
> >  M functests/utils/nodes/nodes.go
> > ```
> > 
> > 
> > Can you please fix it?
> > Sure, on it
Fixed by prev "push", so done.

PR #295, opened 2020-07-07
Verify irqbalance tuning is reverted

gsr-shanks:
https://bugzilla.redhat.com/show_bug.cgi?id=1854245

### DIFF
--- a/functests/2_performance_update/updating_profile.go
+++ b/functests/2_performance_update/updating_profile.go
@@ -4,10 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -30,6 +30,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 	chkKernel := []string{"uname", "-a"}
 	chkCmdLine := []string{"cat", "/proc/cmdline"}
 	chkKubeletConfig := []string{"cat", "/rootfs/etc/kubernetes/kubelet.conf"}
+	chkIrqbalance := []string{"cat", "/rootfs/etc/sysconfig/irqbalance"}
 
 	nodeLabel := map[string]string{fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerCNF): ""}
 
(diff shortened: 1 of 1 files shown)


functests/2_performance_update/updating_profile.go
```diff
@@ -30,6 +30,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 	chkKernel := []string{"uname", "-a"}
 	chkCmdLine := []string{"cat", "/proc/cmdline"}
 	chkKubeletConfig := []string{"cat", "/rootfs/etc/kubernetes/kubelet.conf"}
+	chkIrqbalance := []string{"grep", "-ls" + "ffffffff,fffffffe" + "/etc/sysconfig/irqbalance"}
```
fromanirh, line 33: not sure the "+" is correct here. Shouldn't at least the last component (`/etc/sysconfig/irqbalance`) separated by a comma?

gsr-shanks: Fixed. Thanks.

functests/2_performance_update/updating_profile.go
```diff
@@ -269,6 +270,10 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			kblcfg, err := nodes.ExecCommandOnNode(chkKubeletConfig, newCnfNode)
 			Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("failed to execute %s", chkKubeletConfig))
 			Expect(kblcfg).NotTo(ContainSubstring("reservedSystemCPUs"))
+
+			irqbal, err := nodes.ExecCommandOnNode(chkIrqbalance, newCnfNode)
+			Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("failed to execute %s", chkIrqbalance))
+			Expect(irqbal).NotTo(ContainSubstring("ffffffff,fffffffe"))
```
dshchedr, line 276: "ffffffff,fffffffe" - is only for scenario with  profile parameter  **reserved: "0"**
So, this test may not work correctly on cluster with different "reserved" value. 

How about adding calculation of this value based on profile setting?

marcel-apf:
> BZ1854245

II would use a real link for the BZ, especially for the lazy people around us :)

gsr-shanks:
> > BZ1854245
> 
> II would use a real link for the BZ, especially for the lazy people around us :)

Sure I am one of them :) 
I was hoping(lazy) that it would somehow auto-link itself. Will fix it.

functests/2_performance_update/updating_profile.go
gsr-shanks: Agree it is good to have it this way.

but before that, it needs to be investigated why we see the following in ci/prow/e2e-gcp which probably means that on reverting the label the file is being removed in which case we would need to check if the file exists.

```
  out=cat: /etc/sysconfig/irqbalance: No such file or directory
```

gsr-shanks: Missed to add _/rootfs_ prefix for the no such file error. Fixed it.

PR #299, opened 2020-07-07
scripts: add target to generate the manifests

fromanirh:
Up until now, we fixed the manifests pullspec inside the relevant
Dockerfile, most prominent example is the registry dockerfile.
For the metadata image, we can't do that: the Dockerfile starts FROM
scratch.

Doing it in the relevant makefile target turns out to be a great idea
either, because we want it to be a separate -and composable- step.
IOW fixing manifests should not be tied to (re)generate them from dev.

So this PR moves the manifest generation (which ATM just means "fixing
the pullspec") in its separate composable target.
In order to avoid any disruption to the existing flows, we generate the
manifests intentionally in another subtree under `build/_output`, and we
consume it in the metadata container; the existing targets aren't
affected at all.

Since we planm to phase out the registry images soon enough, we will
later just clean up the registry targets and keep this new flow only.


fromanirh:
/cc @cynepco3hahue @yanirq @MarSik 
openshift CI didn't like the current solution (https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_release/10091/rehearse-10091-pull-ci-openshift-kni-performance-addon-operators-master-e2e-gcp/1280538023213142016) hence this PR

marcel-apf:
I have carefully read the explanations and the code, had a look to the prev error and as far I am concerned:
/lgtm

Makefile
```diff
@@ -150,6 +149,10 @@ generate-latest-dev-csv: operator-sdk dist-csv-generator
 generate-docs: dist-docs-generator
 	hack/docs-generate.sh
 
+.PHONY: generate-manifests-tree
+generate-manifests-tree:
+	hack/generate-manifests-tree.sh
```
MarSik, line 154: I see this takes one argument. Where is it used when not here?

hack/generate-manifests-tree.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -e
+
+OPERATOR_IMAGE="${1:-REPLACE_IMAGE}"
+OUT_DIR="build/_output/manifests"
+
+rm -rf "${OUT_DIR}"
+mkdir -p "${OUT_DIR}"
+cp -a deploy/olm-catalog/performance-addon-operator "${OUT_DIR}"
+find "${OUT_DIR}" -type f -exec sed -i "s|REPLACE_IMAGE|${OPERATOR_IMAGE}|g" {} \; || :
```
MarSik, line 11: I see you default to REPLACE_IMAGE -> REPLACE_IMAGE replacement. Is that robust enough? Would maybe a check based on $# be better?

Makefile
fromanirh: It is no used. This code wants to do what the old code did, which allowed such optional argument which was never actually used in CI as far as I can tell - corrections welcome!

fromanirh: good catch. A variable should have been passed from the Makefile, like the old flow did. Fixing.

hack/generate-manifests-tree.sh
fromanirh: Considering the previous comment, it is probably better to exit with error from this script if the argument is missing.

Makefile
```diff
@@ -36,7 +36,7 @@ BUILD_DATE=$$(date --utc -Iseconds)
 export GO111MODULE=on
 
 .PHONY: build
-build: gofmt golint govet dist
+build: gofmt golint govet dist generate-manifests-tree
```
yanirq, line 39: why are we adding this to build ?

yanirq:
/lgtm
/hold

overall looks good to me. left a small question. feel free to release hold or wait for other reviews if needed

Makefile
fromanirh: because this is the target the openshift CI calls. I believe this is OK because we just create different manifest tree that only the new targets cares about (metadata-container related targets), but I'm open to suggestions here.

yanirq:
@fromanirh  Can we rename Dockerfile.metadata to Dockerfile.bundle ? (just for consistency with the OLM docs)

fromanirh:
> @fromanirh Can we rename Dockerfile.metadata to Dockerfile.bundle ? (just for consistency with the OLM docs)

We can and I will. Let me do this in this PR.

yanirq:
@fromanirh  I see that currently we are using docker cmd directly to create the bundle image (using manually configured annotaion.yaml and dockerfile.bundle)
https://github.com/openshift-kni/performance-addon-operators/blob/4c605c2d58b56e4c9e559c62c91b2e395145308f/Makefile#L105

We should consider (If its not a mess) using opm build  and verify (https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md#build-bundle-image)  that will auto-generate the annotations and the dockerfile, build the image and validate it. (Not for this PR)

fromanirh:
> @fromanirh I see that currently we are using docker cmd directly to create the bundle image (using manually configured annotaion.yaml and dockerfile.bundle)
> 
> https://github.com/openshift-kni/performance-addon-operators/blob/4c605c2d58b56e4c9e559c62c91b2e395145308f/Makefile#L105
> 
> We should consider (If its not a mess) using opm build and verify (https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md#build-bundle-image) that will auto-generate the annotations and the dockerfile, build the image and validate it. (Not for this PR)

AFAIK we cannot do this easily from neither the openshift nor the drone CIs. Will check again though.

Makefile
```diff
@@ -98,11 +98,10 @@ operator-container: build
...
 	$(IMAGE_BUILD_CMD) build --no-cache -f openshift-ci/Dockerfile.deploy -t $(FULL_OPERATOR_IMAGE) --build-arg BIN_DIR="_output/bin/" --build-arg ASSETS_DIR="assets" build/
 
-.PHONY: metadata-container
-metadata-container: generate-latest-dev-csv
+.PHONY: bundle-container
+bundle-container: generate-latest-dev-csv
 	@echo "Building the performance-addon-operator metadata image"
-	@find build/_output/olm-catalog/ -type f -exec sed -i "s|REPLACE_IMAGE|${FULL_OPERATOR_IMAGE}|g" {} \; || :
-	$(IMAGE_BUILD_CMD) build --no-cache -f openshift-ci/Dockerfile.metadata -t "$(FULL_METADATA_IMAGE)" .
+	$(IMAGE_BUILD_CMD) build --no-cache -f openshift-ci/Dockerfile.bundle -t "$(FULL_METADATA_IMAGE)" .
```
yanirq, line 104: lets change *METADATA_IMAGE  to *BUNDLE_IMAGE for consistency

fromanirh: done!

PR #300, opened 2020-07-08
Test verifies that cluster can have multiple profiles

dshchedr:


fromanirh:
should the commit message read "multiple profiles"?

dshchedr:
> should the commit message read "multiple profiles"?

dshchedr:
> should the commit message read "multiple profiles"?
should be multiple, thanks 

oops, closed  pr accidentally by answering to your comment :)

fedepaol:
Can we put this under a flag or have a test name that is easier to skip? 
It's going to have a big impact on tests running time I think, I would like to document it in CNF-tests

dshchedr:
> Can we put this under a flag or have a test name that is easier to skip?
> It's going to have a big impact on tests running time I think, I would like to document it in CNF-tests

No, should not have big impact on time since we don't create new mcp and don't set new label to any nodes.. it's just creates new profile and verifies that pao does the right job (creates new KubeletConfig and new MachineConfig)

fedepaol:
> > Can we put this under a flag or have a test name that is easier to skip?
> > It's going to have a big impact on tests running time I think, I would like to document it in CNF-tests
> 
> Should not have big impact on time since we don't create new mcp and don't set new label to any nodes.. it's just creates new profile and verifies that pao does the right job (creates new KubeletConfig and new MachineConfig)

Ah, good then! Thanks!

functests/1_performance/performance.go
```diff
@@ -189,6 +194,83 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+			Expect(testclient.Client.Delete(context.TODO(), secondProfile)).ToNot(HaveOccurred())
+			key = types.NamespacedName{
+				Name:      components.GetComponentName(profile.Name, components.ComponentNamePrefix),
+				Namespace: components.NamespaceNodeTuningOperator,
+			}
+			err = testclient.GetWithRetry(context.TODO(), key, &machineconfigv1.KubeletConfig{})
+			Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("cannot find KubeletConfig object %s", key.Name))
+			err = testclient.GetWithRetry(context.TODO(), key, &machineconfigv1.MachineConfig{})
+			Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("cannot find MachineConfig object %s", key.Name))
+
```
gsr-shanks, line 271: Should we check again that the initial MCP does not start updating during delete?

dshchedr: I'm not sure we need to cover it since it is the MCO logic, not PAO..  but yeah, maybe it would be good to add one more verification just in case

PR #302, opened 2020-07-09
makefile: fix target interdependencies

fromanirh:
Fix and document the dependencies/priorities between the
`registry-container` and the `bundle-container` targets.


fromanirh:
> So at what step we build the csv?

The step which regenerates the csv is `generate-latest-dev-csv`

PR #305, opened 2020-07-20
Support multiple APIs (v1, v1alpha1)

marcel-apf:


fromanirh:
bad rebase?

pkg/apis/performance/performance_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/ghodss/yaml"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	v1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1alpha1"
+	v1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
```
fromanirh, line 9: the renamed import (`v1`) seems unneeded now

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/coreos/go-systemd/unit"
 	igntypes "github.com/coreos/ignition/config/v2_2/types"
 
-	performancev1alpha1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1alpha1"
+	performancev1alpha1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
```
fromanirh, line 14: why do we need this rename?

pkg/controller/performanceprofile/components/profile/profile.go
```diff
@@ -2,8 +2,8 @@ package profile
 
 import (
 	"fmt"
+	v1alpha1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
```
fromanirh, line 5: same. Also, this import belongs in the group below.

pkg/controller/performanceprofile/components/profile/profile_test.go
```diff
@@ -1,7 +1,7 @@
 package profile
 
 import (
-	"github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1alpha1"
+	v1alpha1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
```
fromanirh, line 4: same

pkg/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -8,7 +8,7 @@ import (
 	"strings"
 	"text/template"
 
-	performancev1alpha1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1alpha1"
+	performancev1alpha1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
```
fromanirh, line 11: same (and for all other files in this package)

functests/1_performance/performance.go
```diff
@@ -135,6 +136,17 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 			}
 		})
 
+		It("[test_id:32375][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] initramfs should not have injected configuration", func() {
```
yanirq, line 139: this pops out after a rebase ? try rebasing anyway

dshchedr:
just curious why do you have 2 commits in this PR, probably you need to rebase to master since @gsr-shanks was already merged and modified after that..

functests/1_performance/performance.go
dshchedr: may be because he has second commit in this PR?
https://github.com/openshift-kni/performance-addon-operators/pull/305/commits/a617ae386c529748702682d59d0a859d2ee30568

pkg/apis/performance/performance_test.go
marcel-apf: will check, thanks

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
marcel-apf: It was a temporary decision to remove v1alpha1 all-together, I don't think is needed anymore

pkg/controller/performanceprofile/components/profile/profile.go
marcel-apf: ok

pkg/controller/performanceprofile/components/profile/profile_test.go
marcel-apf: ok

pkg/controller/performanceprofile/components/tuned/tuned.go
marcel-apf: ok

functests/1_performance/performance.go
marcel-apf: Bad rebase for sure

marcel-apf:
> just curious why do you have 2 commits in this PR, probably you need to rebase to master since @gsr-shanks was already merged and modified after that..

For sure! Thanks

yanirq:
`functests/utils/discovery/discovery.go:43:66: cannot use performanceProfiles.Items (type []"github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1".PerformanceProfile) as type []"github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1alpha1".PerformanceProfile in argument to filter`

Amend the import under discovery.go

pkg/apis/performance/performance_test.go
marcel-apf: right, is already called v1!

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
marcel-apf: Wait, is not a rename, we point to another API, I will change also performancev1alpha1 -> performancev1 in all the controller code

marcel-apf:
> `functests/utils/discovery/discovery.go:43:66: cannot use performanceProfiles.Items (type []"github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1".PerformanceProfile) as type []"github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1alpha1".PerformanceProfile in argument to filter`
> 
> Amend the import under discovery.go

I'll change, thanks!

fromanirh:
> /unhold

/hold cancel

marcel-apf:
> > /unhold
> 
> /hold cancel

Got it

cluster-setup/upgrade-test-cluster/performance/performance_profile.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: performance.openshift.io/v1alpha1
+apiVersion: performance.openshift.io/v1
```
dshchedr, line 1: upgrade test failed with error:
`error: unable to recognize "STDIN": no matches for kind "PerformanceProfile" in version "performance.openshift.io/v1"`

My assumption: in upgrade test we are using subscription channel 4.5, so may be we need to keep `v1alpha1` here?

dshchedr, line 1: upgrade test failed with error:
`error: unable to recognize "STDIN": no matches for kind "PerformanceProfile" in version performance.openshift.io/v1" `

In upgrade test workflow we are using this yaml file for creating profile and or some reason v1 does not know about PerformanceProfile objects..

marcel-apf: How does the test work? What is the first profile we update from? I will look in order to understand. All other e2e tests are passing, so there is something I might have missed.

marcel-apf: I think the problem is the "InitialProfile" is from 4.5, meaning v1alpha1, so if the test is asking for a v1 profile BEFORE the upgrade, it will fail.

marcel-apf: Looking at the prowl yaml, the base image is: ee686d6 which is BEFORE the PR and the job runs e2e-gcp-operator-upgrade, so is it possible we start with a prev PAO (before the PR code) ?

marcel-apf:
> I miss some e2e tests and we should fix the upgrade test, but except these, looks good to me

Thanks!

Re e2e tests:
The basics are actually tested since the "upgrade" ci tests the latest version can work with v1alpha1 profiles, while the regular ci tests that PAO works fine with v1, so we have the OK we can serve both.

I am planning to add a testplan for another ci test: "Verify we can work with multiple API versions in the same time", even if is not actually necessary, since every v1alpha1 is converted "on the fly" and served as "v1", so if the "upgrade" test does work, is actually enough.

Anyway, I will think more about more testplans, but out of the scope of this PR and inside the scope of https://issues.redhat.com/browse/CNF-822.


All that assuming the upgrade test will actually pass!

cluster-setup/upgrade-test-cluster/performance/performance_profile.yaml
marcel-apf: Resolved, thanks!

pkg/apis/performance/v1/zz_generated.deepcopy.go
```diff
@@ -5,7 +5,7 @@
 package v1
 
 import (
-	v1 "github.com/openshift/custom-resource-status/conditions/v1"
+	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
```
fromanirh, line 8: ok, this works and I don't really mind that much, but still, why exactly is this named import needed? what happens if you keep the default import?

fromanirh:
the missing test can be added in a followup PR
/lgtm

pkg/apis/performance/v1/zz_generated.deepcopy.go
marcel-apf: The file will be changed every time we run make generate. Better to do it ourselves, I think.

marcel-apf:
> nice work, kudos.

Thanks!

> I think the only real missing thing is a func/e2e tests which explicitely checks that if you feed a v1alpha1 manifest everything works and you get back a v1 manifest.
> In other words: great the upgrade tests covers this flow, but better to check this explicitely and avoid incidental coverage -which may fade in the coming months, unless we add indeed an explicit check.

I totally agree. The question is how to do it. We cannot create a v1alpha1 CR with the latest PAO. Only a prev PAO can do that.
I will think of a workaround.

marcel-apf:
> the missing test can be added in a followup PR
> /lgtm

Appreciated!

marcel-apf:
Maybe install a v1alpha1 crd and cr without PAO, then install PAO and apply the cr to a node.

PR #308, opened 2020-07-23
generate-manifests-index.sh: fix volume mount

fromanirh:
add missing flag to make the script work nicely with selinux.
Previously, without this flag, the script used to fail with

  panic: unable to open database file

selinux log revealed the culprit:

  avc:  denied  { write } for  pid=35664 comm="initializer" name="database" dev="dm-4" ino=393268 scontext=system_u:system_r:container_t:s0:c809,c992 tcontext=unconfined_u:object_r:user_tmp_t:s0 tclass=dir permissive=0


fromanirh:
@cynepco3hahue

PR #310, opened 2020-07-27
Dockerfiles: remove useless argument

fromanirh:
We don't really need the "FULL_OPERATOR_IMAGE" in the bundle
Dockerfiles: it is only needed in the build step, when we update the
manifests. IOW, when we build the containers, the replacement must have
been done already.

Clean up for clarity; the variable was left over because of a copypaste
artifact.


fromanirh:
@yanirq @cynepco3hahue

PR #313, opened 2020-07-27
Fix interaction diagram

MarSik:
There was an error in the depiction of isolated behavior.
The proper way is to have isolated span all cores that are
not reserved and let CPU manager to dedicate cores to
containers that ask for it.

There were also style cleanups:
    
- The presentation title was removed,
- line styles were fixed,
- old Google Slides export artifacts cleaned up,
etc.


MarSik:
I need to fix the diagram to clearly show that the tuning works properly only when all cores are specified in the profile and reserved and isolated are kept complementary.

PR #315, opened 2020-07-29
docs/performance_profile: mention managed_irq

marcel-apf:
Emphasize the isolated CPUs won't be handling managed IRQs.


yanirq:
I don't see this as something we really need under the doc for the field.
We have a lot of "hidden from the user" settings going behind the scenes which we don't have to mention.
Either way im fine with leaving this but leaning towards dropping it

marcel-apf:
I do think if we don't pass to users ALL the information, this piece of information is not necessary

PR #318, opened 2020-08-02
Fix the RT kernel e2e test

cynepco3hahue:
Improve the PREEMPT RT check, to avoid situation when the kernel version
will have unexpected changes, like the recent change from `PREEMPT RT` to
`PREEMPT_RT`.

The change occured under the kernel version:
```
Linux test-66vhx 4.18.0-211.rt5.23.el8.x86_64 #1 SMP PREEMPT_RT Thu Jun 4 13:01:45 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

The change includes:
- check if the node booted with the RT kernel
- check if the RT kernel compiled with the `CONFIG_PREEMPT_RT=y`

### DIFF
--- a/functests/1_performance/rt-kernel.go
+++ b/functests/1_performance/rt-kernel.go
@@ -1,24 +1,18 @@
 package __performance
 
 import (
-	"context"
-	"time"
+	"fmt"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
 	testutils "github.com/openshift-kni/performance-addon-operators/functests/utils"
-	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/discovery"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/nodes"
-	"github.com/openshift-kni/performance-addon-operators/functests/utils/pods"
 	performancev1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
-
-	corev1 "k8s.io/api/core/v1"
 )
 
 var _ = Describe("[performance]RT Kernel", func() {
-	var testpod *corev1.Pod
 	var discoveryFailed bool
 	var profile *performancev1.PerformanceProfile
 	var err error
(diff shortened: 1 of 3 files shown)


functests/1_performance/rt-kernel.go
```diff
@@ -80,7 +80,7 @@ var _ = Describe("[performance]RT Kernel", func() {
 
 			return logs
 
-		}, 15*time.Minute, 30*time.Second).Should(ContainSubstring("PREEMPT RT"))
+		}, 15*time.Minute, 30*time.Second).Should(ContainSubstring(testutils.RTKernel))
```
fromanirh, line 83: please document (commit message) when this changed: first version of distro + kernel (at least either, better both) which highlighted this issue.

functests/utils/consts.go
```diff
@@ -81,3 +81,8 @@ const (
 	// ContainerMachineConfigDaemon contains the name of the machine-config-daemon container
 	ContainerMachineConfigDaemon = "machine-config-daemon"
 )
+
+const (
+	// RTKernel contains the line that the RT kernel should contain under the uname -a
+	RTKernel = "PREEMPT_RT"
```
fromanirh, line 87: we should investigate why this value changed in the first place. Different patchset? This can be done once CI is unblocked, but we should not forget about this.

functests/utils/nodes/nodes.go
```diff
@@ -158,3 +158,20 @@ func MatchingOptionalSelector(toFilter []corev1.Node) ([]corev1.Node, error) {
 
 	return res, nil
 }
+
+// HasPreemptRTKernel returns no error if the node booted with PREEMPT RT kernel
+func HasPreemptRTKernel(node *corev1.Node) error {
+	// verify that the kernel-rt-core installed it also means the the machine booted with the RT kernel
+	// because the machine-config-daemon uninstalls regular kernel once you install the RT one
+	cmd := []string{"rpm", "-q", "kernel-rt-core"}
```
fromanirh, line 166: please add to the comment above that
```
on traditional yum systems, rpm -q kernel can be completely different from what you're booted because yum keeps multiple kernels but only one userspace; with rpm-ostree rpm -q is telling you what you're booted into always, because ostree binds together (kernel, userspace) as a single commit.
```

fromanirh:
thanks!

cynepco3hahue:
@fromanirh I ran all tests locally and all passed, so I want to merge this one, WDYT?

fromanirh:
@cynepco3hahue the failure seems caused by a issue adding/removing arguments, which is unrelated to this PR. So let's merge it. Let us watch CI afterwards to doublecheck this didn't break anything (very unlikely, but still that's the price to pay when we skip automation...)

cynepco3hahue:
@fromanirh Passed, can you please give lgtm.

PR #319, opened 2020-08-05
fix for PAO cleaning script

dshchedr:
this is a fix for nodeSelector parser that gets incorrect data:
```
$ oc get performanceprofile.performance.openshift.io/manual -o=jsonpath='{.spec.nodeSelector}' | awk -F'[/:]' '{print $2}'
worker-cnf"
```
as result node label and mcp were not removed.

### DIFF
--- a/hack/clean-deploy.sh
+++ b/hack/clean-deploy.sh
@@ -6,7 +6,7 @@ OC_TOOL="${OC_TOOL:-oc}"
 profiles=$(${OC_TOOL} get performanceprofile -o name)
 for profileName in $profiles
 do
-  nodeSelector="$(${OC_TOOL} get $profileName -o=jsonpath='{.spec.nodeSelector}'  | awk -F'[/:]' '{print $2}')"
+  nodeSelector="$(${OC_TOOL} get $profileName -o=jsonpath='{.spec.nodeSelector}'  | awk -F'[/"]' '{print $3}')"
 
   if [[ $nodeSelector != "worker" ]]; then
     mcps+=($(${OC_TOOL} get mcp -l machineconfiguration.openshift.io/role=$nodeSelector -o name | awk -F "/" '{print $2}'))


hack/clean-deploy.sh
```diff
@@ -6,7 +6,7 @@ OC_TOOL="${OC_TOOL:-oc}"
 profiles=$(${OC_TOOL} get performanceprofile -o name)
 for profileName in $profiles
 do
-  nodeSelector="$(${OC_TOOL} get $profileName -o=jsonpath='{.spec.nodeSelector}'  | awk -F'[/:]' '{print $2}')"
+  nodeSelector="$(${OC_TOOL} get $profileName -o=jsonpath='{.spec.nodeSelector}'  | awk -F'[/:"]' '{print $3}')"
```
gsr-shanks, line 9: Can we use this instead? awk and print seem to be unreliable. 

`basename `oc get performanceprofiles.performance.openshift.io performance -o json | jq -cr '.spec.nodeSelector | keys[]'``

marcel-apf:
Please add the PR description to the commit message.

hack/clean-deploy.sh
marcel-apf: +1. If we can without them... I am not referring to the proposed change. In general.

fromanirh: I like more the jq approach, but could you please clarify what (and when?) awk is unreliable?

gsr-shanks: Like in this case where the required data was not returned. The awk in itself is not unreliable though. With jq approach, we don't have to worry about the delimiter and print field.

fromanirh: thanks, good point indeed. Definitely +1 for jq.

yuvalk: +1 on using jq.
that's also what we're doing in cnf-feature-deploy and d/s

cynepco3hahue: Just my 5c, IIRC jq not installed by default under the OS, and awk installed.

fromanirh: Right. So if we use jq, we need to add a check jq is there before we actually run commands that can modify the cluster state.

dshchedr: let me explain slightly. First of all I agree with @cynepco3hahue that **jq** may not be present on a server (not a big deal, of course, we can add verification).

Secondly - even if we use jq command - we will need to split the result:

```
# oc get performanceprofiles.performance.openshift.io manual -o json | jq -cr '.spec.nodeSelector | keys[]'
node-role.kubernetes.io/worker-cnf
```
From this output I need to get only **_worker-cnf_**, so anyway I need to use **awk** or something similar for splitting it on 2 parts.. so, the command in this case will look like:
 
`oc get performanceprofiles.performance.openshift.io manual -o json | jq -cr '.spec.nodeSelector | keys[]' | awk -F "/" '{print $2}'`


Comparing this two commands - I don't see any benefits from switching to jq..

gsr-shanks: You can use `basename node-role.kubernetes.io/worker-cnf` instead of awk.

fromanirh: @dshchedr while I see @gsr-shanks 's point, I've nothing against awk, so considering your last point I'm fine keeping awk.

fromanirh:
> this is a fix for nodeSelector parser that gets incorrect data:
> 
> ```
> $ oc get performanceprofile.performance.openshift.io/manual -o=jsonpath='{.spec.nodeSelector}' | awk -F'[/:]' '{print $2}'
> worker-cnf"
> ```
> 
> as result node label and mcp were not removed.

the only pendingi item to merge is to copy this content to the commit message, to have consistent history on git. Reason is git history should be self-sufficient, without *needing* to go back on github to understand why we did what we did.

dshchedr:
> Please add the PR description to the commit message.
done

PR #320, opened 2020-08-06
Provide the latency e2e test and container image

cynepco3hahue:
As first step we should only to test that the oslat image
works correctly, the next step will be to define uppper and
lower bounds and to verify that the latency has value between
bounds.

The latency-test container image gives possibilty to run the latency
tests without need to install go, ginkgo or other go related modules.

To run the latency test inside of container:

```
docker run --rm -v /home/alukiano/Downloads/cluster/auth/kubeconfig:/kubeconfig -e KUBECONFIG=/kubeconfig -e LATENCY_TEST_RUN=true -e LATENCY_TEST_RUNTIME=60 -e OSLAT_MAXIMUM_LATENCY=700 alukiano/latency-test:4.6-snapshot /usr/bin/run-tests.sh
```

You can run the container with different ENV variable, but the bare minimum is to pass
`KUBECONFIG` mount and ENV variable, to give to the test access to the cluster and
`LATENCY_TEST_RUN=true` to run the latency test.

- `LATENCY_TEST_RUN`: indicates if the latency test should run
- `LATENCY_TEST_RUNTIME`: the amount of time in seconds that the latency test should run
- `LATENCY_TEST_IMAGE`: the image use under the latency test
- `OSLAT_MAXIMUM_LATENCY`: the expected maximum latency for all buckets in us

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,29 @@ Unit tests can be executed with `make unittests`.
 The functional tests are located in `/functests`. They can be executed with `make functests-only` on a cluster with a
 deployed Performance Operator and configured MCP and nodes. It will create its own Performance profile!
 
+### Latency test
+
+The latency-test container image gives the possibility to run the latency 
+test without need to install go, ginkgo or other go related modules.
+
+The test himself is running the `oslat` binary and verifies if the maximal latency returned by the `oslat`
+less than specified value under the `OSLAT_MAXIMUM_LATENCY`.
+
+To run the latency test inside of the container:
+
+```
+docker run --rm -v /kubeconfig:/kubeconfig -e KUBECONFIG=/kubeconfig -e LATENCY_TEST_RUN=true -e LATENCY_TEST_RUNTIME=60 -e OSLAT_MAXIMUM_LATENCY=700 alukiano/latency-test:4.6-snapshot /usr/bin/run-tests.sh
+```
+
+You can run the container with different ENV variables, but the bare minimum is to pass
+`KUBECONFIG` mount and ENV variable, to give to the test access to the cluster and
+`LATENCY_TEST_RUN=true` to run the latency test.
+
+- `LATENCY_TEST_RUN` indicates if the latency test should run.
+- `LATENCY_TEST_RUNTIME` the amount of time in seconds that the latency test should run.
+- `LATENCY_TEST_IMAGE` the image that used under the latency test.
+- `OSLAT_MAXIMUM_LATENCY` the expected maximum latency for all buckets in us.
+
 # Contributing
 
 See [CONTRIBUTING](CONTRIBUTING.md) for some guidelines.
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -1,0 +1,236 @@
+package __latency
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	testutils "github.com/openshift-kni/performance-addon-operators/functests/utils"
+	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/nodes"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/pods"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/profiles"
+	v1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
+	"k8s.io/utils/pointer"
+)
+
+var (
+	latencyTestRun     = false
+	latencyTestRuntime = "300"
+	latencyTestImage   = "quay.io/openshift-kni/oslat:latest"
+	maximumLatency     = -1
+)
+
+// LATENCY_TEST_RUN: indicates if the latency test should run
+// LATENCY_TEST_RUNTIME: the amount of time in seconds that the latency test should run
+// LATENCY_TEST_IMAGE: the image use under the latency test
+// OSLAT_MAXIMUM_LATENCY: the expected maximum latency for all buckets in us
+func init() {
+	latencyTestRunEnv := os.Getenv("LATENCY_TEST_RUN")
+	if latencyTestRunEnv != "" {
+		if latencyTestRunEnv == "true" {
+			latencyTestRun = true
+		}
+	}
+
+	latencyTestRuntimeEnv := os.Getenv("LATENCY_TEST_RUNTIME")
+	if latencyTestRuntimeEnv != "" {
+		latencyTestRuntime = latencyTestRuntimeEnv
+	}
+
+	latencyTestImageEnv := os.Getenv("LATENCY_TEST_IMAGE")
+	if latencyTestImageEnv != "" {
+		latencyTestImage = latencyTestImageEnv
+	}
+
+	maximumLatencyEnv := os.Getenv("OSLAT_MAXIMUM_LATENCY")
+	if maximumLatencyEnv != "" {
+		var err error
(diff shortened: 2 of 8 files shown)


yanirq:
@cynepco3hahue  If you see a scenario where a different performance profile then the one we create in config_0 can cause a failure make sure to either adjust the parameters or skip the test with a warning when DISCOVER_MODE=true.

fedepaol:
Also, bear in mind that this (as it is) would get consumed by CNF-Tests.
Maybe it makes sense to have it as a different (4_latency?) test suite so we can choose when to import it, which is not going to happen as part of the 4.6 release.

fedepaol:
Also also (sorry I came late to the party), would it make sense to do as we do with all the other toolery used by the tests, enhance the image itself?

cynepco3hahue:
> @cynepco3hahue If you see a scenario where a different performance profile than the one we create in config_0 can cause a failure make sure to either adjust the parameters or skip the test with a warning when DISCOVER_MODE=true.

I only interested in the amount of the isolated CPUs under the profile and I am looking for profile via https://github.com/openshift-kni/performance-addon-operators/pull/320/files#diff-ffde53f767f84a78ee1f37ccfe0c9870R65, so it should be ok.

cynepco3hahue:
@fedepaol @yuvalk I updated the PR according to our discussion.

cynepco3hahue:
@fedepaol @yuvalk Can you please review?

cynepco3hahue:
> @cynepco3hahue Can we also add instructions to the docs ?

I will provide the README under the latency test suit, does it fine from your side?

functests/4_latency/latency.go
```diff
@@ -0,0 +1,228 @@
...
+			Expect(maximumLatency).ToNot(BeNil())
+
+			for _, lat := range strings.Split(string(latencies[1]), " ") {
+				if lat == "" {
+					continue
+				}
+
+				curr, err := strconv.Atoi(lat)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(curr < maximumLatency).To(BeTrue())
```
yuvalk, line 152: I'd call it expected latency (not max)
and allow curr to be within 10% larger or 10% smaller than that.
or even have another env that will define.

I think it is important to check for latency 'improvements' because sometime they may mean something is wrong with the test/env/code

cynepco3hahue: I called it this way because it the upper bound, so the test latency should not be above the maximum latency. Regarding 10%, you can set any value for the upper bound when you run the container, so if you want to have bigger expected you can pass 22 instead of 20, but the requirement pretty straight forward say that the latency should be less that 20us.

yuvalk: I'm talking about such a test as a way to make sure we don't deteriorate too much. not just to attest that we are in compliance with required latency (which yes, is currently 20us)

cynepco3hahue: The problem that we do not check the single CPU, we are also checking the number of CPUs and each CPU can have different latency, for example:
```
Maximum:    16537 16 16 15 13 13 13 12 12 12 12 12 12 12 12 12 12 (us)
```
also, I unsure who should investigate this difference, I prefer currently to concentrate on the 20us and once we will have a requirement it will be pretty easy to change the test.

README.md
```diff
@@ -101,6 +101,26 @@ Unit tests can be executed with `make unittests`.
...
 
+### Latency test
+
+The latency-test container image gives the possibility to run the latency 
+test without need to install go, ginkgo or other go related modules.
+
+To run the latency test inside of the container:
+
+```
+docker run --rm -v /home/alukiano/Downloads/cluster/auth/kubeconfig:/kubeconfig -e KUBECONFIG=/kubeconfig -e LATENCY_TEST_RUN=true -e LATENCY_TEST_RUNTIME=60 -e OSLAT_MAXIMUM_LATENCY=700 alukiano/latency-test:4.6-snapshot /usr/bin/run-tests.sh
```
yanirq, line 112: why are we using a personal path here ?

functests/4_latency/latency.go
```diff
@@ -0,0 +1,236 @@
...
+				curr, err := strconv.Atoi(lat)
+				Expect(err).ToNot(HaveOccurred())
+
+				// skip the anomaly value
+				if curr > maximumLatency && !anomaly {
+					anomaly = true
+					continue
+				}
+
+				Expect(curr < maximumLatency).To(BeTrue(), fmt.Sprintf("The current latency %d is bigger than the expected one %d", curr, maximumLatency))
```
yanirq, line 164: equals or bigger then ?  is that crucial ?

functests/4_latency/latency.go
```diff
@@ -0,0 +1,236 @@
...
+			}
+
+			maximumRegex, err := regexp.Compile(`Maximum:\t*\s*(.*)\s*\(us\)`)
+			Expect(err).ToNot(HaveOccurred())
+
+			latencies := maximumRegex.FindSubmatch([]byte(out))
+			Expect(maximumLatency).ToNot(BeNil())
+
+			// under the output of the oslat very often we have one anomaly high value, for example
+			// Maximum:    16543 15 15 14 13 12 12 13 12 12 12 12 12 12 12 12 12 (us)
```
yanirq, line 146: is this a future FIXME ? or skip we need do ?  maybe we need a skip threshold here ?

README.md
cynepco3hahue: done

functests/4_latency/latency.go
cynepco3hahue: I think it does not critical here, IIRC we should have the latency less than 20us.

functests/4_latency/latency.go
cynepco3hahue: Its nothing we can do under our test, it probably some issue with the oslat and kernel, at least we know that the anomaly value appears only once under the output.

cynepco3hahue:
The only failure is
```
• Failure [360.786 seconds]
[rfe_id:28761][performance] Updating parameters in performance profile
/go/src/github.com/openshift-kni/performance-addon-operators/functests/2_performance_update/updating_profile.go:28
  Updating of nodeSelector parameter and node labels
  /go/src/github.com/openshift-kni/performance-addon-operators/functests/2_performance_update/updating_profile.go:201
    [test_id:27484]Verifies that node is reverted to plain worker when the extra labels are removed [It]
    /go/src/github.com/openshift-kni/performance-addon-operators/functests/2_performance_update/updating_profile.go:250
    Expected
      
        <string>: BOOT_IMAGE=(hd0,gpt1)/ostree/rhcos-aa30627385612f9907dff9de78c8f7463433f038fdd1634baad7513804c08901/vmlinuz-4.18.0-211.el8.x86_64 rhcos.root=crypt_rootfs random.trust_cpu=on console=tty0 console=ttyS0,115200n8 rd.luks.options=discard ostree=/ostree/boot.0/rhcos/aa30627385612f9907dff9de78c8f7463433f038fdd1634baad7513804c08901/0 ignition.platform.id=gcp skew_tick=1 nohz=on rcu_nocbs=1-3 tuned.non_isolcpus=00000001 intel_pstate=disable nosoftlockup tsc=nowatchdog intel_iommu=on iommu=pt isolcpus=managed_irq,1-3 systemd.cpu_affinity=0 default_hugepagesz=1G hugepagesz=2M hugepages=128 nmi_watchdog=0 audit=0 mce=off processor.max_cstate=1 idle=poll intel_idle.max_cstate=0 skew_tick=1 nohz=on rcu_nocbs=1-3 tuned.non_isolcpus=00000001 intel_pstate=disable nosoftlockup tsc=nowatchdog intel_iommu=on iommu=pt isolcpus=managed_irq,1-3 systemd.cpu_affinity=0 default_hugepagesz=1G hugepagesz=2M hugepages=128 nmi_watchdog=0 audit=0 mce=off processor.max_cstate=1 idle=poll intel_idle.max_cstate=0 skew_tick=1 nohz=on rcu_nocbs=1-3 tuned.non_isolcpus=00000001 intel_pstate=disable nosoftlockup tsc=nowatchdog intel_iommu=on iommu=pt isolcpus=managed_irq,1-3 systemd.cpu_affinity=0 default_hugepagesz=1G hugepagesz=2M hugepages=128 nmi_watchdog=0 audit=0 mce=off processor.max_cstate=1 idle=poll intel_idle.max_cstate=0
    not to contain substring
        <string>: tuned.non_isolcpus
    /go/src/github.com/openshift-kni/performance-addon-operators/functests/2_performance_update/updating_profile.go:281
------------------------------
[rfe_id:28761][performance] Updating parameters in performance profile Updating of nodeSelector parameter and node labels 
  Reverts back nodeSelector and cleaning up leftovers
  /go/src/github.com/openshift-kni/performance-addon-operators/functests/2_performance_update/updating_profile.go:296
I0817 19:00:01.856186   22295 mcps.go:133] MCP is targeting 1 node(s)
•
JUnit report was created: /tmp/artifacts/unit_report_performance_update.xml
```

PR #321, opened 2020-08-06
Update hugepages doc

cynepco3hahue:
Starting from the OpenShift 4.6 the PAO can configure hugepages with the
different sizes on the node. This functionality should be reflected under
the documentation.


pkg/apis/performance/v1/performanceprofile_types.go
```diff
@@ -16,6 +16,8 @@ type PerformanceProfileSpec struct {
 	// CPU defines a set of CPU related parameters.
 	CPU *CPU `json:"cpu,omitempty"`
 	// HugePages defines a set of huge pages related parameters.
+	// It possible to define huge pages with the different size, for example you can in the same time
```
fromanirh, line 19: It *is* possible?

pkg/apis/performance/v1/performanceprofile_types.go
```diff
@@ -16,6 +16,8 @@ type PerformanceProfileSpec struct {
...
+	// define 1G and 2M and the performance-addon-operator will configure both sizes on the node.
```
fromanirh, line 20: "you can define at the same time"?

pkg/apis/performance/v1/performanceprofile_types.go
yanirq: Suggestion: 
It is possible to set huge pages with multiple size values at the same time. For example, hugepages can be set with 1G and 2M, both values will be set on the node by the performance-addon-operator.

cynepco3hahue: done

pkg/apis/performance/v1/performanceprofile_types.go
cynepco3hahue: done

cynepco3hahue:
@fromanirh @yanirq I update the comment, I also added some important note regarding the default hugepages size.

PR #323, opened 2020-08-11
index: fix building of the index images

fromanirh:
It seems that at least from version "18.02.0-ce"
the `initializer` found in the
`quay.io/operator-framework/upstream-registry-builder` image requires a
more precise path specification to read the manifests.

We fix our scripts to comply.
This issue was not triggered earlier on CI because we don't build *index*
images in CI, in turn caused by current CI limitations; we only build
bundle images in CI.

### DIFF
--- a/hack/generate-manifests-index.sh
+++ b/hack/generate-manifests-index.sh
@@ -8,4 +8,4 @@ RUNTIME=${IMAGE_BUILD_CMD:-podman}
 VERSION=${OPERATOR_VERSION:-4.6.0}
 
 mkdir -p ${BASEPATH}/../build/_output/database || :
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -54,6 +54,10 @@ dist:
 .PHONY: dist-tools
 dist-tools: dist-csv-generator dist-csv-replace-imageref
 
+.PHONY: build-clean
+build-clean:
```
yanirq, line 58: do we want to join both clean operation to one call ?  or trigger is prior to build ?

fromanirh: OK, better to remove this hunk from the patch and reevaluate later.

fromanirh:
e2e-gcp unrelated, overrode to unblock.

PR #325, opened 2020-08-13
Add validation to prevent hugepages duplication

cynepco3hahue:
It can mislead what the exact configuration the node will have
when a user will specify duplication entries for the same size
of hugepages. To avoid it the controller will raise a validation
error.

Examples of problematic configuraitons:
```yaml
pages:
- count: 1
  node: 0
  size: 1G
- count: 2
  node: 0
  size: 1G
```

```yaml
pages:
- count: 1
  size: 1G
- count: 2
  size: 1G
```

### DIFF
--- a/pkg/controller/performanceprofile/components/profile/profile.go
+++ b/pkg/controller/performanceprofile/components/profile/profile.go
@@ -112,6 +112,32 @@ func IsPaused(profile *v1.PerformanceProfile) bool {
 	return false
 }
 
+func validatePageDuplication(page *v1.HugePage, pages []v1.HugePage) error {
+	for _, p := range pages {
+		if page.Size != p.Size {
+			continue
+		}
+
+		if page.Node != nil && p.Node == nil {
+			continue
+		}
+
+		if page.Node == nil && p.Node != nil {
+			continue
+		}
+
+		if page.Node == nil && p.Node == nil {
+			return validationError(fmt.Sprintf("the page with the size %q and without the specified NUMA node, has duplication", page.Size))
+		}
+
+		if *page.Node == *p.Node {
+			return validationError(fmt.Sprintf("the page with the size %q and with specified NUMA node %d, has duplication", page.Size, *page.Node))
+		}
+	}
+
+	return nil
+}
+
 func validateHugepages(hugepages *v1.HugePages) error {
 	// validate that default hugepages size has correct value, currently we support only 2M and 1G(x86_64 architecture)
 	if hugepages.DefaultHugePagesSize != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/performanceprofile/components/profile/profile.go
```diff
@@ -121,10 +147,14 @@ func validateHugepages(hugepages *v1.HugePages) error {
 		}
 	}
 
-	for _, page := range hugepages.Pages {
+	for i, page := range hugepages.Pages {
 		if page.Size != hugepagesSize1G && page.Size != hugepagesSize2M {
 			return validationError(fmt.Sprintf("the page size should be equal to %q or %q", hugepagesSize1G, hugepagesSize2M))
 		}
+
+		if err := validatePageDuplication(&page, hugepages.Pages[i+1:]); err != nil {
```
dshchedr, line 155: Does it cover scenario when hp sizes goes not sequentially:

pages:
```
- count: 1
  node: 0
  size: 1G
- count: 128
  node: 0
  size: 2M
- count: 2
  node: 0
  size: 1G
```

or we don't really care about it?

cynepco3hahue: yes, it should cover all cases, the slice that I send just means that if 1 is duplicate of 3, the 3 also duplicate of 1, so no need to check again the 3 to 1 duplication.

dshchedr: thanks for clarification, I need to work on my golang skills :)

fromanirh: It would be nice to add this case to the unit tests

cynepco3hahue:
> Can you add a line or two to the commit message to explain what happens exactly (error message? which?) if we have a duplication?

@fromanirh We do not have any error message, it just the last block duplication block will override all above duplicates, that can lead to the misconfiguration, I wrote it under the commit message, give me know if you want to add more details to it.

pkg/controller/performanceprofile/components/profile/profile.go
cynepco3hahue: added additional unittest

dshchedr:
looks like another known issue:

E0818 21:20:18.140239    2041 writer.go:135] Marking Degraded due to: failed to execute rpm-ostree ["override" "remove" "kernel" "kernel-core" "kernel-modules" "kernel-modules-extra" "--install" "kernel-rt-core" "--install" "kernel-rt-modules" "--install" "kernel-rt-modules-extra" "--install" "kernel-rt-kvm"] : exit status 1

/retest

pkg/controller/performanceprofile/components/profile/profile_test.go
```diff
@@ -87,6 +87,54 @@ var _ = Describe("PerformanceProfile", func() {
...
+						Node:  pointer.Int32Ptr(0),
+					})
+					profile.Spec.HugePages.Pages = append(profile.Spec.HugePages.Pages, v1.HugePage{
+						Count: 64,
+						Size:  hugepagesSize1G,
+						Node:  pointer.Int32Ptr(0),
+					})
+					err := ValidateParameters(profile)
+					Expect(err).Should(HaveOccurred())
+					Expect(err.Error()).To(ContainSubstring(fmt.Sprintf("the page with the size %q and with specified NUMA node 0, has duplication", hugepagesSize1G)))
```
fromanirh, line 106: we will need to figure out a better way to test than test for string in the error messages in the future.

PR #326, opened 2020-08-17
Add dummy 2M huge pages kernel arguments

cynepco3hahue:
Some architectures such as arm, powerpc, and sparc set up hstates for all supported sizes.
Other architectures such as riscv, s390, and x86 only set up hstates for those requested on the command line with hugepagesz=.
Depending on config options, x86 and riscv may or may not set up PUD_SIZE hstates.
So once you specify the `default_hugepagesz` equal to 1G,
the kernel will remove all 2M huge pages related files and directories from the filesystem.
You can find more details under the kernel email thread - https://lkml.org/lkml/2020/3/5/1113.

And it will prevent to allocate 2M huge pages on the specific NUMA node because the expected file does not exist.

To W/A the problem, once a user will request 2M huge pages allocation on the
specific NUMA node and the `default_hugepagesz` equal to 1G, the controller
will add dummy 2M huge pages kernel arguments `hugepagesz=2M hugepages=0`.

### DIFF
--- a/pkg/controller/performanceprofile/components/tuned/tuned.go
+++ b/pkg/controller/performanceprofile/components/tuned/tuned.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"text/template"
 
+	"k8s.io/utils/pointer"
+
 	performancev1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
 	componentsprofile "github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/profile"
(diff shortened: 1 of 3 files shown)


pkg/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -55,19 +57,45 @@ func NewNodePerformance(assetsDir string, profile *performancev1.PerformanceProf
...
 	if profile.Spec.HugePages != nil {
+		var defaultHugepageSize performancev1.HugePageSize
 		if profile.Spec.HugePages.DefaultHugePagesSize != nil {
-			templateArgs[templateDefaultHugepagesSize] = string(*profile.Spec.HugePages.DefaultHugePagesSize)
+			defaultHugepageSize = *profile.Spec.HugePages.DefaultHugePagesSize
+			templateArgs[templateDefaultHugepagesSize] = string(defaultHugepageSize)
 		}
 
-		hugepages := []string{}
+		var is2MHugepagesRequested *bool
```
dshchedr, line 66: just curious,  may be it's better to add dummy arguments for any hugepages sizes (not only 2M) when numa node is selected?

I don't see any issues with adding dummies for 1G as well, as for me in this case the code logic will be more readable/clearer.

cynepco3hahue: I prefer to avoid adding hack values when we do not really need it, I also have hope that in the near future we will have some work under the kernel that will preserve folders and files for all hugepages always

fromanirh:
holding to let other reviewers chime in

pkg/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -48,47 +66,92 @@ var _ = Describe("Tuned", func() {
...
+
+			profile.Spec.HugePages.Pages[0].Node = pointer.Int32Ptr(1)
+			manifest = getTunedManifest(profile)
 			Expect(strings.Count(manifest, "hugepagesz=")).Should(BeNumerically("==", 1))
 			Expect(strings.Count(manifest, "hugepages=")).Should(BeNumerically("==", 2))
 		})
 
+		Context("with 1G default huge pages", func() {
+			Context("with requested 2M huge pages allocation on the specified node", func() {
+				It("should append the dummy 2M huge pages kernel arguments", func() {
```
dshchedr, line 95: How about verification that dummy args were removed when we get rid of 2M from profile?

cynepco3hahue: will add it

cynepco3hahue: done

cynepco3hahue:
@dshchedr @fromanirh Can you please review again?

PR #330, opened 2020-08-25
Bug 1872417: E2e tests - do not panic in discovery mode with no profile installed.

fedepaol:
If discovery mode is enabled but no profile is installed the tests currently panic.
Here, we are forcing all the tests to skip if that happens, providing the reason.

### DIFF
--- a/functests/1_performance/cpu_management.go
+++ b/functests/1_performance/cpu_management.go
@@ -18,6 +18,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils"
 	testutils "github.com/openshift-kni/performance-addon-operators/functests/utils"
 	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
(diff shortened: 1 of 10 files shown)


yanirq:
@fedepaol  what about the other test cases : 2_peformance_update,  3_peformance_status, 4_latency ? 
if we stop the tests run after 1_pefromance suite then its ok, if not then need to consider the rest of the chain.

fedepaol:
> @fedepaol what about the other test cases : 2_peformance_update, 3_peformance_status, 4_latency ?
> if we stop the tests run after 1_pefromance suite then its ok, if not then need to consider the rest of the chain.

Good point, I completely missed them.

`1_performance` is the only one currently vendored in cnf-tests. I remember discussing (and choosing) not to import `2_performance_update`, the latency one is 4.7 material and we can think about vendoring `3_performance_status`.

In any case, I'll extend this PR to cover them as well as it makes sense (I thought they weren't involved in "discovery mode", but since they use the node selector that may driven by discovery mode, they will work). We can have a separate discussion about vendoring the status suite, maybe in 4.7.

cynepco3hahue:
@fedepaol Do you want to cherry-pick it to 4.5?

fedepaol:
> @fedepaol Do you want to cherry-pick it to 4.5?

Nope, we don't have discovery mode there.

PR #333, opened 2020-08-27
Use centos:8 in Dockerfile.tools

MarSik:
The CI started failing with the following message:

STEP 7: RUN initializer --manifests /registry/performance-addon-operator-catalog --output bundles.db
initializer: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by initializer)

That is caused by too old glibc. Lets see if newer release of the base image will solve it.

### DIFF
--- a/openshift-ci/Dockerfile.tools
+++ b/openshift-ci/Dockerfile.tools
@@ -1,4 +1,4 @@
-FROM centos:7
+FROM centos:8
 
 ENV HOME=/home/ci
 ENV GOROOT /usr/local/go


yanirq:
i think the errors are coming from Dockerfile.registry.upstream.dev + Dockerfile.registry.build  using `FROM quay.io/openshift/origin-operator-registry:4.5` ,  there are some discussions about the variations. @SchSeba is trying to adapt that in https://github.com/openshift-kni/performance-addon-operators/pull/334

yanirq:
@MarSik  you can close this.  issue is fixed under openshift/release

PR #339, opened 2020-09-08
Fix registry images

cynepco3hahue:
The latest changes under the base registry images broke our build because
the base image does not have any binaries except ones that related to
the registry under the `/bin` directories. Also, the default user
of the registry base image changed from the root to 1001, that will prevent
to save `bundles.db` database under the root directory.

### DIFF
--- a/openshift-ci/Dockerfile.registry.build
+++ b/openshift-ci/Dockerfile.registry.build
@@ -1,4 +1,8 @@
-FROM quay.io/openshift/origin-operator-registry:4.5
+FROM quay.io/openshift/origin-operator-registry:4.6 as builder
+
+FROM quay.io/openshift/origin-operator-registry:4.6
+
+COPY --from=builder /usr/bin/sh /bin/sh
 
 ARG OPENSHIFT_BUILD_NAMESPACE
 
@@ -11,8 +15,13 @@ ENV REG_URL=registry.svc.ci.openshift.org
 
 # replaces performance-addon-operator image with the one built by openshift ci
 RUN find /registry/performance-addon-operator-catalog/ -type f -exec sed -i "s|REPLACE_IMAGE|${REG_URL}/${OPENSHIFT_BUILD_NAMESPACE}/stable:performance-addon-operator|g" {} \; || :
+
 # Initialize the database
+USER root
 RUN initializer --manifests /registry/performance-addon-operator-catalog --output bundles.db
+RUN chown 1001 bundles.db
+
+USER 1001
 
 # There are multiple binaries in the origin-operator-registry
 # We want the registry-server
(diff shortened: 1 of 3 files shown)


openshift-ci/Dockerfile.registry.build
```diff
@@ -11,8 +15,13 @@ ENV REG_URL=registry.svc.ci.openshift.org
 
 # replaces performance-addon-operator image with the one built by openshift ci
 RUN find /registry/performance-addon-operator-catalog/ -type f -exec sed -i "s|REPLACE_IMAGE|${REG_URL}/${OPENSHIFT_BUILD_NAMESPACE}/stable:performance-addon-operator|g" {} \; || :
+
 # Initialize the database
+USER root
```
fromanirh, line 20: why is this needed now?

openshift-ci/Dockerfile.registry.upstream.dev
```diff
@@ -9,7 +13,11 @@ COPY ${OLM_SOURCE} /registry/performance-addon-operator-catalog
 RUN find /registry/performance-addon-operator-catalog/ -type f -exec sed -i "s|REPLACE_IMAGE|${FULL_OPERATOR_IMAGE}|g" {} \; || :
```
fromanirh, line 13: another option is getting rid of this entirely, possibly enhancing our `csv-replace-imageref` tool and/or packing in a different image

openshift-ci/Dockerfile.registry.build
cynepco3hahue: Because the base image now runs with USER 1001, but we trying to put the `bundle.db` under the `/` and the user 1001 not allowed to write under it.

openshift-ci/Dockerfile.registry.upstream.dev
cynepco3hahue: The problem, not just a find or sed binaries, but the call to the `RUN` at all, because under the neat `RUN` calls to `/bin/sh -c`.

But to remove this line can make our Dockerfile cleaner, so I definitely not versus it, but it should be part of the different PR :)

openshift-ci/Dockerfile.registry.build
fromanirh: ok, so another change caused bu the base image changes :\

openshift-ci/Dockerfile.registry.upstream.dev
fromanirh: ok, but if so every `RUN` instance will fail, not just this one, right?

cynepco3hahue: Yes, but to avoid it I just copy `sh` from /usr/bin to /bin

cynepco3hahue:
Fixed by https://github.com/operator-framework/operator-registry/pull/437

PR #340, opened 2020-09-08
Enable support for stall daemon in base profile

yanirq:
Enable by default the stall daemon in the base profile.
Stall daemon is supported by NTO : https://github.com/openshift/cluster-node-tuning-operator/pull/152  
and tuned service plugin: https://github.com/redhat-performance/tuned/pull/293

### DIFF
--- a/build/assets/tuned/openshift-node-performance
+++ b/build/assets/tuned/openshift-node-performance
@@ -25,6 +25,9 @@ governor=performance                          #  latency-performance
 energy_perf_bias=performance                  #  latency-performance 
 min_perf_pct=100                              #  latency-performance 
 
+[service]
+service.stalld=start,enable
+
 [vm]
 transparent_hugepages=never                   #  network-latency
 
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -151,6 +151,15 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				Expect(string(initrd)).ShouldNot(ContainSubstring("'/etc/systemd/system.conf /etc/systemd/system.conf.d/setAffinity.conf'"))
 			}
 		})
+
+		It("[test_id:][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] stalld daemon is running on the host", func() {
+			for _, node := range workerRTNodes {
+				tuned := tunedForNode(&node)
+				_, err := pods.ExecCommandOnPod(tuned, []string{"pidof", "stalld"})
+				Expect(err).ToNot(HaveOccurred())
+			}
+		})
+
 	})
 
 	Context("Additional kernel arguments added from perfomance profile", func() {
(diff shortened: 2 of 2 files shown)


build/assets/tuned/openshift-node-performance
```diff
@@ -25,6 +25,9 @@ governor=performance                          #  latency-performance
 energy_perf_bias=performance                  #  latency-performance 
 min_perf_pct=100                              #  latency-performance 
 
+[service]
+service.stalld=start,enable
```
MarSik, line 29: Is it going to work fine with the default arguments? Or should we pass in a config file with env vars?

yanirq: @MarSik  default arguments are set via NTO here :  https://github.com/openshift/cluster-node-tuning-operator/pull/152/files#diff-98484aa29ef1977106bf3b4930e1199a

jmencak: I guess Martin is asking whether there should be a way to specify an environment file to override the default arguments.  You can do this by something like:

```
      [service]
      service.stalld=start,enable,file:/host/etc/tuned/openshift-realtime/stalld.conf
```

However, you'd need to implement a mechanism to place the environment file on the host.  And this is probably what the question is about I'm guessing?

yanirq: If the arguments set in NTO should be the default for performance then the answer should be yes.
We should strive to reduce using machine configs on PAO to keep restarts at minimum. 
But we should decide whether we want to enable this kind of configuration on PAO side

jmencak:
CI is good at telling
/joke

cynepco3hahue:
@yanirq Can you please add some small e2e test, that will verify that by default the service truly runs on the node?

functests/1_performance/performance.go
```diff
@@ -151,6 +151,15 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				Expect(string(initrd)).ShouldNot(ContainSubstring("'/etc/systemd/system.conf /etc/systemd/system.conf.d/setAffinity.conf'"))
 			}
 		})
+
+		It("[test_id:][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] stalld daemon is running on the host", func() {
```
yanirq, line 155: @gsr-shanks  @dshchedr  Can I get a test ID to add here ?   this is checking that stalld service is running on a tuned daemon for woker-cnf nodes

gsr-shanks: @yanirq test_id: 35363 please

build/assets/tuned/openshift-node-performance
MarSik: In fact I was just asking if the defaults are good enough :) And I think they are.

MarSik (approved):

Issue #342, opened 2020-09-09
bug: cannot run `make` twice in a row


fromanirh:
**Version information**
git master snapshot 347b6e8

**Description of the problem**
If we try to run `make` from a clean source snapshot, it works as expected:
```bash
$ make
Running gofmt
gofmt -s -w `find . -path ./vendor -prune -o -type f -name '*.go' -print`
Running go lint
hack/lint.sh
/usr/bin/golint
Running go vet
go vet ./...
mkdir -p build/_output/bin || :
Building operator binary
mkdir -p build/_output/bin; \
    LDFLAGS="-s -w "; \
    LDFLAGS+="-X github.com/openshift-kni/performance-addon-operators/version.Version=${CI_UPSTREAM_VERSION:-$(git describe --always --tags)} "; \
    LDFLAGS+="-X github.com/openshift-kni/performance-addon-operators/version.GitCommit=${CI_UPSTREAM_COMMIT:-$(git rev-list -1 HEAD)} "; \
    LDFLAGS+="-X github.com/openshift-kni/performance-addon-operators/version.BuildDate=$(date --utc -Iseconds) "; \
env GOOS=linux GOARCH=amd64 go build -i -ldflags="$LDFLAGS" \
  -mod=vendor -o build/_output/bin/performance-addon-operators ./cmd/manager
Using operator-sdk cached at _cache/tools/operator-sdk-v0.18.2-x86_64-linux-gnu
Building csv-generator tool
Generating developer csv

OPERATOR_SDK="""_cache"/tools"/"operator-sdk-"v0.18.2"-"x86_64-linux-gnu""" FULL_OPERATOR_IMAGE="REPLACE_IMAGE" hack/csv-generate.sh -dev
INFO[0000] Generating CSV manifest version 4.6.0        
WARN[0025] ClusterServiceVersion validation: [OperationFailed] provided API should have an example annotation 
WARN[0025] Required csv fields not filled in file performance-addon-operator.v4.6.0.clusterserviceversion.yaml:
	spec.maintainers 
INFO[0025] CSV manifest generated successfully          
CSV written to build/_output/olm-catalog/performance-addon-operator/4.6.0/performance-addon-operator.v4.6.0.clusterserviceversion.yaml
INFO[0000] Running CRD generator.                       
INFO[0000] CRD generation complete.                     
New OLM manifests created at build/_output/olm-catalog
[...]

PR #338, opened 2020-09-07
Fix generate scripts to support generation of z-stream manifests

cynepco3hahue:
1. csv-generate: pass `--from-version` parameter to prevent errors
from the `operator-sdk generate csv` and create the separate ENV
variable for the CSV channel. Also, copy CRD and package yamls to the
repository directory after the generation to have updated files.

2. generate-manifest-index: remove the database folder before generating
the new bundle database.

3. generate-manifests-tree: fix the path for generated files.

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

PR #343, opened 2020-09-09
Makefile: hack: let repeated `make` calls works

fromanirh:
Fix `generate-manifests-tree.sh` to clean up the output directory before
to run. This makes sense in general because the scripts can't deal with
dirty output directory: it expects implicitely a clean slate, so let's
enforce this constraints.

As side effect, we enable repeated `make` calls to work cleanly now.

Resolves: https://github.com/openshift-kni/performance-addon-operators/issues/342

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,10 @@ export GO111MODULE=on
 .PHONY: build
 build: gofmt golint govet dist generate-manifests-tree
 
+.PHONY: clean
+clean:
+	rm -rf build/_output
+
 .PHONY: dist
 dist: build-output-dir
 	@echo "Building operator binary"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -42,6 +42,10 @@ export GO111MODULE=on
 .PHONY: build
 build: gofmt golint govet dist generate-manifests-tree
 
+.PHONY: clean
+clean:
+	rm -rf build/_output
```
marcel-apf, line 47: that's a nice and needed touch, thanks!

fromanirh:
the core issue was actually fixed by https://github.com/openshift-kni/performance-addon-operators/pull/338

PR #345, opened 2020-09-14
e2e tests: fix the RT kernel check

cynepco3hahue:
Flags for the RT kernel configuration are different under
`rhel-rt-8.2` and `rhel-rt-8.3`, `CONFIG_PREEMPT_RT_FULL` vs
`CONFIG_PREEMPT_RT`. To avoid need to change the flag check
we have better check if the RT kernel is enabled via
`/sys/kernel/realtime` file.

### DIFF
--- a/functests/utils/nodes/nodes.go
+++ b/functests/utils/nodes/nodes.go
@@ -172,9 +172,14 @@ func HasPreemptRTKernel(node *corev1.Node) error {
 		return err
 	}
 
-	cmd = []string{"/bin/bash", "-c", "grep -Ee '^CONFIG_PREEMPT_RT=y' /rootfs/usr/lib/modules/$(uname -r)/config"}
-	if _, err := ExecCommandOnNode(cmd, node); err != nil {
+	cmd = []string{"/bin/bash", "-c", "cat /rootfs/sys/kernel/realtime"}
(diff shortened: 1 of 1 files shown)


MarSik (approved):

cynepco3hahue:
We have some stable failures:
```
[Fail] [rfe_id:28761][performance] Updating parameters in performance profile Verify that all performance profile parameters can be updated Verify that profile parameters were updated [It] verify that the right number of hugepages 2M is available on the system 
/go/src/github.com/openshift-kni/performance-addon-operators/functests/2_performance_update/updating_profile.go:129
[Fail] [rfe_id:28761][performance] Updating parameters in performance profile Verify that all performance profile parameters can be updated [It] [test_id:28612]Verify that Kernel arguments can me updated (added, removed) thru performance profile 
/go/src/github.com/openshift-kni/performance-addon-operators/functests/2_performance_update/updating_profile.go:165
[Fail] [rfe_id:28761][performance] Updating parameters in performance profile Updating of nodeSelector parameter and node labels [It] [test_id:27484]Verifies that node is reverted to plain worker when the extra labels are removed 
/go/src/github.com/openshift-kni/performance-addon-operators/functests/2_performance_update/updating_profile.go:288
```

PR #350, opened 2020-09-21
Remove bitbucket.org/ww/goautoneg dependency

cynepco3hahue:
Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

PR #351, opened 2020-09-21
Force PAO installation on master nodes 

marcel-apf:
PAO is meant to run on "control plane".

Use podAffinity to schedule PAO only on master nodes.
(https://docs.openshift.com/container-platform/4.5/nodes/scheduling/nodes-scheduler-pod-affinity.html)
Use taints tolerations to allow PAO scheduling on the master nodes.
(https://docs.openshift.com/container-platform/4.5/nodes/scheduling/nodes-scheduler-taints-tolerations.html)

Add a test verifying PAO runs in master nodes
Straight forward test that queries the node PAO is running on
is a master node.

### DIFF
--- a/deploy/olm-catalog/performance-addon-operator/4.6.0/performance-addon-operator.v4.6.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/performance-addon-operator/4.6.0/performance-addon-operator.v4.6.0.clusterserviceversion.yaml
@@ -118,6 +118,13 @@ spec:
               labels:
                 name: performance-operator
             spec:
+              affinity:
+                nodeAffinity:
+                  requiredDuringSchedulingIgnoredDuringExecution:
+                    nodeSelectorTerms:
+                    - matchExpressions:
+                      - key: node-role.kubernetes.io/master
+                        operator: Exists
               containers:
               - command:
                 - performance-operator
@@ -137,6 +144,9 @@ spec:
                 name: performance-operator
                 resources: {}
               serviceAccountName: performance-operator
+              tolerations:
+              - effect: NoSchedule
+                key: node-role.kubernetes.io/master
       permissions:
       - rules:
         - apiGroups:
--- a/functests/0_config/config.go
+++ b/functests/0_config/config.go
@@ -5,10 +5,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -27,6 +29,8 @@ import (
 	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/discovery"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/mcps"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/nodes"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/pods"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/profiles"
 	"github.com/openshift-kni/performance-addon-operators/pkg/apis"
 	performancev1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
@@ -36,6 +40,25 @@ import (
 )
 
 var _ = Describe("[performance][config] Performance configuration", func() {
+
+	It("Should run performance profile pod on a master node", func() {
+		pod, err := pods.GetPerformanceOperatorPod()
+		Expect(err).ToNot(HaveOccurred(), "Failed to find the Performance Addon Operator pod")
+
+		Expect(strings.HasPrefix(pod.Name, "performance-operator")).To(BeTrue(),
+			"Performance Addon Operator pod name should start with performance-operator prefix")
+
+		masterNodes, err := nodes.GetByRole(testutils.RoleMaster)
+		Expect(err).ToNot(HaveOccurred(), "Failed to query the master nodes")
+		for _, node := range masterNodes {
+			if node.Name == pod.Spec.NodeName {
+				return
+			}
+		}
+
+		// Fail
+		Expect(true).To(Reject(), "Performance Addon Operator is not running in a master node")
+	})
 
 	It("Should successfully deploy the performance profile", func() {
 
--- a/functests/utils/pods/pods.go
+++ b/functests/utils/pods/pods.go
@@ -11,9 +11,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	testutils "github.com/openshift-kni/performance-addon-operators/functests/utils"
 	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
@@ -142,4 +144,24 @@ func GetContainerIDByName(pod *corev1.Pod, containerName string) (string, error)
 		}
 	}
 	return "", fmt.Errorf("failed to find the container ID for the container %q under the pod %q", containerName, pod.Name)
+}
+
+// GetPerformanceOperatorPod returns the pod running the Performance Profile Operator
+func GetPerformanceOperatorPod() (*corev1.Pod, error) {
+	selector, err := labels.Parse(fmt.Sprintf("%s=%s", "name", "performance-operator"))
+	if err != nil {
+		return nil, err
+	}
+
+	pods := &corev1.PodList{}
+
+	opts := &client.ListOptions{LabelSelector: selector, Namespace: testutils.PerformanceOperatorNamespace}
+	if err := testclient.Client.List(context.TODO(), pods, opts); err != nil {
+		return nil, err
+	}
+	if len(pods.Items) != 1 {
+		return nil, fmt.Errorf("incorrect performance operator pods count: %d", len(pods.Items))
+	}
+
+	return &pods.Items[0], nil
 }
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -13,6 +13,16 @@ spec:
         name: performance-operator
     spec:
       serviceAccountName: performance-operator
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-role.kubernetes.io/master
+                operator: Exists
+      tolerations:
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
       containers:
         - name: performance-operator
           # Replace this with the built image name
--- a/functests/utils/consts.go
+++ b/functests/utils/consts.go
@@ -67,6 +67,8 @@ func init() {
 const (
 	// RoleWorker contains the worker role
 	RoleWorker = "worker"
+	// RoleMaster contains the master role
+	RoleMaster = "master"
 )
 
 const (


marcel-apf:
Depends on https://github.com/openshift-kni/performance-addon-operators/pull/350

functests/utils/pods/pods.go
```diff
@@ -143,3 +145,23 @@ func GetContainerIDByName(pod *corev1.Pod, containerName string) (string, error)
...
+		return nil, err
+	}
+
+	pods := &corev1.PodList{}
+
+	opts := &client.ListOptions{LabelSelector: selector, Namespace: testutils.PerformanceOperatorNamespace}
+	if err := testclient.Client.List(context.TODO(), pods, opts); err != nil {
+		return nil, err
+	}
+	if len(pods.Items) != 1 {
```
MarSik, line 162: I wonder if we should just use len(pods.Items) < 1 (just in case we eventually add HA and leader election). Or maybe it would be premature to change the test now, so just asking.

functests/0_config/config.go
```diff
@@ -37,6 +40,22 @@ import (
...
+		masterNodes, err := nodes.GetByRole(testutils.RoleMaster)
+		Expect(err).ToNot(HaveOccurred(), "Failed to query the master nodes")
+		for _, node := range masterNodes {
+			if node.Name == pod.Spec.NodeName {
+				return
+			}
+		}
+
+		// Fail
+		Expect(true).To(Reject(), "PAO is not running in a master node")
```
yanirq, line 56: change PAO to full name or just "the operator"

functests/utils/pods/pods.go
fromanirh: Fair point, I prefer to keep the test as-is and see it break first when/if we add the HA/leader election.

functests/0_config/config.go
```diff
@@ -37,6 +40,22 @@ import (
 
 var _ = Describe("[performance][config] Performance configuration", func() {
 
+	It("Should run performance profile pod on a master node", func() {
+		pod, err := pods.GetPerformanceProfilePod()
+		Expect(err).ToNot(HaveOccurred(), "Failed to find the PAO pod")
+
+		masterNodes, err := nodes.GetByRole(testutils.RoleMaster)
+		Expect(err).ToNot(HaveOccurred(), "Failed to query the master nodes")
+		for _, node := range masterNodes {
```
yanirq, line 49: alternatively you can get the node by the name and check its role (both ways work)

deploy/olm-catalog/performance-addon-operator/4.6.0/performance-addon-operator.v4.6.0.clusterserviceversion.yaml
```diff
@@ -118,6 +118,13 @@ spec:
               labels:
                 name: performance-operator
             spec:
+              affinity:
```
fromanirh, line 121: it seems OCP control plane is using
```yaml
    nodeSelector:
      node-role.kubernetes.io/master: ""
```
but the end result is the same and affinity rules are more powerful than nodeSelector, so it seems is fine.

functests/0_config/config.go
fromanirh: Yeah, I'd do like @yanirq wrote, but both works so I don't really mind.

functests/0_config/config.go
fromanirh: +1

functests/0_config/config.go
```diff
@@ -37,6 +40,22 @@ import (
 
 var _ = Describe("[performance][config] Performance configuration", func() {
 
+	It("Should run performance profile pod on a master node", func() {
+		pod, err := pods.GetPerformanceProfilePod()
+		Expect(err).ToNot(HaveOccurred(), "Failed to find the PAO pod")
```
fromanirh, line 45: same comment as per line 56

functests/utils/pods/pods.go
```diff
@@ -143,3 +145,23 @@ func GetContainerIDByName(pod *corev1.Pod, containerName string) (string, error)
 	}
 	return "", fmt.Errorf("failed to find the container ID for the container %q under the pod %q", containerName, pod.Name)
 }
+
+// GetPerformanceProfilePod returns the pod running the Performance Profile Operator
+func GetPerformanceProfilePod() (*corev1.Pod, error)  {
```
fromanirh, line 150: If we are looking for the performance-operator pod, why GetPerformance*Profile*Pod? I'd call it
`GetPerformanceOperatorPod`

functests/utils/pods/pods.go
```diff
@@ -143,3 +145,23 @@ func GetContainerIDByName(pod *corev1.Pod, containerName string) (string, error)
...
+	pods := &corev1.PodList{}
+
+	opts := &client.ListOptions{LabelSelector: selector, Namespace: testutils.PerformanceOperatorNamespace}
+	if err := testclient.Client.List(context.TODO(), pods, opts); err != nil {
+		return nil, err
+	}
+	if len(pods.Items) != 1 {
+		return nil, fmt.Errorf("incorrect performance operator pods count: %d", len(pods.Items))
+	}
+
```
fromanirh, line 165: any way to check that pod is indeeded running `performance-operator` and not any random pod which is happen to have the same label?

fromanirh (commented):
looks mostly OK. Most (relatively minor) concerns about the e2e test and the `GetPerformanceProfilePod` function, changes to CSV looks good

functests/utils/pods/pods.go
marcel-apf: Will do, thanks

functests/utils/pods/pods.go
marcel-apf: In the same namespace as PAO  "openshift-performance-addon" ? I think is a relatively safe bet...

deploy/olm-catalog/performance-addon-operator/4.6.0/performance-addon-operator.v4.6.0.clusterserviceversion.yaml
marcel-apf: I preferred affinity since is is reflected in the official docs

functests/0_config/config.go
marcel-apf: Will do

functests/utils/pods/pods.go
marcel-apf: Done

functests/0_config/config.go
marcel-apf: I am missing something, the code does queries the node "name" from the pod, iterates over the master nodes to see if is one of the master nodes.
The other way would be to go over all nodes and find the node with the right name, then query its labels to see is if master.

The above seems more concise/straight forward... but I can rewrite is you see an issue

functests/utils/pods/pods.go
fromanirh: It's still a bet, meaning is still anecdotal evidence, and we need quite some of it to gain confidence. Checking the image name seems stronger (does it contain 'performance-operator'). Feel free to add more checks, or to lookup better ones.
The best way would probably be to check a PAO endpoint and to verify it is behaving correctly; that's sufficient proof the service is there. It's probably complex, maybe too complex to be done here, though.

functests/0_config/config.go
marcel-apf: Done

functests/0_config/config.go
marcel-apf: Done

deploy/olm-catalog/performance-addon-operator/4.6.0/performance-addon-operator.v4.6.0.clusterserviceversion.yaml
fromanirh: Is affinity reflected in the openshift official docs or the kubernetes official docs? or both?

functests/0_config/config.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
```
fromanirh, line 12: unused?

functests/0_config/config.go
fromanirh: The alternative is to get node by name (using server-side filtering ofc), getting the node from pod Spec, and finally checking the node you got has the master labels (thus, it's a master node). Both approaches work, and I don't mind, because masters are supposed to be few in number anyway.

functests/utils/pods/pods.go
marcel-apf: We are also checking we have only one pod in the namespace. The check is the first step in the func tests, I suppose the next steps will fail if for some reason the PAO is not in the namespace, but another pod with same label.
While I agree that theoretically is not enough, is it the place to add more checks?
I can add a check the actual pod name starts with performace-operator-{something}, but this is a side effect on how k8s names pods.Or maybe would add to confidence?

functests/0_config/config.go
marcel-apf: used by Reject()

functests/utils/pods/pods.go
marcel-apf: Added check the performance operator pod name (not label) starts with "performance-operator"

deploy/olm-catalog/performance-addon-operator/4.6.0/performance-addon-operator.v4.6.0.clusterserviceversion.yaml
marcel-apf: both

functests/0_config/config.go
marcel-apf: got it, thanks

functests/utils/pods/pods.go
marcel-apf: That should increase the confidence in the bet

fromanirh (approved):
/lgtm
good enough for now

functests/0_config/config.go
fromanirh: Fair enough. And BTW this is exacty why I dislike the dot imports.

functests/utils/pods/pods.go
fromanirh: Still not completely happy with this solution, but good enough for now.

cynepco3hahue:
@marcel-apf Please fix CI job
```
export GOROOT=$(go env GOROOT); """_cache"/tools"/"operator-sdk-"v0.18.2"-"x86_64-linux-gnu""" generate k8s
time="2020-09-22T16:10:05Z" level=info msg="Running deepcopy code-generation for Custom Resource group versions: [performance:[v1 v1alpha1], ]\n"
time="2020-09-22T16:10:47Z" level=info msg="Code-generation complete."
hack/verify-generated.sh
 M functests/utils/pods/pods.go
```

marcel-apf:
> @marcel-apf Please fix CI job
> 
> ```
> export GOROOT=$(go env GOROOT); """_cache"/tools"/"operator-sdk-"v0.18.2"-"x86_64-linux-gnu""" generate k8s
> time="2020-09-22T16:10:05Z" level=info msg="Running deepcopy code-generation for Custom Resource group versions: [performance:[v1 v1alpha1], ]\n"
> time="2020-09-22T16:10:47Z" level=info msg="Code-generation complete."
> Verifying that all code is committed after updating deps and formatting and generating code
> hack/verify-generated.sh
> uncommitted generated files. run 'make generate' and commit results.
>  M functests/utils/pods/pods.go
> ```

Done, thanks!

fromanirh (approved):
/lgtm

fromanirh:
known issue about dup kernel args

cynepco3hahue:
The single flakiness because of 
```
out=Unable to connect to the server: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

fromanirh:
> The single flakiness because of
> 
> ```
> out=Unable to connect to the server: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
> ```

ok, for the record I agree with the evalutation and I was about to override myself. @cynepco3hahue was just faster.

cynepco3hahue:
@marcel-apf The failure connects to the functional test under the PR
```
• Failure [0.012 seconds]
[performance][config] Performance configuration
/go/src/github.com/openshift-kni/performance-addon-operators/functests/0_config/config.go:42
  Should run performance profile pod on a master node [It]
  /go/src/github.com/openshift-kni/performance-addon-operators/functests/0_config/config.go:44
  Failed to find the Performance Addon Operator pod
      
  Unexpected error:
      <*errors.errorString | 0xc00021e7a0>: {
          s: "incorrect performance operator pods count: 0",
      }
      incorrect performance operator pods count: 0
  occurred
  /go/src/github.com/openshift-kni/performance-addon-operators/functests/0_config/config.go:46
```

cynepco3hahue:
The new test runs before the deployment of the PAO, so it not surprising that the test fails. I propose to make the check where the PAO pod run as part of the deployment test.

fromanirh:
> The new test runs before the deployment of the PAO, so it not surprising that the test fails. I propose to make the check where the PAO pod run as part of the deployment test.

This means moving the test from the `0_config` to the `1_performance` suite, right? Because this is what I was thinking, and it's fine for me.

fromanirh:
> > The new test runs before the deployment of the PAO, so it not surprising that the test fails. I propose to make the check where the PAO pod run as part of the deployment test.
> 
> This means moving the test from the `0_config` to the `1_performance` suite, right? Because this is what I was thinking, and it's fine for me.

let's see: https://github.com/openshift-kni/performance-addon-operators/pull/373/commits/c61e2d120408b39c51b9d8b1796494095867abde

I'll take care of this PR and have it merged ASAP.

fromanirh:
actually obsoleted because #373 got merged

marcel-apf:
yay!, closing this one

PR #362, opened 2020-09-23
Fix for CPU Affinity mask test

dshchedr:
Cpu affinity test may fail in scenario when we have some CPUs that not a part
of Reserved or Isolated groups. After changes in this PR we will be verifying
that rcu processes are not running on isolated cores.


cynepco3hahue:
oh all tests passed, it's a miracle
/approve

cynepco3hahue:
The single tests that fails is
```
[Fail] [rfe_id:28761][performance] Updating parameters in performance profile Updating of nodeSelector parameter and node labels [It] [test_id:27484]Verifies that node is reverted to plain worker when the extra labels are removed 
```

PR #331, opened 2020-08-25
Added test that verifies rcu_nocbs kernel argument on the node

dshchedr:
Also added some fixes for cpu affinity test

PR #371, opened 2020-09-24
[release-4.4] Fix for cpu affinity test

dshchedr:
Changed the pid that verifies cpu affinity mask (now using pid 1) for having more generic approach
Removed colon for reservedCPUSet variable in BeforeEach() (otherwise this variable is Null in tests)

(technically, same changes as we had in commit: https://github.com/openshift-kni/performance-addon-operators/pull/331/commits/5f545eb31f1e18c5163851c0211b455d44ea3d1d, can't do cherry pick because of different functest folder structure)

### DIFF
--- a/functests/performance/cpu_management.go
+++ b/functests/performance/cpu_management.go
@@ -57,7 +57,7 @@ var _ = Describe("[performance] CPU Management", func() {
 
 		Expect(profile.Spec.CPU.Reserved).NotTo(BeNil())
 		reservedCPU = string(*profile.Spec.CPU.Reserved)
-		reservedCPUSet, err := cpuset.Parse(reservedCPU)
+		reservedCPUSet, err = cpuset.Parse(reservedCPU)
 		Expect(err).ToNot(HaveOccurred())
 		listReservedCPU = reservedCPUSet.ToSlice()
 	})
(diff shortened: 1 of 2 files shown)


cynepco3hahue:
The job failed because of the registry
@dshchedr Why do we need this test to be available under release-4.4?

dshchedr:
> The job failed because of the registry
> @dshchedr Why do we need this test to be available under release-4.4?

We still support 4.4 releases and still running tests in d/s. This test is failed on every run (there is no rcu_sched process on node).
Moreover, I found that in our github u/s 4.4 repo this test is also failed:

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-kni_performance-addon-operators/347/pull-ci-openshift-kni-performance-addon-operators-release-4.4-e2e-gcp/1305919938321453056/artifacts/e2e-gcp/unit_report.xml

We didn't catch it because our test runner script has a flaw: we have 2 set of tests and in our current approach we don't catch failed tests from first set (I mean if some of tests from the first set failed but all tests from second set passed - e2e will be marked green)

cynepco3hahue:
@dshchedr To pass the CI, please change the registry to build02 under the PR.

PR #373, opened 2020-09-25
Pao run master nodes

fromanirh:
Followup of https://github.com/openshift-kni/performance-addon-operators/pull/351 with the functest moved out from the `0_config` suite.


fromanirh:
ok, that went way too smoother than expected :\

PR #375, opened 2020-09-28
functests: ensure deleted objects are really gone

fromanirh:
In our test suite we must ensure deleted objects are really gone.
Add WaitForDeletion helper for profiles, and use it in the test which previously just used other means to check a profile was gone.
Make sure we consistently wait for pod deletion once we delete a pod.


fedepaol:
Since we are adding this to ensure that the `CPU Management` suite is not fetching the wrong profile, would you mind adding a printf here https://github.com/openshift-kni/performance-addon-operators/blob/7cdd2ce1a3af05564d6382067b6d01f36c1c09a0/functests/1_performance/cpu_management.go#L54 to dump the name of the profile that is being picked?

Ideally, in a separate commit.

PR #377, opened 2020-09-28
Fix test: "Verify rcu_nocbs kernel argument on the node"

fedepaol:
A given process may end up being assigned in a cpu not belonging to the
reserved cpu set (i.e. if the host has more than reserved + isolated cpus). This
leads to flakes. Here we just check that cpu does not belong to the isolated
cpus because there is no guarantee for belonging to the reserved cpus.

### DIFF
--- a/functests/1_performance/cpu_management.go
+++ b/functests/1_performance/cpu_management.go
@@ -37,7 +37,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 	var balanceIsolated bool
 	var reservedCPU, isolatedCPU string
 	var listReservedCPU, listIsolatedCPU []int
-	var reservedCPUSet, isolatedCPUSet cpuset.CPUSet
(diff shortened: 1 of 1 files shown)


fedepaol:
/assign @fromanirh 
As discussed offline

fromanirh:
the issue here is that the expectations is that reserved_set union isolated_set equals all_cores for each worker node.
If the profile is created as such, the observed flake doesn't exist.
However, we cannot guarantee this is correct because the profile is created without knowing how many cores do the worker node have, and also how the reserved/isolated split should look like. At this moment, these are left to the cluster admin to be set.
We can argue this test should fail if indeed the cluster is misconfigured, thus the aformentioned expectation is not met.
However, the test will fail with an obscure and hard to debug message.
If we want to test this expectation, we should add a separate explicit test. Even better, we should add some explicit validation.

fedepaol:
/close
Already fixed, all it needs is being vendored

PR #378, opened 2020-09-29
controller: validation: check for cpu cores split

fromanirh:
We always assume, and document, that the reserved and
isolated cpus set must not overlap.
Add some validation code to enforce this constraint.

### DIFF
--- a/functests/3_performance_status/status.go
+++ b/functests/3_performance_status/status.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -17,12 +18,14 @@ import (
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/mcps"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/nodes"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/profiles"
+	performancev1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
 
 	corev1 "k8s.io/api/core/v1"
 	nodev1beta1 "k8s.io/api/node/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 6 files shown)


fromanirh:
/hold
still WIP
missing e2e tests and better commit message

cynepco3hahue:
@fromanirh IIUC it only will prevent the overlaps?(that totally fine for me, because our d/s CI does not include all CPUs under isolated and reserved)

fromanirh:
> @fromanirh IIUC it only will prevent the overlaps?(that totally fine for me, because our d/s CI does not include all CPUs under isolated and reserved)

Yep, only overlaps. This is because other checks are hard/impossible unless we know exactly the worker node HW configuration (how many cores?) and, to some extent, without secondguessing the admin perspective.

TL;DR: overlapping cores is the only wrong setting we can detect reliably.

cynepco3hahue:
@fromanirh Do you need to add something else under the PR?

fromanirh:
> @fromanirh Do you need to add something else under the PR?

yep. e2e test(s).

fromanirh:
/hold cancel
/cc @fedepaol 
Added e2e test and @fedepaol as reviewer because I want to be nice with the discovery mode feature

functests/3_performance_status/status.go
```diff
@@ -133,4 +135,52 @@ var _ = Describe("Status testing of performance profile", func() {
...
+						Enabled: pointer.BoolPtr(true),
+					},
+					NUMA: &performancev1.NUMA{
+						TopologyPolicy: pointer.StringPtr("restricted"),
+					},
+				},
+			}
+			Expect(testclient.Client.Create(context.TODO(), overlappingProfile)).ToNot(HaveOccurred())
+			defer func() {
+				err := testclient.Client.Delete(context.TODO(), overlappingProfile)
```
fedepaol, line 175: You should check that the profile is really deleted & that MCP is not updating, otherwise there's a risk that the profile will leak. Check the profile2 test for inspiration

functests/3_performance_status/status.go
```diff
@@ -133,4 +135,52 @@ var _ = Describe("Status testing of performance profile", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+			}()
+
+			nodeLabels := map[string]string{
+				newLabel: "",
+			}
+			condMessage := profiles.GetConditionMessage(nodeLabels, v1.ConditionDegraded)
```
fedepaol, line 182: Should this be wrapped by an eventually?

fedepaol (commented):
A couple of comments

functests/3_performance_status/status.go
fromanirh: Ack. Do we have a way to recover from a leaked profile?

functests/3_performance_status/status.go
fromanirh: Probably better, yes.

functests/3_performance_status/status.go
fedepaol: Not sure I understand the question, just make sure that when the test exits the profile is really deleted (as we do with the other profile2 test)

fromanirh: Sorry, I was too terse. I got it about making sure the profile should not leak. Followup question: should a profile actually leak (because someone like @fromanirh added a broken test!), do we have a way to recover or this means we polluted the tested cluster and the only remedy is let the cluster admin know so it can be fixed manually?

fedepaol: It's the tests responsability not to leave leftovers. In many cases this is achieved by cleaning the namespaces.
In the case of perf profiles though, the only way to clean them is to explicilty delete them. Whatever mechanism you come with, it will require adding the profile to be deleted somewhere (because you also don't want to delete them all), so it's still the dev's responsibility to make sure state don't leak (as with all the other cases).

fromanirh: Got it and makes sense. Thanks!

fromanirh:
ok, looks like this failure is real: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-kni_performance-addon-operators/378/pull-ci-openshift-kni-performance-addon-operators-master-e2e-gcp-operator-upgrade/1313366629148528640

I'm not completely sure why the e2e-gcp lane passed cleanly and the upgrade lane is not, so this deserve some extra investigation

functests/3_performance_status/status.go
```diff
@@ -133,4 +136,66 @@ var _ = Describe("Status testing of performance profile", func() {
...
+
+			}()
+
+			nodeLabels := map[string]string{
+				newLabel: "",
+			}
+
+			var cond *v1.Condition
+			Eventually(func() bool {
+				cond = profiles.GetCondition(nodeLabels, v1.ConditionDegraded)
```
fedepaol, line 192: I had a hard time understanding what this does. Would it be possible to have a local "isProfileDegraded" function (or something along that line) wrapping the code?

pkg/controller/performanceprofile/components/utils.go
```diff
@@ -108,3 +108,18 @@ func CPUListToMaskList(cpulist string) (hexMask string, err error) {
 	trimmedCPUMaskList := b.String()
 	return trimmedCPUMaskList, nil
 }
+
+// CPUListIntersect returns cpu ids found in both the provided cpulists, if any
+func CPUListIntersect(cpulistA, cpulistB string) ([]int, error) {
```
fedepaol, line 113: nit: cpuList

fedepaol (commented):
Just a couple of nits, looks good!

PR #385, opened 2020-10-01
Provide manifests and script to run upgrade test under the CI

cynepco3hahue:
Now that the CI install for use the operator, we should only:

- deploy MCP
- label the worker with the worker-cnf label
- deploy the performance profile


cynepco3hahue:
@fromanirh Can you please review?

PR #386, opened 2020-10-04
e2e: fix the runtime class test

cynepco3hahue:
Under our CI it happens pretty often that the test
does not get the path to the container cpuset cgroup,
probably it takes some time to sync cgroup changes between host
and machine-config-daemon container.

To see if it indeed the case I will add `Eventually` that will try
to fetch container cpuset cgroup for two minutes with 30 seconds samples.

### DIFF
--- a/openshift-ci/Dockerfile.registry.build
+++ b/openshift-ci/Dockerfile.registry.build
@@ -7,7 +7,7 @@ COPY deploy/olm-catalog /registry/performance-addon-operator-catalog
 # easier switching between CI clusters...
 # ENV REG_URL=default-route-openshift-image-registry.apps.build01.ci.devcluster.openshift.com
 # ENV REG_URL=registry.svc.ci.openshift.org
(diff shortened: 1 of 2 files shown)


openshift-ci/Dockerfile.registry.build
```diff
@@ -7,7 +7,7 @@ COPY deploy/olm-catalog /registry/performance-addon-operator-catalog
 # easier switching between CI clusters...
 # ENV REG_URL=default-route-openshift-image-registry.apps.build01.ci.devcluster.openshift.com
 # ENV REG_URL=registry.svc.ci.openshift.org
-ENV REG_URL=registry.build02.ci.openshift.org
+ENV REG_URL=registry.build01.ci.openshift.org
```
fromanirh, line 10: This should be in a separate commit

fromanirh (commented):
/lgtm
Adding `Eventually` works for me in general, I don't see any drawback

PR #391, opened 2020-10-07
e2e: fix getting a profile by the node selector

cynepco3hahue:
When the range used under the `for` loop it will create a new
scoped variable, that will have a different value for each loop cycle.
Once you will get the pointer on this variable it will point to the latest
element that the `for` loop will pass.

### DIFF
--- a/functests/utils/profiles/profiles.go
+++ b/functests/utils/profiles/profiles.go
@@ -25,12 +25,12 @@ func GetByNodeLabels(nodeLabels map[string]string) (*performancev1.PerformancePr
 	}
 
 	var result *performancev1.PerformanceProfile
-	for _, profile := range profiles.Items {
-		if reflect.DeepEqual(profile.Spec.NodeSelector, nodeLabels) {
+	for i := 0; i < len(profiles.Items); i++ {
+		if reflect.DeepEqual(profiles.Items[i].Spec.NodeSelector, nodeLabels) {
 			if result != nil {
 				return nil, fmt.Errorf("found more than one performance profile with specified node selector %v", nodeLabels)
 			}
-			result = &profile
+			result = &profiles.Items[i]
 		}
 	}
 


functests/utils/profiles/profiles.go
```diff
@@ -25,12 +25,12 @@ func GetByNodeLabels(nodeLabels map[string]string) (*performancev1.PerformancePr
 	}
 
 	var result *performancev1.PerformanceProfile
-	for _, profile := range profiles.Items {
-		if reflect.DeepEqual(profile.Spec.NodeSelector, nodeLabels) {
+	for i := 0; i < len(profiles.Items); i++ {
```
fromanirh, line 28: well, the broken code passed a review, so could you please add a couple lines to the commit message to emphasize why was broken before, so we know where to pay more attention next time?

functests/utils/profiles/profiles.go
```diff
@@ -25,12 +25,12 @@ func GetByNodeLabels(nodeLabels map[string]string) (*performancev1.PerformancePr
...
+		if reflect.DeepEqual(profiles.Items[i].Spec.NodeSelector, nodeLabels) {
 			if result != nil {
 				return nil, fmt.Errorf("found more than one performance profile with specified node selector %v", nodeLabels)
 			}
-			result = &profile
+			result = &profiles.Items[i]
```
yanirq, line 33: looks like `break` here could also have done the trick

cynepco3hahue: IIUC you want to see error message if you have more than one PP with the same nodeSelector

fromanirh (commented):
/lgtm
crystal clear now, thanks.

functests/utils/profiles/profiles.go
fromanirh: yes, that's the reason why break is not good. Otherwise it would have been just fine.

PR #394, opened 2020-10-08
functests: don't read stderr when getting logs

fromanirh:
We expect nothing useful from stderr, so let's not even read it
Capturing stderr is useful to troubleshoot failures, but it should not mixed with stdout and should be captured and kept separate.

### DIFF
--- a/functests/utils/utils.go
+++ b/functests/utils/utils.go
@@ -19,7 +19,7 @@ func BeforeAll(fn func()) {
 }
 
 func ExecAndLogCommand(name string, arg ...string) ([]byte, error) {
-	out, err := exec.Command(name, arg...).CombinedOutput()
+	out, err := exec.Command(name, arg...).Output()
 	klog.Infof("run command '%s %v':\n  out=%s\n  err=%v", name, arg, out, err)
 	return out, err
 }
--- a/functests/utils/utils.go
+++ b/functests/utils/utils.go
@@ -20,6 +20,9 @@ func BeforeAll(fn func()) {
 
 func ExecAndLogCommand(name string, arg ...string) ([]byte, error) {
 	out, err := exec.Command(name, arg...).Output()
-	klog.Infof("run command '%s %v':\n  out=%s\n  err=%v", name, arg, out, err)
+	klog.Infof("run command '%s %v' (err=%v):\n  stdout=%s\n", name, arg, err, out)
+	if exitError, ok := err.(*exec.ExitError); ok {
+		klog.Infof("run command '%s %v' (err=%v):\n  stderr=%s", name, arg, err, exitError.Stderr)
+	}
 	return out, err
 }


fromanirh:
/hold
let's address the troubleshooting part.

functests/utils/utils.go
```diff
@@ -19,7 +19,7 @@ func BeforeAll(fn func()) {
 }
 
 func ExecAndLogCommand(name string, arg ...string) ([]byte, error) {
-	out, err := exec.Command(name, arg...).CombinedOutput()
+	out, err := exec.Command(name, arg...).Output()
```
MarSik, line 22: Is there a way to at least log the stderr output? Or start the command in such a way that stderr goes to system stderr?

fromanirh: on it already. Quoting go docs (https://golang.org/pkg/os/exec/#Cmd.Output):
```
Any returned error will usually be of type *ExitError. If c.Stderr was nil, Output populates ExitError.Stderr. 
```
So the data is there, it's just about properly logging it if `err != nil`.

MarSik: Hmm we might get it indeed via err at least. The docs say: If c.Stderr was nil, Output populates ExitError.Stderr.

yanirq: How will this play our with tests running:  `Expect(err).ToNot(HaveOccurred())`  right after this command ?

fedepaol: @fromanirh you can assign a buffer to cmd.Stderr and eventually dump it if the execution returns an error.
See https://stackoverflow.com/questions/31095699/return-cmd-stdout-and-stderr-as-string-instead-of-printing-to-console-in-golang for example

fromanirh:
/hold cancel
I don't mind assigning a buf to `cmd.Stderr` if the current approach doesn't work, but I just need to understand why the current approach doesn't work: AFAIK should be a fine alternative to using `cmd.Stderr` :)

functests/utils/utils.go
fromanirh: @yanirq the function signature will not change

functests/utils/utils.go
```diff
@@ -19,7 +19,10 @@ func BeforeAll(fn func()) {
 }
 
 func ExecAndLogCommand(name string, arg ...string) ([]byte, error) {
-	out, err := exec.Command(name, arg...).CombinedOutput()
-	klog.Infof("run command '%s %v':\n  out=%s\n  err=%v", name, arg, out, err)
+	out, err := exec.Command(name, arg...).Output()
+	klog.Infof("run command '%s %v' (err=%v):\n  stdout=%s\n", name, arg, err, out)
+	if exitError, ok := err.(*exec.ExitError); ok {
+		klog.Infof("run command '%s %v' (err=%v):\n  stderr=%s", name, arg, err, exitError.Stderr)
```
MarSik, line 25: This will log stderr twice I think: err is printed as a whole using %v, isn't it?

cynepco3hahue:
works for me
will put on hold if someone wants to add an additional comment
/hold

functests/utils/utils.go
fromanirh: Nope: see https://golang.org/src/os/exec/exec.go line 477

MarSik (approved):

fedepaol:
Looks great, thanks!
/lgtm

PR #397, opened 2020-10-09
latency-test: add and pass initial delay

fromanirh:
We add an optional delay before to actually run the latency test
to give more room to the kubelet reconcile loop to kick in and make
sure all the pods are running in the proper cpusets.

Without this delay is harder to guarantee that other pods do not
run on exclusively allocated cpus.

Please note that using a delay is a mitigation of the inherent (lack of)
synchronization between the test and the kubelet cpumanager reconcile
loop. It is very possible that the reconciliation happens fast enough
even with initial zero delay, so this change add a test/troubleshooting
knob rather than fixing a misbehaviour.


fromanirh:
to be used alongside with: `https://quay.io/repository/fromani/oslat`

PR #400, opened 2020-10-12
e2e: fix runtime class test failures

cynepco3hahue:
Fix runtime class test failures when it runs on the environment with more than one worker-cnf node.

- add proper node selector to the pod
- revert `Eventually` that I added before because thought it was some race
the condition between the start of the container and the update of the cgroup

### DIFF
--- a/functests/1_performance/cpu_management.go
+++ b/functests/1_performance/cpu_management.go
@@ -276,6 +276,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 			// use the CPU load balancing runtime class
 			runtimeClassName := components.GetComponentName(profile.Name, components.ComponentNamePrefix)
 			testpod.Spec.RuntimeClassName = &runtimeClassName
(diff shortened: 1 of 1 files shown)


fromanirh (commented):
I would have preferred a separate revert + explicit fix
so I'll let the other reviewers decide

yanirq:
> I would have preferred a separate revert + explicit fix
> so I'll let the other reviewers decide

The description in the PR might be confusing so maybe indeed that's the way to go

cynepco3hahue:
@fromanirh Do you want it under the separate PR or commit?

cynepco3hahue:
@fromanirh @yanirq Updated the PR

fromanirh (approved):
/lgtm

fromanirh:
/hold
let's wait few hours until we reach consensus about https://github.com/openshift-kni/performance-addon-operators/pull/355
this is just about serializing changes and minimizing disruption, this PR is completely fine.

PR #405, opened 2020-10-15
Fix for update test - make verification of cpu affinity mask flexible

dshchedr:
test_id:28025 Test failed on the node with more than 4 CPUs. Switching
verification to regexp should make it more flexible.

Note: currently updating tests are disabled on CI because of bug so the fix will not be checked. I verified in manually on my local cluster.

### DIFF
--- a/functests/2_performance_update/updating_profile.go
+++ b/functests/2_performance_update/updating_profile.go
@@ -137,7 +137,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			table.Entry("[test_id:28070] verify that hugepages updated (NUMA node unspecified)", chkCmdLine, []string{"hugepagesz=2M"}, true, false),
 			table.Entry("verify that the right number of hugepages 1G is available on the system", chkHp1G, []string{"3"}, true, false),
 			table.Entry("verify that the right number of hugepages 2M is available on the system", chkHp2M, []string{"256"}, true, false),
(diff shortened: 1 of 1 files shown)


fromanirh (approved):
/lgtm

functests/2_performance_update/updating_profile.go
```diff
@@ -137,7 +137,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			table.Entry("[test_id:28070] verify that hugepages updated (NUMA node unspecified)", chkCmdLine, []string{"hugepagesz=2M"}, true, false),
 			table.Entry("verify that the right number of hugepages 1G is available on the system", chkHp1G, []string{"3"}, true, false),
 			table.Entry("verify that the right number of hugepages 2M is available on the system", chkHp2M, []string{"256"}, true, false),
-			table.Entry("[test_id:28025] verify that cpu affinity mask was updated", chkCmdLine, []string{"tuned.non_isolcpus=00000009"}, true, false),
+			table.Entry("[test_id:28025] verify that cpu affinity mask was updated", chkCmdLine, []string{"tuned.non_isolcpus=.*9"}, true, true),
```
cynepco3hahue, line 140: Can you please explain what exactly `.*9` mean, does it a regular expression?

dshchedr: Yes, it's regular expression (second "true" in this line means - use regex when checks parameters).
In cases when we have more than 4 CPUs we will have different non_isolcpus, for example for 8 CPU it will be 000000f9, so current regex should allow any characters with 9 at the end

PR #392, opened 2020-10-07
Conversion Webhook implementation

marcel-apf:
v8:
 - addressed @cynepco3hahue comments, improve the conversion code

v7:
 - addressed @cynepco3hahue comments, fixed bugs in the e2e test

v6:
 - addressed @fromanirh comments and improve the e2e test

v5:
- add v1<->v1alpha1 Webhook conversion test

v4:
 - rebased on "master" branch

v3:
 - rebased "operator-sdk-v1.0.0" branch

v2:
 -  Full working version.

Let v1 be the "Hub" and all other versions be spokes.
That means all API versions other than v1 shall supply conversion from/to v1.

PR #406, opened 2020-10-19
docs: Add API versions support doc

marcel-apf:
v2:
 - Added a Q&A section

Describe the relation between the Performance Addon Operator
and the v1,v1alpha1 Performance Profile APIs.

### DIFF
--- a/docs/api-versions.md
+++ b/docs/api-versions.md
@@ -1,0 +1,33 @@
+# API versions
+
+## Supported API Versions
+
+The Performance Addon Operator supports *v1* and *v1alpha1* [Performance Profile](docs/performance_profile.md) versions.
+There are no differences between the *v1* and *v1alpha1* versions, the upgrade has been made in order to
+mark the Performance Profile API as stable.
+
+## Upgrade from *v1alpha1* to *v1*.
+When upgrading from an older Performance Addon Operator version supporting only *v1alpha1* API to a newer one supporting
+also the *v1* API, the existing *v1alpha1* Performance Profiles will be converted on-the-fly using a "None" Conversion
+strategy and served to the Performance Addon Operator as *v1*.
+
+## Q&A
+What happens in practice if I install a v1-enabled PAO on a cluster? What should I expect?
+- PAO will expect v1 Performance Profiles and query only them. Existing v1alpha1 profiles will be served as v1 and
+  PAO works with them as usual.
+
+What happens if I submit a v1alpha1 profile, will I always get back a v1 profile?
+- Any of the existent Performance Profile CRs can be retrieved as both v1 and v1alpha1 since the Performance Profile CRD
+supports both API versions.
+    For example if we have a "manual" Performance Profile in the system we can query it as both v1alpha1 and v1:
+
+    ```oc get performanceprofiles.v1.performance.openshift.io manual```
+
+    ```oc get performanceprofiles.v1alpha1.performance.openshift.io manual```
+
+    However, the Performance Addon Operator will use all the existent Performance Profiles as v1 no matter
+    if they have been submitted as v1alpha1 or v1.
+
+Where can I find more information on API versioning?
+- https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning
+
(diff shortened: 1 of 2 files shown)


docs/api-versions.md
```diff
@@ -0,0 +1,12 @@
...
+## Supported API Versions
+
+The Performance Addon Operator supports *v1* and *v1alpha1* [Performance Profile](docs/performance_profile.md) versions.
+There are no differences between the *v1* and *v1alpha1* versions, the upgrade has been made in order to
+mark the Performance Profile API as stable.
+
+## Upgrade from *v1alpha1* to *v1*.
+When upgrading from an older Performance Addon Operator version supporting only *v1alpha1* API to a newer one supporting
+also the *v1* API, the existing *v1alpha1* Performance Profiles will be converted on-the-fly using a "None" Conversion
+strategy and served to the Performance Addon Operator as *v1*.
```
fromanirh, line 12: what happens in practice if I install a v1-enabled PAO on my cluster? what should I expect?

marcel-apf: PAO will expect v1 Performance Profiles and query only them. Existing v1alpha1 profiles are served as v1 and PAO works with them as usual

fromanirh: yep, and these flows should explained in this doc, these are the examples people will look up for :)

fromanirh: Also, worth pointing out that there is a strong overlap (and that's good!) between the flows we document here and the flow we test in https://github.com/openshift-kni/performance-addon-operators/pull/392/commits/fca81fd9940925befa0f944c2baf7464e7737520

marcel-apf: Added a Q&A section with the above. Regarding the commit you mentioned - I will update the doc as soon as the Conversion Webhook will be merged.

fromanirh: Not sure I understood completely: we should be able to check before we merge that the webhook e2e tests cover the flow we document. Or do we have reasons I'm not seeing to have the two PRs diverge?

marcel-apf: These are not related PRs. This PR is to be applied on the current master and does not depend on the Webhook Conversion PR.
This PR covers the state "before" the webhook conversion PR

marcel-apf: I plan to add on top of this doc explanations as part of V2 API PR. I added this PR as an out-of-band one since I have a JIRA issue for this sprint. It just covers current state: v1alha1 <-> v1 with None conversion strategy.

docs/api-versions.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Q&A
+What happens in practice if I install a v1-enabled PAO on a cluster? What should I expect?
+
+PAO will expect v1 Performance Profiles and query only them. Existing v1alpha1 profiles will be served as v1 and PAO works with them as usual.
```
fromanirh, line 17: so this means that existing v1alpha1 profiles will be stored in the cluster as v1 objects, right?
IOW, even if we submit a v1alpha1 profile, we will always get back a v1 profile, right?
Same: in etcd there always is v1 objects, isn't it?
This is what should be clarified in this doc: we can obviously learn the same from the code and/or from the functests, but the point of the docs is allowing to avoid that :)

docs/api-versions.md
fromanirh: Yes, the two PRs obviously don't depend on each other, but they are both about the same topic, that's what I meant

marcel-apf: yes, they are, this is why it will need a re-spin soon, is on my todo list.

docs/api-versions.md
marcel-apf: Submitting a new v1alpha1 profile after the PerformanceProfile has been updated to v1 will be stored as a v1 profile.
You can retrieve it as both v1alpha1 and v1 profile.
I am not sure if the existing v1alpha1 profiles are converted into v1 as part of the upgrade and stored in etcd as v1 or remain v1alpha1 and get converted on the fly. I am not sure it matters for us.

I am not sure this resolution is part of the PAO doc though

fromanirh: ok, let's just add this to the A above and that's good enough.

fromanirh (approved):
/lgtm

PR #408, opened 2020-10-20
Fix for test - MCP status in performance profile

dshchedr:
For some reason old approach with changing currentConfig does not work -
sometimes it takes more than 30 minuted for MCP to update the status to
"Degraded"

MCO team adviced to rewrite the test - put the node/mcp to degraded state
by creating bad machineConfig.


fromanirh (approved):
/lgtm

PR #412, opened 2020-10-20
Bug 1889633: Fix for test: hugepages with NUMA node

dshchedr:
Making test more robust. Currently it fails on the cluster with several numa nodes because
we compare the number of hugepages on one numa with total number of hugepages on the whole node.

### DIFF
--- a/functests/1_performance/hugepages.go
+++ b/functests/1_performance/hugepages.go
@@ -44,7 +44,9 @@ var _ = Describe("[performance]Hugepages", func() {
 
 		profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
 		Expect(err).ToNot(HaveOccurred())
-		Expect(profile.Spec.HugePages).ToNot(BeNil())
+		if profile.Spec.HugePages == nil || len(profile.Spec.HugePages.Pages) == 0 {
+			Skip("Hugepages is not configured in performance profile")
+		}
 	})
 
 	// We have multiple hugepages e2e tests under the upstream, so the only thing that we should check, if the PAO configure
@@ -59,10 +61,10 @@ var _ = Describe("[performance]Hugepages", func() {
 				hugepagesSize, err := machineconfig.GetHugepagesSizeKilobytes(page.Size)
 				Expect(err).ToNot(HaveOccurred())
 
-				availableHugepagesFile := fmt.Sprintf("/sys/kernel/mm/hugepages/hugepages-%skB/nr_hugepages", hugepagesSize)
+				availableHugepagesFile := fmt.Sprintf("/sys/devices/system/node/node%d/hugepages/hugepages-%skB/nr_hugepages", *page.Node, hugepagesSize)
 				nrHugepages := checkHugepagesStatus(availableHugepagesFile, workerRTNode)
 
-				freeHugepagesFile := fmt.Sprintf("/sys/kernel/mm/hugepages/hugepages-%skB/nr_hugepages", hugepagesSize)
+				freeHugepagesFile := fmt.Sprintf("/sys/devices/system/node/node%d/hugepages/hugepages-%skB/free_hugepages", *page.Node, hugepagesSize)
 				freeHugepages := checkHugepagesStatus(freeHugepagesFile, workerRTNode)
 
 				Expect(int32(nrHugepages)).To(Equal(page.Count), "The number of available hugepages should be equal to the number in performance profile")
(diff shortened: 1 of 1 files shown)


fromanirh (commented):
/lgtm
/hold
waiting for @cynepco3hahue 's review

functests/1_performance/hugepages.go
```diff
@@ -59,10 +59,10 @@ var _ = Describe("[performance]Hugepages", func() {
 				hugepagesSize, err := machineconfig.GetHugepagesSizeKilobytes(page.Size)
 				Expect(err).ToNot(HaveOccurred())
 
-				availableHugepagesFile := fmt.Sprintf("/sys/kernel/mm/hugepages/hugepages-%skB/nr_hugepages", hugepagesSize)
+				availableHugepagesFile := fmt.Sprintf("/sys/devices/system/node/node%d/hugepages/hugepages-%skB/nr_hugepages", *page.Node, hugepagesSize)
```
MarSik, line 62: We need to test the case where no numa node is specified. In such case the hugepages are evenly split across all numa nodes.

What I am saying is that I think this will break in such case when running in discovery mode.

cynepco3hahue:
```
if page.Node == nil {
					continue
				}
```
so it should work as expected

cynepco3hahue (commented):
I found additional problematic places:
- https://github.com/openshift-kni/performance-addon-operators/blob/334665976ecc7b267560b1215875deb9723e9967/functests/1_performance/hugepages.go#L47 when no hugepages defined under the performance profile we probably should skip hugepages tests instead of failing it under the `BeforeEach`
- https://github.com/openshift-kni/performance-addon-operators/blob/334665976ecc7b267560b1215875deb9723e9967/functests/1_performance/hugepages.go#L115 we should check that `len(profile.Spec.HugePages.Pages) > 0` before accessing the 0 element
- https://github.com/openshift-kni/performance-addon-operators/blob/334665976ecc7b267560b1215875deb9723e9967/functests/1_performance/hugepages.go#L144 we should use `/sys/kernel/mm/hugepages/hugepages-%skB/nr_hugepages`  because if the test will run on the environment with two or more NUMA nodes, no one will guarantee to you, that the pod will allocate hugepages from the NUMA node 0
the same true for https://github.com/openshift-kni/performance-addon-operators/blob/334665976ecc7b267560b1215875deb9723e9967/functests/1_performance/hugepages.go#L147

I do not have a problem if all the above problems will be handled by the separate PR

dshchedr:
Thanks Artyom!


> * https://github.com/openshift-kni/performance-addon-operators/blob/334665976ecc7b267560b1215875deb9723e9967/functests/1_performance/hugepages.go#L47
>    when no hugepages defined under the performance profile we probably should skip hugepages tests instead of failing it under the `BeforeEach`
> * https://github.com/openshift-kni/performance-addon-operators/blob/334665976ecc7b267560b1215875deb9723e9967/functests/1_performance/hugepages.go#L115
>    we should check that `len(profile.Spec.HugePages.Pages) > 0` before accessing the 0 element
> * https://github.com/openshift-kni/performance-addon-operators/blob/334665976ecc7b267560b1215875deb9723e9967/functests/1_performance/hugepages.go#L144
 I added "skip" for both these cases. Current tests will not work on a cluster without Hugepages configured.

>    we should use `/sys/kernel/mm/hugepages/hugepages-%skB/nr_hugepages`  because if the test will run on the environment with two or more NUMA nodes, no one will guarantee to you, that the pod will allocate hugepages from the NUMA node 0
>   the same true for https://github.com/openshift-kni/performance-addon-operators/blob/334665976ecc7b267560b1215875deb9723e9967/functests/1_performance/hugepages.go#L147
 good point! fixed

> I do not have a problem if all the above problems will be handled by the separate PR, @dshchedr give me know what do you prefer
This PR is about fixing tests issues, so let's have all fixes in one place

functests/1_performance/hugepages.go
```diff
@@ -44,7 +44,9 @@ var _ = Describe("[performance]Hugepages", func() {
 
 		profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
 		Expect(err).ToNot(HaveOccurred())
-		Expect(profile.Spec.HugePages).ToNot(BeNil())
+		if profile.Spec.HugePages == nil || profile.Spec.HugePages.Pages == nil {
```
cynepco3hahue, line 47: can you please change `profile.Spec.HugePages.Pages == nil` to `len(profile.Spec.HugePages.Pages == 0)`, I just unsure when under the API the list value is nil and when it not nil but does not have any elements, so the len check will cover both cases

dshchedr: hm, I was afraid if we can catch "nil pointer exceptions" in that case..
verified locally - works fine, thanks. Updated.

PR #416, opened 2020-10-22
e2e: check that it no error before accessing the profile fields

cynepco3hahue:
It will prevent panics under our CPU tests:
```
    runtime error: invalid memory address or nil pointer dereference
    /usr/local/go/src/runtime/panic.go:199
    Full Stack Trace
    github.com/openshift-kni/performance-addon-operators/functests/1_performance.glob..func1.1()
    	/go/src/github.com/openshift-kni/performance-addon-operators/functests/1_performance/cpu_management.go:54 +0x506
    github.com/openshift-kni/performance-addon-operators/functests/1_performance_test.TestPerformance(0xc0002a9a00)
    	/go/src/github.com/openshift-kni/performance-addon-operators/functests/1_performance/test_suite_performance_test.go:49 +0x239
    testing.tRunner(0xc0002a9a00, 0x1831a28)
    	/usr/local/go/src/testing/testing.go:909 +0xc9
    created by testing.(*T).Run
    	/usr/local/go/src/testing/testing.go:960 +0x350
```

### DIFF
--- a/functests/1_performance/cpu_management.go
+++ b/functests/1_performance/cpu_management.go
@@ -51,11 +51,9 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 		Expect(workerRTNodes).ToNot(BeEmpty())
 		workerRTNode = &workerRTNodes[0]
 		profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
-		By(fmt.Sprintf("Checking the profile %s with cpus %#v", profile.Name, profile.Spec.CPU))
-
 		Expect(err).ToNot(HaveOccurred())
-		Expect(profile.Spec.HugePages).ToNot(BeNil())
 
+		By(fmt.Sprintf("Checking the profile %s with cpus %#v", profile.Name, profile.Spec.CPU))
 		balanceIsolated = true
 		if profile.Spec.CPU.BalanceIsolated != nil {
 			balanceIsolated = *profile.Spec.CPU.BalanceIsolated


fromanirh:
Do we know why this started to break?

cynepco3hahue:
@fromanirh Probably the reason, the test failed to find the profile by the node selector, but we can not see the error, because it fails before with the panic. The PR does not fix any test, it only exposes the error before access to the profile fields that will prevent panic.

fromanirh:
> @fromanirh Probably the reason, the test failed to find the profile by the node selector, but we can not see the error, because it fails before with the panic. The PR does not fix any test, it only exposes the error before access to the profile fields that will prevent panic.

Sure, I'm just trying to understand the context: if this an old error which surfaced just now, or if a recent PR made it worse/broke something.

cynepco3hahue:
It definitely some flakiness or bug that started to happen recently I did not see such errors in the past.

functests/1_performance/cpu_management.go
```diff
@@ -51,11 +51,9 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 		Expect(workerRTNodes).ToNot(BeEmpty())
 		workerRTNode = &workerRTNodes[0]
 		profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
-		By(fmt.Sprintf("Checking the profile %s with cpus %#v", profile.Name, profile.Spec.CPU))
-
 		Expect(err).ToNot(HaveOccurred())
-		Expect(profile.Spec.HugePages).ToNot(BeNil())
```
fromanirh, line 57: ok, hugepages may be missing, and that's legit.

fromanirh (approved):
/lgtm

PR #413, opened 2020-10-21
Move Performance Profile to V2 API

marcel-apf:
v5:
 - Rebased on master
 - Update the 4.7 CSV and CRD instead of 4.6 (@cynepco3hahue )

v4:
 - Rebased again on master after a e2e test update
 - Fix a minor bug in the e2e test following the rebase

v3:
 - Rebased on master
 - Fixed all v2 issues exposed by the "Webhook Conversion" PR

v2:
 - Rebased on  "webhook conversion PR"
 - Fixed the E2E API versioning test

Add v2 API for the Performance Profile that is identical to v1 and v1apha1 as a first step.
The real gain will be the next PR that will deal with the real issues in hand will be much easier
to review.

Steps:
Generate the new v2 API and add the generated and  webhook conversion code.
Point the code to the new v2 API (the controller and tests will work only with v2 API)
Point the docs and the cluster-setup code to the new API
Add an E2E test for v1 <-> v2 conversion and make the necessary for the v1 <-> v1alpha1  test.

PR #417, opened 2020-10-22
e2e: fix conversion test

cynepco3hahue:
- change the node selector for the second profile, just to avoid any missconfigurations
- wait for the v1alpha1Profile deletion

### DIFF
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -15,10 +15,12 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/api/node/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog"
 	"k8s.io/utils/pointer"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


fromanirh (approved):
makes a lot of sense. Great investigation and kudos for the quick fix.

fromanirh:
/hold
giving time to other reviewes.

functests/1_performance/performance.go
```diff
@@ -360,15 +362,26 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 			By("Checking v1alpha1 -> v1 conversion")
 			v1alpha1Profile.Name = "v1alpha"
 			v1alpha1Profile.ResourceVersion = ""
+			v1alpha1Profile.Spec.NodeSelector = map[string]string{"test": "test"}
```
marcel-apf, line 365: Why do we need this?

functests/1_performance/performance.go
```diff
@@ -360,15 +362,26 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+			defer func() {
+				Expect(testclient.Client.Delete(context.TODO(), v1alpha1Profile)).ToNot(HaveOccurred())
+
+				Eventually(func() bool {
+					if err := testclient.Client.Get(context.TODO(), key, v1alpha1Profile); errors.IsNotFound(err) {
+						klog.Errorf("failed to get performance profile %s: %v", v1alpha1Profile.Name, err)
+						return true
+					}
+					return false
+				}, time.Minute, 10*time.Second).Should(Equal(true))
```
marcel-apf, line 381: A more compact version
	Expect(profiles.WaitForDeletion(key, 60*time.Second)).ToNot(HaveOccurred())
I saw it somewhere else in the code.  I don't mind one wait or another, but maybe we should follow the same conventions.

cynepco3hahue: yep but it assumes that you will pass v1 Profile to it and not the v1alpha1

marcel-apf: I have a patch that fixes also this... in API V2 PR:
https://github.com/openshift-kni/performance-addon-operators/pull/413/commits/3e009043f7558abde7a90621022bef09b0f53315

functests/1_performance/performance.go
cynepco3hahue: in general, we should have validation that should check that we do not have two profile with the same node selector(but I failed to find it under the code, so I think we need to add it), but anyway I want to avoid the creation of duplicate MC and KC for the same MCP because it can start update flow for the MCP

functests/1_performance/performance.go
cynepco3hahue: feel free to refactor the test under the API v2 PR :) I want to avoid additional changes under the current PR

functests/1_performance/performance.go
fromanirh: @cynepco3hahue please file a PAO issue about the missing validation

functests/1_performance/performance.go
marcel-apf: ok

fromanirh: Uhm https://github.com/openshift-kni/performance-addon-operators/commit/3e009043f7558abde7a90621022bef09b0f53315#diff-97cae1e7b24671bf357a6d87a359f597b8b728991d065521ede1c26cc146f73bL52 always expects a v2 Profile, so it doesn't seem a full replacement (yet) of what @cynepco3hahue is doing on his code.

marcel-apf: @fromanirh it is a full replacement, PAO with a V2 PerformanceProfile expects only  v2 profiles. 
Any non v2 profiles will be converted before served and this is why the above patch is ok (tested)

marcel-apf: I will rebase the V2 PR aftert his one gets merged and make the necessary changes.

fromanirh: ok, will review the new PR.

marcel-apf: So sad, it already passed CI, never mind.

PR #419, opened 2020-10-22
Generate CSV 4.7 and move all variable from 4.6 to 4.7

cynepco3hahue:
Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

PR #420, opened 2020-10-26
Change upgrade test PerformanceProfile version to v1

cynepco3hahue:
When new 4.7 CSV was introduced(#419), it included the change for the upgrade test flow from `4.5->4.6` to `4.6->4.7`, but `4.6` does not have a conversion webhook and does not support the creation of new `PerformanceProfile` v1alpha1 resource.

To fix the upgrade flow the new `PerformanceProfile` should have version v1.

### DIFF
--- a/cluster-setup/ci-upgrade-test-cluster/performance/performance_profile.yaml
+++ b/cluster-setup/ci-upgrade-test-cluster/performance/performance_profile.yaml
@@ -1,4 +1,4 @@
-apiVersion: performance.openshift.io/v1alpha1
+apiVersion: performance.openshift.io/v1
 kind: PerformanceProfile
 metadata:
   name: upgrade-test
--- a/cluster-setup/upgrade-test-cluster/performance/performance_profile.yaml
+++ b/cluster-setup/upgrade-test-cluster/performance/performance_profile.yaml
@@ -1,4 +1,4 @@
-apiVersion: performance.openshift.io/v1alpha1
+apiVersion: performance.openshift.io/v1
 kind: PerformanceProfile
 metadata:
   name: upgrade-test


cluster-setup/ci-upgrade-test-cluster/performance/performance_profile.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: performance.openshift.io/v1alpha1
+apiVersion: performance.openshift.io/v1
```
fromanirh, line 1: makes sense, but can we have a bit more context in the commit message (few lines) like
1. what's the error we see
2. (thus) why this patch fixes it
3. (if possible) when the issue started to happened
?

marcel-apf: I also want to see the error. PAO should be able to work with v1alpha1 profiles.

cynepco3hahue: @fromanirh @marcel-apf I updated the commit and description message, @marcel-apf please correct me if I wrong.

marcel-apf: @cynepco3hahue as we discussed offline the conversion webhook is available as soon as the PAO pod is up and running.
We should test it before creating old profiles. This patch will work, hiding the mentioned issue

marcel-apf: The PR subject remains incorrect

cynepco3hahue: Yep I tested it locally and it worked for me, I will close the PR, because it does not really fix anyhting

PR #422, opened 2020-10-27
[release-4.5] ci: adaptations to operator pipeline

cynepco3hahue:


fromanirh:
/lgtm
not sure we need this in 4.5 - which should still work with appregistry
but we can give it a try and see how it goes

PR #424, opened 2020-10-28
Add globallyDisableIrqLoadBalancing flag to Performance Profile

marcel-apf:
v5:
 - Solved a bug in the E2E test limiting it to 64 CPUs machines (@fromanirh)
 - Rebased on master

v4:
 - Add an E2E testing the GloballyDisableIrqLoadBalancing Spec field

v3:
 - Fix API conversion test bug found by CI

v2:
- Rebased on master
- Add a comment on why we are making the change (@fromanirh )
- Solved a bug and updated the template with a new logic


build/assets/tuned/openshift-node-performance
```diff
@@ -42,6 +42,13 @@ net.core.busy_poll=50                         # network-latency
 net.ipv4.tcp_fastopen=3                       # network-latency
 vm.stat_interval = 10                         # cpu-partitioning  #realtime
 
+[irqbalance]
+{{if .GloballyDisableIrqLoadBalancing}}
+banned_cpus=${isolated_cores}
+{{else}}
+banned_cpus=""
+{{end}}
```
fromanirh, line 50: let's loop in @yanirq

build/assets/tuned/openshift-node-performance
```diff
@@ -42,6 +42,13 @@ net.core.busy_poll=50                         # network-latency
 net.ipv4.tcp_fastopen=3                       # network-latency
 vm.stat_interval = 10                         # cpu-partitioning  #realtime
 
+[irqbalance]
+{{if .GloballyDisableIrqLoadBalancing}}
+banned_cpus=${isolated_cores}
+{{else}}
+banned_cpus=""
```
fromanirh, line 49: do we need to have this if GloballyDisableIrqLoadBalancing is false? meaning: could the following work?
```
{{if .GloballyDisableIrqLoadBalancing}}
[irqbalance]
banned_cpus=${isolated_cores}
{{end}}
```

fromanirh (commented):
at glance looks nice, will have a deeper review later

build/assets/tuned/openshift-node-performance
marcel-apf: The parent tuned profile has:
```
[irqbalance]
banned_cpus=${isolated_cores}
```
I wanted to override that. [I will double check]

build/assets/tuned/openshift-node-performance
marcel-apf: Good idea

build/assets/tuned/openshift-node-performance
fromanirh: OK, if you need to override it it's fine like it is.

marcel-apf: Now I am too (depends on the implementation of tuned). We will know in about 1 hour

yanirq: I would dig a bit more in https://github.com/openshift/cluster-node-tuning-operator/blob/master/assets/tuned/daemon/tuned/plugins/plugin_irqbalance.py to see the implications of an empty string and also see what you actually set on the system

yanirq: we had a test once checking IRQBALANCE_BANNED_CPUS value in the irqbalance config file when we had initramfs generated. maybe its time to bring this test back.

marcel-apf: @yanirq I actually tested live both cases and on empty string the cpu mask was 0 as expected
Regarding the mentioned test, I have something in mind, but  a cheat-sheet is always welcomed :) , do you remember the file or the test name? I will dig anyway.

marcel-apf: I've gone fast over the tuned code and it should be ok.
I will triple-check anyway, better safe...

marcel-apf:
Several tests failed, looks like a glitch in CI (system too slow?)
Worth immediate retest
/retest

build/assets/tuned/openshift-node-performance
marcel-apf: @fromanirh it took me a lot of time, but at least tuned started to work (MCO still can't install the rt kernel)
To answer your question, as I thought, if we don't add  banned_cpus="" the  isolated CPU set remains masked.
The reason is we inherit the cpu-partitioning bits, so we need to override banned_cpus with an empty value.
The other way around would work, e.g. setting banned_cpus only to empty string when needed, but I think a verbose mode is desired

marcel-apf:
Since all the errors don't make sense, and it seems the system was slow all day long, last try today
/retest

build/assets/tuned/openshift-node-performance
fromanirh: thanks for the extra check @marcel-apf . I'm fine with your approach. I just think this
```
f we don't add banned_cpus="" the isolated CPU set remains masked.
The reason is we inherit the cpu-partitioning bits, so we need to override banned_cpus with an empty value.
```
should be added as code comment for future reference.

marcel-apf:
Another day, another conclusion. The last failure is actually ok.
The reason is I started a new section [irqbalance] in the middle of [sysctl] one...

build/assets/tuned/openshift-node-performance
marcel-apf: I changed the template to:
```{{if not .GloballyDisableIrqLoadBalancing}}
[irqbalance]
# Override the value set by cpu-partitioning with an empty one
banned_cpus=""
{{end}}
```
The reason is we don't know what exactly the parent use as banned cpus, they may update the formula in the future.
What we do know is we don't want to bann the CPUs for our case

functests/utils/nodes/nodes.go
```diff
@@ -184,3 +185,18 @@ func HasPreemptRTKernel(node *corev1.Node) error {
 
 	return nil
 }
+
+func BannedCPUs(node corev1.Node) (banned uint64, err error) {
+	cmd := []string{"sed", "-n", "s/^IRQBALANCE_BANNED_CPUS=\\(.*\\)/\\1/p", "/rootfs/etc/sysconfig/irqbalance"}
```
fromanirh, line 190: so this is a bitmask, right?
this worries me a bit because it suggests it can't hold more than 64 CPUs :\

functests/2_performance_update/updating_profile.go
```diff
@@ -55,6 +56,67 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+			if irqLoadBalancingDisabled {
+				expectedBannedCPUs = isolatedCPUs
+			} else {
+				expectedBannedCPUs = 0
+			}
+			profile.Spec.GloballyDisableIrqLoadBalancing = &irqLoadBalancingDisabled
+
+			By("Updating the performance profile")
+			Expect(testclient.Client.Update(context.TODO(), profile)).ToNot(HaveOccurred())
+			defer func() { // return initial configuration
```
fromanirh, line 93: it's a test, so we don't care too much about the traffic. Can't we just update back with the original profile and save some complexity?

functests/2_performance_update/updating_profile.go
```diff
@@ -55,6 +56,67 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+				spec, err := json.Marshal(initialProfile.Spec)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(testclient.Client.Patch(context.TODO(), profile,
+					client.RawPatch(
+						types.JSONPatchType,
+						[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
+					),
+				)).ToNot(HaveOccurred())
```
fromanirh, line 101: do we need to wait here?

functests/2_performance_update/updating_profile.go
```diff
@@ -55,6 +56,67 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+			irqLoadBalancingDisabled = !irqLoadBalancingDisabled
+			if irqLoadBalancingDisabled {
+				expectedBannedCPUs = isolatedCPUs
+			} else {
+				expectedBannedCPUs = 0
+			}
+			profile.Spec.GloballyDisableIrqLoadBalancing = &irqLoadBalancingDisabled
+
+			By("Updating the performance profile")
+			Expect(testclient.Client.Update(context.TODO(), profile)).ToNot(HaveOccurred())
```
fromanirh, line 92: I think we need to wait here, don't we?

functests/2_performance_update/updating_profile.go
```diff
@@ -55,6 +56,67 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+				Expect(err).ToNot(HaveOccurred())
+				Expect(testclient.Client.Patch(context.TODO(), profile,
+					client.RawPatch(
+						types.JSONPatchType,
+						[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
+					),
+				)).ToNot(HaveOccurred())
+			}()
+
+			Eventually(func() error {
```
fromanirh, line 104: with the added waits(), we can use the same loop as line 74 above, which is nice for consistency

fromanirh (commented):
my biggest concern is to wait after we restore the original profile, everything else which caught my eye is about possible improvements that can wait

functests/utils/nodes/nodes.go
marcel-apf: indeed, however the default /etc/sysconfig/irqbalance value suggests a 64bit field. I will dig more to see what is going on, I will increase the field space if necessary. (which probably is)

functests/2_performance_update/updating_profile.go
marcel-apf: The snippet is taken from the same file: https://github.com/openshift-kni/performance-addon-operators/blob/master/functests/2_performance_update/updating_profile.go#L264

I really don't care one way or another, but should we be consistent?

functests/2_performance_update/updating_profile.go
marcel-apf: We updated the profile. Now the tuned profile will be updated and the node will be updated. It happens async, right?

marcel-apf: Following the other snippet I would say no. The update is very "fast". If we do wait, we should wait on both, right?

functests/2_performance_update/updating_profile.go
marcel-apf: Not really, we wait once for the node to have the right settings. Anything in between does not worth the wait.

functests/2_performance_update/updating_profile.go
marcel-apf: What do you mean? 
The first loop doesn't have to wait for anything, this one does.
If you have a nicer way to code it, please point me to it, I will gladly improve (I am not mastering the Ginko stuff yet)

functests/utils/nodes/nodes.go
fromanirh: IMO the best option would be the list format of cpuset, like linux does (https://man7.org/linux/man-pages/man7/cpuset.7.html#FORMATS)

marcel-apf: I Will follow, thanks!

fromanirh: sure, even though it seems for irqbalance we are out of luck and stuck with the bitmask :\
```
#
# IRQBALANCE_BANNED_CPUS
# 64 bit bitmask which allows you to indicate which cpu's should
# be skipped when reblancing irqs. Cpu numbers which have their
# corresponding bits set to one in this mask will not have any
# irq's assigned to them on rebalance
#
#IRQBALANCE_BANNED_CPUS=
```

functests/2_performance_update/updating_profile.go
fromanirh: ok

functests/2_performance_update/updating_profile.go
fromanirh: let's keep it consistent and change everything in one go.

functests/2_performance_update/updating_profile.go
fromanirh: ok, just checking

functests/2_performance_update/updating_profile.go
fromanirh: ok, we can keep as it is.

functests/utils/nodes/nodes.go
fromanirh: OK, I reviewed the irqbalance code and on their side they can handle up until 4096 cpus encoded as bitmask. So irqbalance is fine. The potential issue is on our side (test) if we represent the bitmap using native types, so we are limited by (u)int64.

marcel-apf: Right, also there is a parsing issue. It turns out they delimit each 32 CPUs with a comma, something like 00000000,00000000,....
Another thing that caught my mind is:
```
From Red Hat Enterprise Linux 7.2, the irqbalance tool automatically avoids IRQs on CPU cores isolated via the isolcpus= kernel parameter if IRQBALANCE_BANNED_CPUS is not set in the /etc/sysconfig/irqbalance file.
```
See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_for_real_time/7/html/tuning_guide/interrupt_and_process_binding


That means what we are doing actually makes sense, setting the mask to "nothing", which is different from not setting it at all, in which case the irq load balancing will actually be disabled for the isolated cpus.

functests/2_performance_update/updating_profile.go
marcel-apf: agreed, we need a "cleaning" PR.

functests/2_performance_update/updating_profile.go
marcel-apf: Please elaborate more, I want to at lest know it. Can you add more context to the "Wait()" ?
Also we should have conventions.

functests/utils/nodes/nodes.go
marcel-apf: And regarding the CPU mask set, you are perfectly write. I didn't even know about it, thanks! I am going to see how it is done in go.

functests/2_performance_update/updating_profile.go
marcel-apf: Once we re-write, we can add the wait as a good practice though.

functests/utils/nodes/nodes.go
marcel-apf: Turns out we are already using kubelet's (!!) cpuset.CPUSet type. Nice, I will use it.

functests/2_performance_update/updating_profile.go
fromanirh: Setting up conventions (but most notably enforce them) would be interesting.
Thing is the usage of Eventually() here suggests that changes to the cluster made in the previous steps of the tests (e.g. change the profile) may take some time to be accepted, and then the state of the system will Eventually() be what's expected - otherwise the test will fail.
That works.
Another way is to wait explicitely for the changes to be accepted by the cluster, hence after each stage is performed the state of the cluster is known (or otherwise a previous expectation failed, letting the test fail). Once the state of the cluster is known, there is often no need to use Eventually(), which may hide bugs.
But again is not a clear cut or an obvious choice: it depends on how the test is written, and sometimes depends also on other factos (e.g. are we trying to cut down the test execution time?)

marcel-apf: Thanks for the explanation, so basically if we change the profile we should test that the PAO updated the profile and NTO also updated the tuned config AND finished to update the node. I think it will be doable if we'll have all this info in the PAO status section. Then we can simply wait for the status to be updated.  This is what you have in mind?

fromanirh: pretty much, yes. We can wait directly on MCP/NTO statuses if required or simpler. Anyway, no big deal for this PR.

fromanirh (approved):
/lgtm
/hold
nice!

letting other reviewers chime in

marcel-apf:
```[1mSTEP[0m: Creating the PerformanceProfile
[1mSTEP[0m: Getting MCP for profile
W1104 14:56:55.547801   15587 config.go:57] A PerformanceProfile with name performance already exists! If created externally, tests might have unexpected behaviour
[1mSTEP[0m: MCP is already unpaused
[1mSTEP[0m: Waiting for the MCP to pick the PerformanceProfile's MC
[1mSTEP[0m: Waiting for MCP being updated
I1104 14:56:55.575281   15587 mcps.go:133] MCP is targeting 1 node(s)

[91m[1m• Failure [1800.050 seconds][0m```

Slow CI? I will wait 1 hour and try again

File: functests/1_performance/performance.go [hunk 8]
```diff
@@ -671,6 +671,15 @@ func verifyV2Conversion(v2Profile *performancev2.PerformanceProfile, v1Profile *
 		}
 	}
 
+	for _, f := range v2Profile.GetObjectMeta().GetManagedFields() {
+		if f.APIVersion == performancev1alpha1.GroupVersion.String() ||
+			f.APIVersion == performancev1.GroupVersion.String() {
+			if v2Profile.Spec.GloballyDisableIrqLoadBalancing == nil || !*v2Profile.Spec.GloballyDisableIrqLoadBalancing {
+				return fmt.Errorf("globallyDisableIrqLoadBalancing field must be set to true")
```
yanirq, line 678: must be set as in must have a value ?  can't it be nil ?  its defined as omitempty : 
`json:"globallyDisableIrqLoadBalancing,omitempty"`

marcel-apf: Yes, is on purpose, when converting a v1 profile to v2 we must set the flag to true on order to keep the irq load balancing disabled as in v1 profile.

functests/2_performance_update/updating_profile.go
```diff
@@ -55,6 +56,70 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 		Expect(err).ToNot(HaveOccurred())
 	})
 
+	Context("Verify GloballyDisableIrqLoadBalancing Spec field", func() {
+		It("[test_id:36150] Verify that IRQ load balancing is enabled/disabled correctly", func() {
```
dshchedr, line 60: just curious, is it possible to make this test as part of common "performance profile updating" test?
https://github.com/openshift-kni/performance-addon-operators/blob/dfe203e14f050dda0f8f39dda24b567db6b91782/functests/2_performance_update/updating_profile.go#L123

It may help us to reduce wasting time on "updating / reverting back" node tunnings.

marcel-apf:
slow CI, trying again
/retest

functests/2_performance_update/updating_profile.go
marcel-apf: I thought about it, but is a very fast test, is only updating the tuned profile, all the test takes a few seconds.
Also the test has checks before and after the update, it doesn't feel right to check the business logic in "Before All"

marcel-apf:
/unhold
we have a second lgtm

marcel-apf:
/retest
Ci bug fixed?

fromanirh:
good news: CI seems to be healed
bad news: a (trivial) rebase is needed :\

marcel-apf:
> good news: CI seems to be healed
> bad news: a (trivial) rebase is needed :\

done

PR #428, opened 2020-11-01
Small improvement to the generation of content for the bundle image

cynepco3hahue:
- move common constatns to the separate file
- do not generate CSV for older versions to avoid change of older manifests
- generate metadata during the container building


hack/csv-generate.sh
```diff
@@ -1,26 +1,17 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
marcel-apf, line 1: ?

hack/csv-generate.sh
```diff
@@ -1,26 +1,17 @@
...
+
 GOROOT=$(go env GOROOT)
 export GOROOT
 
-OLD="4.4.0"
-PREV="4.6.0"
-LATEST="4.7.0"
-LATEST_CHANNEL="4.7"
-CSV_SKIP_RANGE=">=${PREV} <${LATEST}"
+CSV_SKIP_RANGE=">=${PREV_CSV_VERSION} <${CSV_VERSION}"
```
marcel-apf, line 11: Why not move all constants to common?

hack/generate-metadata.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e
+
+# shellcheck source=common.sh
+source "$(dirname "$0")/common.sh"
+
+# nothing to do, the annotation file already exists
+if [ -f "${METADATA_DIR}/${CSV_VERSION}/annotations.yaml" ]; then
```
marcel-apf, line 9: I would always create this file

hack/csv-generate.sh
```diff
@@ -94,17 +85,14 @@ if [[ "$CSV_VERSION" != "$PREV" ]] && [[ "$CSV_VERSION" != "$OLD" ]]; then
     "${MAINTAINERS}" \
     "${EXTRA_ANNOTATIONS}"
 
-    # restore the deleted CRD
-    cp "${CSV_DIR}/performance.openshift.io_performanceprofiles_crd.yaml" \
-      "${OUT_CSV_DIR}/performance.openshift.io_performanceprofiles_crd.yaml"
-fi
```
marcel-apf, line 100: fi deleted?

marcel-apf (commented):
Other than some questions I am ok with it.
However it looks it can be divided in different patches, one that extracts the variables, one that deals with indentation fixes (thanks!) and so on

hack/csv-generate.sh
cynepco3hahue: Running a command through `/usr/bin/env` has the benefit of looking for whatever the default version of the program is in your current environment, see https://stackoverflow.com/questions/16365130/what-is-the-difference-between-usr-bin-env-bash-and-usr-bin-bash for more details. We have some trouble with using `#!/bin/bash` under the KubeVirt project, so I just change it to a more flexible version.

hack/csv-generate.sh
cynepco3hahue: Currently, I moved only constants that used under different scripts, so we can move this one as well, once we will see that it needed.

hack/generate-metadata.sh
cynepco3hahue: ok will override it

hack/csv-generate.sh
cynepco3hahue: you can found `f` here https://github.com/openshift-kni/performance-addon-operators/pull/428/files#diff-abe5157e9fb248d431200c7a3e6a18a9507be8497dd209f93569411139fb3952R98

cynepco3hahue:
> Other than some questions I am ok with it.
> However it looks it can be divided in different patches, one that extracts the variables, one that deals with indentation fixes (thanks!) and so on

I will move the generate-metadata to the separate commit

hack/csv-generate.sh
marcel-apf: thanks, this should be wide available knowledge 
We should also have that in all scripts, right?

hack/csv-generate.sh
marcel-apf: ok

hack/csv-generate.sh
marcel-apf: aha

hack/csv-generate.sh
cynepco3hahue: yep, I will create a separate PR for this.

PR #434, opened 2020-11-04
validate: check for no isolated/reserved cpus

fromanirh:
It's legal, albeit probably not what users want, to
have empty sets for either reserved and isolated CPUs.
Hence make sure tests cover this case.

### DIFF
--- a/pkg/controller/performanceprofile/components/profile/profile.go
+++ b/pkg/controller/performanceprofile/components/profile/profile.go
@@ -25,7 +25,7 @@ func ValidateParameters(profile *performancev2.PerformanceProfile) error {
 	} else if profile.Spec.CPU.Isolated == nil {
 		return validationError("you should provide CPU.Isolated section")
 	}
-	if err := validateCPUCoresGrouping(profile.Spec.CPU); err != nil {
+	if err := validateCPUCores(profile.Spec.CPU); err != nil {
 		return err
 	}
 
@@ -177,12 +177,15 @@ func validateNUMA(numa *performancev2.NUMA) error {
 	return nil
 }
 
-func validateCPUCoresGrouping(cpus *performancev2.CPU) error {
-	overlap, err := components.CPUListIntersect(string(*cpus.Reserved), string(*cpus.Isolated))
+func validateCPUCores(cpus *performancev2.CPU) error {
+	cpuLists, err := components.NewCPULists(string(*cpus.Reserved), string(*cpus.Isolated))
 	if err != nil {
-		return validationError(fmt.Sprintf("internal error: %v", err))
+		return validationError(fmt.Sprintf("internal error parsing cpus: %v", err))
 	}
-	if len(overlap) != 0 {
+	if cpuLists.CountIsolated() == 0 {
+		return validationError("a performance profile must contain isolated cpus")
+	}
+	if overlap := cpuLists.Intersect(); len(overlap) != 0 {
 		return validationError(fmt.Sprintf("reserved and isolated cpus overlap: %v", overlap))
 	}
 	return nil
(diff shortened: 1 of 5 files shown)


fromanirh:
the errors look silly: `vet: pkg/controller/performanceprofile/components/profile/profile_test.go:92:20: undeclared name: v1` but we use already the very same name in a very similar way. hence, retrying to check that was a glitch or not.

pkg/controller/performanceprofile/components/profile/profile_test.go
```diff
@@ -88,6 +88,24 @@ var _ = Describe("PerformanceProfile", func() {
...
+		It("should allow cpus allocation with no reserved CPUs", func() {
+			reservedCPUs := v1.CPUSet("")
+			isolatedCPUs := v1.CPUSet("0-7")
+			profile.Spec.CPU.Reserved = &reservedCPUs
+			profile.Spec.CPU.Isolated = &isolatedCPUs
+			err := ValidateParameters(profile)
+			Expect(err).Should(Not(HaveOccurred()))
+		})
+
+		It("should allow cpus allocation with no isolated CPUs", func() {
```
cynepco3hahue, line 100: We have strict validation regarding isolated CPUs, https://github.com/openshift-kni/performance-addon-operators/blob/af31700a496376c6cf70508ba0e65289479bc127/pkg/controller/performanceprofile/components/profile/profile.go#L23

fromanirh: sure, adding the field with empty set is legal, isn't it?

cynepco3hahue: Yep, it should work, but it definitely wasn't an intention :)

fromanirh: Fair enough, so in this PR I think we should either add tests to check this behaviour (because we allow it) or pivot the PR and add a validation to reject a setting with no isolated CPUs. Either is fine for me. Does this plan make sense?

fromanirh: cc @MarSik

cynepco3hahue: totally makes sense, +1 to add vaalidation

fromanirh:
botched rebase, took like 2 attempts to get it right

pkg/controller/performanceprofile/components/profile/profile.go
```diff
@@ -177,12 +177,15 @@ func validateNUMA(numa *performancev2.NUMA) error {
...
-func validateCPUCoresGrouping(cpus *performancev2.CPU) error {
-	overlap, err := components.CPUListIntersect(string(*cpus.Reserved), string(*cpus.Isolated))
+func validateCPUCores(cpus *performancev2.CPU) error {
+	cpuLists, err := components.NewCPULists(string(*cpus.Reserved), string(*cpus.Isolated))
 	if err != nil {
-		return validationError(fmt.Sprintf("internal error: %v", err))
+		return validationError(fmt.Sprintf("internal error parsing cpus: %v", err))
 	}
-	if len(overlap) != 0 {
+	if numIsolated := cpuLists.CountIsolated(); numIsolated == 0 {
```
cynepco3hahue, line 185: nit: you do not need here a new variable
```
if cpuLists.CountIsolated() == 0
```
should work as well, but if from your perspective it has better readability I will not resist it :)

cynepco3hahue:
I have just some nit
/lgtm
/hold

pkg/controller/performanceprofile/components/profile/profile.go
fromanirh: I like your way more!

marcel-apf:
Looks like the PR is ready

/unhold
/lgtm

fromanirh:
/hold
we need e2e tests.

fromanirh:
deleted (thus closed) by mistake

fromanirh:
/retest
network error

fromanirh:
/retest
provisioning failed, seems platform issue.

fromanirh:
/retest
a cluster bootstrap failure can't be caused by this patch

fromanirh:
/retest
according to updates, CI should have been fixed

fromanirh:
/retest
CI is reportedly fixed, so let's retry. Worst case we have a fresh failure to report.

fromanirh:
/retest
trying the fixes

PR #437, opened 2020-11-05
Bug 1894861: tests: allow override operator namespace

fromanirh:
Allow to override the namespace name the tests wants to consume;
this is needed because when doing validation, we need to make sure to
check what was actually installed in the cluster, and there is no hard
requirement on the namespace name - only suggestions.

Additionally, change the fallback name to the one suggested in the last
docs.

### DIFF
--- a/functests/utils/consts.go
+++ b/functests/utils/consts.go
@@ -79,8 +79,6 @@ const (
 )
 
 const (
-	// PerformanceOperatorNamespace contains the name of the performance operator namespace
-	PerformanceOperatorNamespace = "openshift-performance-addon"
 	// NamespaceMachineConfigOperator contains the namespace of the machine-config-opereator
 	NamespaceMachineConfigOperator = "openshift-machine-config-operator"
 	// NamespaceTesting contains the name of the testing namespace
--- a/cluster-setup/base/performance/operator_namespace.yaml
+++ b/cluster-setup/base/performance/operator_namespace.yaml
@@ -3,5 +3,5 @@ kind: Namespace
 metadata:
   labels:
     openshift.io/cluster-monitoring: "true"
-  name: openshift-performance-addon
+  name: openshift-performance-addon-operator
 spec: {}
--- a/cluster-setup/base/performance/operator_operatorgroup.yaml
+++ b/cluster-setup/base/performance/operator_operatorgroup.yaml
@@ -2,4 +2,4 @@ apiVersion: operators.coreos.com/v1
 kind: OperatorGroup
 metadata:
   name: performance-addon-operator
-  namespace: openshift-performance-addon
+  namespace: openshift-performance-addon-operator
--- a/cluster-setup/base/performance/operator_subscription.yaml
+++ b/cluster-setup/base/performance/operator_subscription.yaml
@@ -2,7 +2,7 @@ apiVersion: operators.coreos.com/v1alpha1
 kind: Subscription
 metadata:
   name: performance-addon-operator
-  namespace: openshift-performance-addon
+  namespace: openshift-performance-addon-operator
 spec:
   channel: "4.7"
   name: performance-addon-operator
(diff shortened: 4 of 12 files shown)


fromanirh:
/hold
waiting for @fedepaol 's comments/ack

functests/utils/consts.go
```diff
@@ -24,7 +24,15 @@ var NodesSelector string
 // ProfileNotFound is true when discovery mode is enabled and no valid profile was found
 var ProfileNotFound bool
 
+// PerformanceOperatorNamespace contains the name of the performance operator namespace
+var PerformanceOperatorNamespace string
```
fedepaol, line 28: I feel like namespaces.PerformanceOperator would be a better fit for this variable.
Also, (but it's a matter of taste) I'd like to see the variable initialized with the default value and overridden if needed

fromanirh: Right, but this calls for a bigger refactoring than what I did here.
... Which is actually a great idea. Let's see what I can do.

functests/utils/consts.go
```diff
@@ -24,7 +24,15 @@ var NodesSelector string
...
+
 func init() {
+	PerformanceOperatorNamespace = os.Getenv("PERFORMANCE_OPERATOR_NAMESPACE")
```
fedepaol, line 31: I'd use `LookupEnv`

functests/utils/consts.go
fedepaol: Not that big.
The namespace package is already there, you can just move the variable and the initialization.

functests/utils/consts.go
fromanirh: no strong opinions here, hence I will do that. But what's the benefit in this specific case? Could you please elaborate for my understanding?

fedepaol:
/hold

@fromanirh should we change also the namespace we deploy the performance operator to in CI?

fromanirh:
> /hold
> 
> @fromanirh should we change also the namespace we deploy the performance operator to in CI?

I think is probably better to keep everything consistent, yes. /cc @cynepco3hahue to make sure he's aware. I can prepare the PR myself.

fedepaol:
> > /hold
> > @fromanirh should we change also the namespace we deploy the performance operator to in CI?
> 
> I think is probably better to keep everything consistent, yes. /cc @cynepco3hahue to make sure he's aware. I can prepare the PR myself.

Yep, and also: without that change, CI would start to fail

cynepco3hahue:
> I think is probably better to keep everything consistent, yes. /cc @cynepco3hahue to make sure he's aware. I can prepare the PR myself.

Works for me, I also thought to change it to be consistent with the documentation.

fromanirh:
ack. Will post the PR to fix u/s CI (link them here) and then unhold.

fromanirh:
> ack. Will post the PR to fix u/s CI (link them here) and then unhold.

this is because IIRC we need to change `openshift/release`. Let me update this PR with the changes to the manifests/scripts

fromanirh:
the upgrade lane failure looks legit, will check later

fromanirh:
/retest
now that CI healed, let's check one last time if failures are real (as they seem) or not.

fromanirh:
deleted (thus closed) by mistake

fedepaol:
/retitle bug 1894861: tests: allow override operator namespace

fedepaol:
/retitle Bug 1894861: tests: allow override operator namespace

fromanirh:
/retest
CI is reportedly fixed, so let's retry. Worst case we have a fresh failure to report.

fromanirh:
/retest
it seems CI healed (this time for real?)

fedepaol:
/override ci/prow/e2e-gcp 
This is a chicken - egg problem since in order to have the tests passing, we need https://github.com/openshift/release/pull/13545 (and viceversa).

PR #441, opened 2020-11-11
[release-4.4] Change functest run parameters

Avielyo10:
- replace `keepGoing` flag by `failFast`, our CI requires that all tests should pass,
it no taste to wait until all tests will finish to run if one of them failed

- to avoid flakiness, add `flakeAttempts=1`, so once test fail the ginkgo will re-run it once

### DIFF
--- a/hack/run-functests.sh
+++ b/hack/run-functests.sh
@@ -6,7 +6,9 @@ if [ $? -ne 0 ]; then
 	go install github.com/onsi/ginkgo/ginkgo
 fi
 
+# --failFast: ginkgo will stop the suite right after the first spec failure
+# --flakeAttempts: rerun the test if it fails
 # after running "updating profile" tests - some other functional tests might be failed, so it's better to run them separately
-GOFLAGS=-mod=vendor ginkgo functests -- -junit /tmp/artifacts/unit_report.xml -ginkgo.skip "Updating parameters in performance profile"
-GOFLAGS=-mod=vendor ginkgo functests -- -junit /tmp/artifacts/updating_profile_unit_report.xml -ginkgo.focus "Updating parameters in performance profile"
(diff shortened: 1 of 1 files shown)


fromanirh:
Are we sure this:
```
 [WARN] Deployment failed, retrying in 10 sec, 28 retries left.
[INFO] Deploying performance operator and profile.
namespace/openshift-performance-addon unchanged
machineconfigpool.machineconfiguration.openshift.io/worker-rt unchanged
operatorgroup.operators.coreos.com/openshift-performance-addon-operatorgroup unchanged
catalogsource.operators.coreos.com/performance-addon-operator-catalogsource unchanged
subscription.operators.coreos.com/performance-addon-operator-subscription unchanged
error: unable to recognize "STDIN": no matches for kind "PerformanceProfile" in version "performance.openshift.io/v1alpha1" 
```
is just a flake?

cynepco3hahue:
@fromanirh It easily can be a registry problem, the main question do we need any updates under the 4.4 branch?
IIUC we will need to release additional z-stream build only in case if we will have some new critical CVE.

fromanirh:
> @fromanirh It easily can be a registry problem, the main question do we need any updates under the 4.4 branch?
> IIUC we will need to release additional z-stream build only in case if we will have some new critical CVE.

I agree there is no clear need for this PR. OTOH I don't mind merging it if cleanly passes tests/CI. Also, better to check from time to time that 4.4 is healthy enough we can release from it.

Avielyo10:
@fromanirh @cynepco3hahue 
This is for our Jenkins pipeline tests

cynepco3hahue:
I will take a look today, from a fast glance I can see that it uses the right registry, but the OLM fails to deploy relevant things because of the timeout, so looks like the creation of the performance-profile operator failed.

PR #442, opened 2020-11-11
Tests: fix default namespace, and allow override

fedepaol:
Manual cherry pick of https://github.com/openshift-kni/performance-addon-operators/pull/437

Always use the namespaces suggested in the official documentation.

Allow to override the namespace name the tests wants to consume;
this is needed because when doing validation, we need to make sure to
check what was actually installed in the cluster, and there is no hard
requirement on the namespace name - only suggestions.

### DIFF
--- a/cluster-setup/base/performance/operator_namespace.yaml
+++ b/cluster-setup/base/performance/operator_namespace.yaml
@@ -3,5 +3,5 @@ kind: Namespace
 metadata:
   labels:
     openshift.io/cluster-monitoring: "true"
-  name: openshift-performance-addon
+  name: openshift-performance-addon-operator
 spec: {}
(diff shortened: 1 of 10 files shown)


must-gather/collection-scripts/namespace
```diff
@@ -6,6 +6,6 @@ function namespace() {
   # trying again with the pods, which are _usually_ reliable - but users can change them
   [ -z "${ns}" ] && ns=$( oc get pods -A -l name='performance-operator' -o=jsonpath='{.items[0].metadata.namespace}{"\n"}' 2> /dev/null )
   # if namespace not found, fallback to the old default
-  [ -z "${ns}" ] && ns="openshift-performance-addon"
+  [ -z "${ns}" ] && ns="openshift-performance-addon-operator"
```
fromanirh, line 9: please leave this and this alone out, becuase I prefer to use https://github.com/openshift-kni/performance-addon-operators/pull/443 to have consistent trees.

fromanirh (changes requested):
looks almost good, please check the inline comment

must-gather/collection-scripts/namespace
fedepaol: Ok, thanks

fromanirh (approved):
/lgtm

PR #446, opened 2020-11-19
csv: csv-processor: don't hardcode annotations

fromanirh:
We have support to inject custom annotations from
a JSON file. So let's use this facility in our default
flow instead of hardcoding them in the tool.

This also fixes a typo found in the hardcoded annotation.


fromanirh:
we will likely need a backport to 4.6

hack/csv-generate.sh
```diff
@@ -83,8 +83,8 @@ if [[ ${CSV_VERSION} =~ 4.7.* ]]; then
     --replaces-csv-version "${REPLACES_CSV_VERSION}" \
     --skip-range "${CSV_SKIP_RANGE}" \
     --min-kube-version "${MIN_KUBE_VERSION}" \
-    "${MAINTAINERS}" \
-    "${EXTRA_ANNOTATIONS}"
+    "${EXTRA_ANNOTATIONS}" \
```
marcel-apf, line 86: You changed the order, in purpose?

fromanirh: yes, and also renamed. The order is not really relevant here. But I can split to a different commit if needed.

marcel-apf: No need, thanks

fromanirh:
/hold
I need Yanir's lgtm for awareness

PR #448, opened 2020-11-24
Replace tools base image from centos to ubi

cynepco3hahue:
Docker hub added rate limit that prevents from us to build
the tools image. To avoid the problem we will use the
ubi image from the redhat registry.


fromanirh:
/approve
/lgtm
/hold
looks good! if CI likes it, we can merge

cynepco3hahue:
in general, it will not merge without CI :) so I will unhold it
/hold cancel

PR #449, opened 2020-11-24
dockerfile: add ping-helper image

fromanirh:
we want a dead-simple, minimal, trivial to maintain
image which we can use for connectivity tests
in our and openshift e2e tests.


MarSik:
A good idea with an image like this is to make it so it supports this usage pattern:

oc run helper ping 8.8.8.8

Basically so arguments are directly executed in the shell.

fromanirh:
> A good idea with an image like this is to make it so it supports this usage pattern:
> 
> podman run helper ping 8.8.8.8
> 
> Basically so arguments are directly executed in the shell.

good point! fixed and added usage example in the commit message

marcel-apf:
@fromanirh what is the status of this PR? Do we need it?

fromanirh:
not needed for now, we will re-evaluate later.

PR #450, opened 2020-11-24
machineconfig: Set RPS mask for both containers and host

marcel-apf:
Add a pre-start OCI hook setting the CPUs RPS mask to the performance profile's reserved CPUs.

To-do:
- Some tests

### DIFF
--- a/build/assets/scripts/low-latency-hooks.sh
+++ b/build/assets/scripts/low-latency-hooks.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+mask="${1}"
+[ -n "${mask}" ] || { logger "${0}: The rps-mask parameter is missing" ; exit 0; }
+
+pid=$(jq '.pid' /dev/stdin 2>&1)
+[[ $? -eq 0 && -n "${pid}" ]] || { logger "${0}: Failed to extract the pid: ${pid}"; exit 0; }
+
+ns=$(ip netns identify "${pid}" 2>&1)
+[[ $? -eq 0 && -n "${ns}" ]] || { logger "${0} Failed to identify the namespace: ${ns}"; exit 0; }
+
+mode=$(ip netns exec "${ns}" [ -w /sys ] && echo "rw" || echo "ro" 2>&1)
+[ $? -eq 0 ] || { logger "${0} Failed to determine if the /sys is writable: ${mode}"; exit 0; }
+
+if [ "${mode}" = "ro" ]; then
+    res=$(ip netns exec "${ns}" mount -o remount,rw /sys 2>&1)
+    [ $? -eq 0 ] || { logger "${0}: Failed to remount /sys as rw: ${res}"; exit 0; }
+fi
+
+# /sys/class/net can't be used recursively to find the rps_cpus file, use /sys/devices instead
+res=$(ip netns exec "${ns}" find /sys/devices -type f -name rps_cpus -exec sh -c "echo ${mask} | cat > {}" \; 2>&1)
+[[ $? -eq 0 && -z "${res}" ]] || logger "${0}: Failed to apply the RPS mask: ${res}"
+
+if [ "${mode}" = "ro" ]; then
+    ip netns exec "${ns}" mount -o remount,ro /sys
+    [ $? -eq 0 ] || exit 1 # Error out so the pod will not start with a writable /sys
+fi
--- a/pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -43,6 +43,8 @@ const (
 	ociHooksConfigDir   = "/etc/containers/oci/hooks.d"
 	ociHooksConfig      = "99-low-latency-hooks"
 	ociTemplateRPSMask  = "RPSMask"
+	udevRulesDir        = "/etc/udev/rules.d"
+	udevRpsRule         = "99-netdev-rps"
 )
 
 const (
@@ -158,6 +160,18 @@ func getIgnitionConfig(assetsDir string, profile *performancev2.PerformanceProfi
 		return nil, err
 	}
 
+	// add rps udev rule
+	rpsRuleMode := 0644
+	rule := fmt.Sprintf("%s.rules", udevRpsRule)
+	if err := addFile(
+		ignitionConfig,
+		filepath.Join(assetsDir, "configs", rule),
+		filepath.Join(udevRulesDir, rule),
+		&rpsRuleMode,
+	); err != nil {
+		return nil, err
+	}
+
 	if profile.Spec.HugePages != nil {
 		for _, page := range profile.Spec.HugePages.Pages {
 			// we already allocated non NUMA specific hugepages via kernel arguments
@@ -185,6 +199,23 @@ func getIgnitionConfig(assetsDir string, profile *performancev2.PerformanceProfi
 				Name:     getSystemdService(fmt.Sprintf("%s-%skB-NUMA%d", hugepagesAllocation, hugepagesSize, *page.Node)),
 			})
 		}
+	}
+
+	if profile.Spec.CPU != nil && profile.Spec.CPU.Reserved != nil {
+		rpsMask, err := components.CPUListToMaskList(string(*profile.Spec.CPU.Reserved))
+		if err != nil {
+			return nil, err
+		}
+
+		rpsService, err := getSystemdContent(getRPSUnitOptions(rpsMask))
+		if err != nil {
+			return nil, err
+		}
+
+		ignitionConfig.Systemd.Units = append(ignitionConfig.Systemd.Units, igntypes.Unit{
+			Contents: rpsService,
+			Name:     getSystemdService("update-rps@"),
+		})
 	}
 
 	return ignitionConfig, nil
@@ -269,6 +300,20 @@ func getHugepagesAllocationUnitOptions(hugepagesSize string, hugepagesCount int3
 		// [Install]
 		// WantedBy
 		unit.NewUnitOption(systemdSectionInstall, systemdWantedBy, systemdTargetMultiUser),
+	}
+}
+
+func getRPSUnitOptions(rpsMask string) []*unit.UnitOption {
+	cmd := fmt.Sprintf("/bin/find /sys/class/net/%%i/queues -type f -name rps_cpus -exec sh -c \"echo %s | cat > {}\" \\;", rpsMask)
+	return []*unit.UnitOption{
+		// [Unit]
+		// Description
+		unit.NewUnitOption(systemdSectionUnit, systemdDescription, "Sets network devices RPS mask"),
+		// [Service]
+		// Type
+		unit.NewUnitOption(systemdSectionService, systemdType, systemdServiceTypeOneshot),
+		// ExecStart
+		unit.NewUnitOption(systemdSectionService, systemdExecStart, cmd),
 	}
 }
 
--- a/pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -1,11 +1,13 @@
 package machineconfig
 
 import (
+	"bytes"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"text/template"
 
 	"github.com/coreos/go-systemd/unit"
 	igntypes "github.com/coreos/ignition/config/v2_2/types"
@@ -37,6 +39,10 @@ const (
 	bashScriptsDir      = "/usr/local/bin"
 	crioConfd           = "/etc/crio/crio.conf.d"
 	crioRuntimesConfig  = "99-runtimes"
+	ociHooks            = "low-latency-hooks"
+	ociHooksConfigDir   = "/etc/containers/oci/hooks.d"
+	ociHooksConfig      = "99-low-latency-hooks"
+	ociTemplateRPSMask  = "RPSMask"
 )
 
 const (
@@ -116,7 +122,7 @@ func getIgnitionConfig(assetsDir string, profile *performancev2.PerformanceProfi
 
 	// add script files under the node /usr/local/bin directory
 	mode := 0700
-	for _, script := range []string{hugepagesAllocation} {
+	for _, script := range []string{hugepagesAllocation, ociHooks} {
 		src := filepath.Join(assetsDir, "scripts", fmt.Sprintf("%s.sh", script))
 		if err := addFile(ignitionConfig, src, getBashScriptPath(script), &mode); err != nil {
 			return nil, err
@@ -131,6 +137,23 @@ func getIgnitionConfig(assetsDir string, profile *performancev2.PerformanceProfi
 		filepath.Join(assetsDir, "configs", config),
 		filepath.Join(crioConfd, config),
 		&crioConfdRuntimesMode,
+	); err != nil {
+		return nil, err
+	}
+
+	// add crio hooks config  under the node cri-o hook directory
+	crioHooksConfigsMode := 0644
+	config = fmt.Sprintf("%s.json", ociHooksConfig)
+	ociHooksConfigContent, err := GetOCIHooksConfigContent(filepath.Join(assetsDir, "configs", config), profile)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := addContent(
+		ignitionConfig,
+		ociHooksConfigContent,
+		filepath.Join(ociHooksConfigDir, config),
+		&crioHooksConfigsMode,
 	); err != nil {
 		return nil, err
 	}
@@ -188,6 +211,31 @@ func getSystemdContent(options []*unit.UnitOption) (string, error) {
 	return string(outBytes), nil
 }
 
+// GetOCIHooksConfigContent reads and returns the content of the OCI hook file
+func GetOCIHooksConfigContent(configFile string, profile *performancev2.PerformanceProfile) ([]byte, error) {
+	content, err := ioutil.ReadFile(configFile)
+	if err != nil {
+		return nil, err
+	}
+
+	rpsMask := "0" // RPS disabled
+	if profile.Spec.CPU != nil && profile.Spec.CPU.Reserved != nil {
+		rpsMask, err = components.CPUListToMaskList(string(*profile.Spec.CPU.Reserved))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	outContent := &bytes.Buffer{}
+	templateArgs := map[string]string{ociTemplateRPSMask: rpsMask}
+	template := template.Must(template.New("crio").Parse(string(content)))
+	if err := template.Execute(outContent, templateArgs); err != nil {
+		return nil, err
+	}
+
+	return outContent.Bytes(), nil
+}
+
 // GetHugepagesSizeKilobytes retruns hugepages size in kilobytes
 func GetHugepagesSizeKilobytes(hugepagesSize performancev2.HugePageSize) (string, error) {
 	switch hugepagesSize {
@@ -224,11 +272,7 @@ func getHugepagesAllocationUnitOptions(hugepagesSize string, hugepagesCount int3
 	}
 }
 
-func addFile(ignitionConfig *igntypes.Config, src string, dst string, mode *int) error {
-	content, err := ioutil.ReadFile(src)
-	if err != nil {
-		return err
-	}
+func addContent(ignitionConfig *igntypes.Config, content []byte, dst string, mode *int) error {
 	contentBase64 := base64.StdEncoding.EncodeToString(content)
 	ignitionConfig.Storage.Files = append(ignitionConfig.Storage.Files, igntypes.File{
 		Node: igntypes.Node{
@@ -243,4 +287,12 @@ func addFile(ignitionConfig *igntypes.Config, src string, dst string, mode *int)
 		},
 	})
 	return nil
+}
+
+func addFile(ignitionConfig *igntypes.Config, src string, dst string, mode *int) error {
+	content, err := ioutil.ReadFile(src)
+	if err != nil {
+		return err
+	}
+	return addContent(ignitionConfig, content, dst, mode)
 }
--- a/build/assets/configs/99-low-latency-hooks.json
+++ b/build/assets/configs/99-low-latency-hooks.json
@@ -1,0 +1,11 @@
+{
+  "version": "1.0.0",
+  "hook": {
+    "path": "/usr/local/bin/low-latency-hooks.sh",
+    "args": ["low-latency-hooks.sh", "{{.RPSMask}}"]
+  },
+  "when": {
+    "always": true
+  },
+  "stages": ["prestart"]
+}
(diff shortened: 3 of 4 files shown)


build/assets/configs/99-high-performance-hooks.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": "1.0.0",
+  "hook": {
+    "path": "/usr/local/bin/high-performance-hooks.sh",
+    "args": ["high-performance-hooks.sh", "{{.RpsMask}}"]
```
cynepco3hahue, line 5: probably all occurrences of high-performance should be renamed to be low-latency, I remember @MarSik asked for the same thing for me

build/assets/scripts/high-performance-hooks.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+read -r in
+
+pid=$(sed -n "s/^.*\"pid\":\([0-9]*\),.*/\1/p" <<< "$in")
+mask=$1
+ns=$(ip netns identify "${pid}")
+
+[[ -n "${pid}"  ]] && [[ -n "${mask}"  ]] &&  [[ -n "${ns}" ]] || exit 0
```
cynepco3hahue, line 9: `[[ -n "${mask}"  ]] &&  [[ -n "${ns}" ]]` two spaces after &&

cynepco3hahue (commented):
Except some small comments looks good for now

build/assets/configs/99-high-performance-hooks.json
marcel-apf: @MarSik please ack, I don;t care one way or another :)

build/assets/scripts/high-performance-hooks.sh
marcel-apf: thanks, changed

build/assets/scripts/high-performance-hooks.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+pid=$(jq '.pid' /dev/stdin)
+mask=$1
+ns=$(ip netns identify "${pid}")
+
+[[ -n "${pid}" ]] && [[ -n "${mask}" ]] && [[ -n "${ns}" ]] || exit 0
+
+ip netns exec "${ns}" mount -o remount,rw /sys
```
fromanirh, line 9: If this fails there is no point in continuing. OTOH if we do we should "only" get a bunch of errors somewhere

build/assets/scripts/high-performance-hooks.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+pid=$(jq '.pid' /dev/stdin)
+mask=$1
+ns=$(ip netns identify "${pid}")
+
+[[ -n "${pid}" ]] && [[ -n "${mask}" ]] && [[ -n "${ns}" ]] || exit 0
```
fromanirh, line 7: is it worth to exit early if mask exists and it is equal to "0" (= rps disabled)?

build/assets/scripts/high-performance-hooks.sh
```diff
@@ -0,0 +1,11 @@
...
+
+ip netns exec "${ns}" mount -o remount,rw /sys
+ip netns exec "${ns}" find /sys/devices -type f -name rps_cpus -exec sh -c "echo ${mask} | cat > {}" \;
+ip netns exec "${ns}" mount -o remount,ro /sys
```
fromanirh, line 11: I wonder if it makes sense to split this script in two, one which identifies the NS and executes the real script in the right namespace, and another one which just does the work. Something like

```
$ cat assets/scripts/run-in-ns.sh
#!/usr/bin/env bash

pid=$(jq '.pid' /dev/stdin)
mask=$1
what=$2
ns=$(ip netns identify "${pid}")

[[ -n "${pid}" ]] && [[ -n "${mask}" ]] && [[ -n "${ns}" ]] || exit 0

ip netns exec "${ns}" ${what}
```
and
```
$ cat assets/scripts/high-performace-hooks.sh
#!/usr/bin/env bash
mount -o remount,rw /sys
find /sys/devices -type f -name rps_cpus -exec sh -c "echo ${mask} | cat > {}" \;
mount -o remount,ro /sys
```
Whatever's easier to write and maintain, I guess.

build/assets/scripts/high-performance-hooks.sh
marcel-apf: yep, I better bail out [no yet though about tightening things up]

build/assets/scripts/high-performance-hooks.sh
marcel-apf: in the future we may want to disable rps... and the default value is ) anyway, I would leave it as is

fromanirh: works for me!

build/assets/scripts/high-performance-hooks.sh
marcel-apf: I will add a "set -e" to the script to bail out if anything goes wrong, would be good enough?

fromanirh: using `set -e` is a good idea in general, but it seems our case it's a bit more complex.
let's consider the last three commands, which is the real deal here.
1. if remounting sysfs rw fails, we do want to bail out (but keep doing will do no harm besides wasted work)
2. if find fails, we still want to remount sysfs ro!
3. if remounting sysfs ro fails, we have no way to recover anyway so nothing to do here

marcel-apf: right, I'll be more precise

build/assets/scripts/high-performance-hooks.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+pid=$(jq '.pid' /dev/stdin)
+mask=$1
+ns=$(ip netns identify "${pid}")
+
+[[ -n "${pid}" ]] && [[ -n "${mask}" ]] && [[ -n "${ns}" ]] || exit 0
+
+ip netns exec "${ns}" mount -o remount,rw /sys || exit 0
```
fromanirh, line 9: nice!

marcel-apf:
Addressed @fromanirh;s comments
Didn't rename yet the files to low-latency-..., waiting for @MarSik ack
Add support for dynamic host network device rps updates

build/assets/scripts/high-performance-hooks.sh
marcel-apf: :)

build/assets/scripts/high-performance-hooks.sh
```diff
@@ -0,0 +1,11 @@
...
+ip netns exec "${ns}" find /sys/devices -type f -name rps_cpus -exec sh -c "echo ${mask} | cat > {}" \;
```
fromanirh, line 10: any reason to use a different find here with respect to https://github.com/openshift-kni/performance-addon-operators/pull/450/commits/bde1fc3aa43f0b62e1c72259d47ab9c6d24b663f#diff-b550247b819f015b31bacae7c601e2ef1045d94b6ff48ecb29c3cb362578be75R306 ?

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -43,6 +43,8 @@ const (
 	crioHooksConfigDir  = "/etc/containers/oci/hooks.d"
 	crioHooksConfig     = "99-high-performance-hooks"
 	crioTemplateRpsMask = "RpsMask"
+	udevRulesDir        = "/etc/udev/rules.d"
+	udevRpsRule         = "99-netdev-rps"
```
fromanirh, line 47: this seems to be used just once in line 165 below. If so why can't we just call this "99-netdev-rps.rules"  ?

fromanirh (commented):
very nice improvements, kudos!

build/assets/scripts/high-performance-hooks.sh
marcel-apf: funny thing, find doesn't find the rps_cpus files  in /sys/class/net/ dir.
It does find them in  /sys/class/net/vethxyz/ or in /sys/devices/

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
marcel-apf: I have this weird rule to use the same coding standard in the file, especially the micro patterns, so readers would not wonder why is different. I agree there is no other need for that.
I am OK one way or the other.

build/assets/scripts/high-performance-hooks.sh
marcel-apf: is something about an implementation difference of the sys/class vs /sys/devices

cynepco3hahue, line 10: After the @fromanirh comment regarding error handling, IMHO we should add -e and add a trap that will call to `ip netns exec "${ns}" mount -o remount,ro /sys`, because if the command
`ip netns exec "${ns}" find /sys/devices -type f -name rps_cpus -exec sh -c "echo ${mask} | cat > {}" \;` will fail but the remount command after it will succeed, the hook will exit with 0 code, and from the cri-o perspective it will mean that hook succeeded, so a user can get unexpected results

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -224,11 +302,21 @@ func getHugepagesAllocationUnitOptions(hugepagesSize string, hugepagesCount int3
 	}
 }
 
-func addFile(ignitionConfig *igntypes.Config, src string, dst string, mode *int) error {
-	content, err := ioutil.ReadFile(src)
-	if err != nil {
-		return err
+func getRpsUnitOptions(rpsMask string) []*unit.UnitOption {
```
cynepco3hahue, line 305: should this service start before the kubelet service?

cynepco3hahue, line 305: nit: use upper case letters for acronyms - getRPSUnitOptions

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -188,6 +242,30 @@ func getSystemdContent(options []*unit.UnitOption) (string, error) {
 	return string(outBytes), nil
 }
 
+func getCrioHooksConfigContent(configFile string, profile *performancev2.PerformanceProfile) ([]byte, error) {
```
cynepco3hahue, line 245: nit: use upper case letters for acronyms - getCRIOHooksConfigContent

build/assets/scripts/high-performance-hooks.sh
marcel-apf: This is on purpose, I am interested to return as less errors as possible, if the hook fails I want the pod to continue to run.  I prefer the pod to be running with the rps mask not set, than not running at all

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
marcel-apf: The server is only triggered by the udev rule

marcel-apf: will change, thanks

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
marcel-apf: Missed that one, thanks

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
fromanirh: I'm ok following the existing conventions, so we can keep it this way.

build/assets/scripts/high-performance-hooks.sh
fromanirh: ok, no problem. Worth a comment too. Others (or our future selves) may well wonder the same thing in the future.

marcel-apf: Will add a comment in the script, thanks

marcel-apf:
Fixes:
 - Rename files to low-latency... (@cynepco3hahue )
 - Correct naming for functions (@cynepco3hahue )
 - Add a needed comment in script (@fromanirh )

marcel-apf:
Fixed a minor typo in a script comment

build/assets/configs/99-high-performance-hooks.json
marcel-apf: changed, thanks

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
marcel-apf: (Is not connected to the kubelet at all)

cynepco3hahue: I thought it should configure all masks before starting all containers?

build/assets/scripts/high-performance-hooks.sh
cynepco3hahue: hm if you expect some functionality and you do not get it silently it can lead to unexpected behavior, but I will not block on it

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
marcel-apf: well, this is a new approach ending in the same way.
The udev will trigger an event for each new net device, so is enough to handle only the device.

The previous approach failed as it didn't take into account other net devices created after the container started.


This approach has 2 steps
1. Set the rps mask for each pod when the pod is started
2. Set the rps mask for each host net device as soon as is created

build/assets/scripts/high-performance-hooks.sh
marcel-apf: yeah, is a hack. we will push for a sane RPS interface, sadly is the best it can be done at current time.

fromanirh: The question is: even if we properly exit with error and we log the reason, where do the log ends and how we can check it? If it ends up somewhere we can check (crio logs?) it's worth improving. Even something trivial like well defined exit codes for the failure conditions can be enough.

fromanirh: In general, what happens if a hook exits with error?

marcel-apf: The point is we don't actually want to error out. We can use the Logger to send the logs to the syslog, but is not necessary since is the caller's responsibility to catch the stderr of the script and log it.

The current approach does the best it can:
1. if it can't remount the sysfs as rw, exits with 0 so it will let the pod continue to run, it makes no sense to punish some random pod because of this.
2. At this point we go for the rps_cpus files write, even if it doesn't succeed, we still need to try to unmount and the pod should still continue to run.
3. The last try to remount as ro is tricky, we shouldn't let an unprivileged pod to run with sysfs rw. But if so, since is the last script line, it will return a non-zero error code and the pod will not run.

All of the above is (should be) logged by the caller.

fromanirh: > The point is we don't actually want to error out. We can use the Logger to send the logs to the syslog, but is not necessary since is the caller's responsibility to catch the stderr of the script and log it.

True. Do we know where do these logs end? crio logs?

> The current approach does the best it can:
> 
>     1. if it can't remount the sysfs as rw, exits with 0 so it will let the pod continue to run, it makes no sense to punish some random pod because of this.

Well, this is debatable, and that's indeed why we are discussing. Guaranteed pods which opt in to the behaviour we are implementing here expect some preconditions. If we fail to set up the RPS mask these preconditions aren't guaranteed anymore. At this point keep going with the pod start up may or may be not the right thing to do but what @cynepco3hahue AFAIU raised (and if so I totally agree) is that we should never *silently* fail to comply with the pod expectations.

So if the hook cannot properly set up the environment the guarantees aren't met anymore and this fact should be evident somewhere. Which is indeed responsability of the caller (crio) to log, but we need to know where this happens and how we can check later on.

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -224,11 +302,21 @@ func getHugepagesAllocationUnitOptions(hugepagesSize string, hugepagesCount int3
...
+	cmd := fmt.Sprintf("/bin/find /sys/class/net/%%i/queues -type f -name rps_cpus -exec sh -c \"echo %s | cat > {}\" \\;", rpsMask)
+	return []*unit.UnitOption{
+		// [Unit]
+		// Description
+		unit.NewUnitOption(systemdSectionUnit, systemdDescription, "Sets network devices RPS mask"),
+		// [Service]
+		// Type
+		unit.NewUnitOption(systemdSectionService, systemdType, systemdServiceTypeOneshot),
+		// ExecStart
+		unit.NewUnitOption(systemdSectionService, systemdExecStart, cmd),
```
MarSik, line 315: Make sure the unit is executed in a deterministic order. Eg. use After and Before as needed (After=network, Before=kubelet? <-- not sure about this one).

build/assets/scripts/low-latency-hooks.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
MarSik, line 1: This is executed for all the pods, not just guaranteed ones, right?

build/assets/scripts/high-performance-hooks.sh
MarSik: But this hook is called for all the pods in the system, right? I see two issues:

1) Last time we used cri-o code approach instead of wrapper, because the OCP maintainers were worried about scalability (what happens when huge amount of pods starts at the same time). This we can't probably fix atm.
2) A failing hook prevents the start of pods, we must be super careful to not compromise the whole node

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
MarSik: Oh, isn't the service also started once on boot?

build/assets/scripts/high-performance-hooks.sh
marcel-apf: > In general, what happens if a hook exits with error?

The container ends up in status: CreateContainerError
And in the pod status we see the exact error:
  Warning  Failed          1s               kubelet, ocp47-worker-1.demo.lab  Error: container create failed: time="2020-11-26T08:26:48Z" level=error msg="container_linux.go:366: starting container process caused: process_linux.go:472: container init caused: Running hook #0:: error running hook: exit status 1, stdout: , stderr: /usr/local/bin/high-performance-hooks.sh: line 13: /bin/d: Read-only file system\n"

marcel-apf: > True. Do we know where do these logs end? crio logs?
> 

The pod event status has the detailed output, please see the above exampe

fromanirh: Thanks, container ending up in `CreateContainerError` matches my expectations and it's indeed better than silently fail to meet the expectations.

marcel-apf: > Well, this is debatable, and that's indeed why we are discussing. Guaranteed pods which opt in to the behaviour we are implementing here expect some preconditions. If we fail to set up the RPS mask these preconditions aren't guaranteed anymore. At this point keep going with the pod start up may or may be not the right thing to do but what @cynepco3hahue AFAIU raised (and if so I totally agree) is that we should never _silently_ fail to comply with the pod expectations.
> 

We can opt to return error in this case, that means is possible no container would start and the node will appear as "Not Ready". But if the consensus is that is a critical matter I will change it to return error.


> So if the hook cannot properly set up the environment the guarantees aren't met anymore and this fact should be evident somewhere. Which is indeed responsability of the caller (crio) to log, but we need to know where this happens and how we can check later on.

If the script is returning with exit 0, no interim error is logged and sensing something to syslog does not work either.

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
marcel-apf: No, the service is not started on boot, is started only by udev on net dev creation.

build/assets/scripts/high-performance-hooks.sh
marcel-apf: > But this hook is called for all the pods in the system, right? I see two issues:
> 
> 1. Last time we used cri-o code approach instead of wrapper, because the OCP maintainers were worried about scalability (what happens when huge amount of pods starts at the same time). This we can't probably fix atm.

Nope, we can't, it will be a short delay on each pod startup. The "price" to pay for the low-latency.
But, as you pointed out, the systems requiring RT are quite static, running a few pods.

> 2. A failing hook prevents the start of pods, we must be super careful to not compromise the whole node
Exactly what I said! 

We have 2 points of view and we need a decision:
If the pod hook fails to do its job it will affect the Guaranteed Pods latency.
So what should we do?
 - Deny it to run can and will compromise the node, for example it may not allow an operator pod to run. 
 - Allow it to run anyway will compromise the Guaranteed Pods latency.

build/assets/scripts/low-latency-hooks.sh
marcel-apf: Right

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
marcel-apf: It does not run on system boot, the dependencies are not relevant.

build/assets/scripts/high-performance-hooks.sh
marcel-apf: > Thanks, container ending up in `CreateContainerError` matches my expectations and it's indeed better than silently fail to meet the expectations.

Right, we don't have a security risk.

Our Achilles' heel remains if the writing the rps_cpus fails and the remount to ro succeeded. In this case the hook returns with 0. I found no way to log something if the hook succeeded

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
MarSik: Ack, I did not know about systemd unit templates feature. A nifty trick!

build/assets/scripts/low-latency-hooks.sh
MarSik: We need a scale test.. starting 250 pods at the same time and such.

marcel-apf:
Renamed "crioHools" -> ociHooks

marcel-apf:
Improved the low-latency-hooks.sh script:
 - check the input
 - add logs for every step
 - handle the  case when the container has /sys rw already (fx bug that returned it to rw)

build/assets/scripts/low-latency-hooks.sh
```diff
@@ -0,0 +1,27 @@
...
+fi
+
+# /sys/class/net can't be used recursively to find the rps_cpus file, use /sys/devices instead
+res=$(ip netns exec "${ns}" find /sys/devices -type f -name rps_cpus -exec sh -c "echo ${mask} | cat > {}" \; 2>&1)
+[[ $? -eq 0 && -z "${res}" ]] || logger "${0}: Failed to apply the RPS mask: ${res}"
+
+if [ "${mode}" = "ro" ]; then
+    ip netns exec "${ns}" mount -o remount,ro /sys
+    [ $? -eq 0 ] || exit 1 # Error out so the pod will not start with a writable /sys
+fi
```
fromanirh, line 27: this looks *very* nice and well thought, kudos.

cynepco3hahue:
@fromanirh @MarSik Can you please review again and unhold once the PR looks good to you.

fromanirh (approved):
/lgtm
/hold cancel

PR #451, opened 2020-11-24
Add test verifying that IRQ balance can be disabled per POD

dshchedr:
Also combined some duplicated code in cpu_management.go file

### DIFF
--- a/functests/1_performance/cpu_management.go
+++ b/functests/1_performance/cpu_management.go
@@ -31,9 +31,10 @@ import (
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
 )
 
+var workerRTNode *corev1.Node
+var profile *performancev2.PerformanceProfile
+
 var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
-	var workerRTNode *corev1.Node
-	var profile *performancev2.PerformanceProfile
 	var balanceIsolated bool
 	var reservedCPU, isolatedCPU string
 	var listReservedCPU, listIsolatedCPU []int
@@ -170,11 +171,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 		})
 
 		AfterEach(func() {
-			err := testclient.Client.Delete(context.TODO(), testpod)
-			Expect(err).ToNot(HaveOccurred())
-
-			err = pods.WaitForDeletion(testpod, 60*time.Second)
-			Expect(err).ToNot(HaveOccurred())
+			deleteTestPod(testpod)
 		})
 
 		table.DescribeTable("Verify CPU usage by stress PODs", func(guaranteed bool) {
@@ -254,45 +251,14 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 			defaultFlags, err = getCPUSchedulingDomainFlags(0)
 			Expect(err).ToNot(HaveOccurred())
 
-			testpod = pods.GetTestPod()
-			testpod.Annotations = map[string]string{
+			annotations := map[string]string{
 				"cpu-load-balancing.crio.io": "disable",
 			}
-			testpod.Namespace = testutils.NamespaceTesting
-
-			cpus := resource.MustParse("1")
-			memory := resource.MustParse("256Mi")
-
-			// change pod resource requirements, to change the pod QoS class to guaranteed
-			testpod.Spec.Containers[0].Resources = corev1.ResourceRequirements{
-				Limits: corev1.ResourceList{
-					corev1.ResourceCPU:    cpus,
-					corev1.ResourceMemory: memory,
-				},
-			}
-
-			// use the CPU load balancing runtime class
-			runtimeClassName := components.GetComponentName(profile.Name, components.ComponentNamePrefix)
-			testpod.Spec.RuntimeClassName = &runtimeClassName
-			testpod.Spec.NodeSelector = map[string]string{testutils.LabelHostname: workerRTNode.Name}
+			testpod = getTestPodWithAnnotations(annotations)
 		})
 
 		AfterEach(func() {
-			// it possible that the pod already was deleted as part of the test, in this case we want to skip teardown
-			key := types.NamespacedName{
-				Name:      testpod.Name,
-				Namespace: testpod.Namespace,
-			}
-			err := testclient.Client.Get(context.TODO(), key, testpod)
-			if errors.IsNotFound(err) {
-				return
-			}
-
-			err = testclient.Client.Delete(context.TODO(), testpod)
-			Expect(err).ToNot(HaveOccurred())
-
-			err = pods.WaitForDeletion(testpod, 60*time.Second)
-			Expect(err).ToNot(HaveOccurred())
+			deleteTestPod(testpod)
 		})
 
 		It("[test_id:32646] should disable CPU load balancing for CPU's used by the pod", func() {
@@ -331,11 +297,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 			}
 
 			By("Deleting the pod")
-			err = testclient.Client.Delete(context.TODO(), testpod)
-			Expect(err).ToNot(HaveOccurred())
-
-			err = pods.WaitForDeletion(testpod, 60*time.Second)
-			Expect(err).ToNot(HaveOccurred())
+			deleteTestPod(testpod)
 
 			By("Getting the CPU scheduling flags")
 			flags, err = getCPUSchedulingDomainFlags(cpu)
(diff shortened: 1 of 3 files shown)


functests/utils/nodes/nodes.go
```diff
@@ -201,3 +201,31 @@ func BannedCPUs(node corev1.Node) (banned cpuset.CPUSet, err error) {
...
+}
+
+// GetOnlineCPUsSet returns the list of online (being scheduled) CPUs on the node
+func GetOnlineCPUsSet(node *corev1.Node) (cpuset.CPUSet, error) {
+	command := []string{"cat", "/sys/devices/system/cpu/online"}
+	onlineCPUs, err := ExecCommandOnNode(command, node)
+	if err != nil {
+		return cpuset.NewCPUSet(), err
+	}
+	onlineCPUsSet, err := cpuset.Parse(onlineCPUs)
```
marcel-apf, line 226: You can just return the output of cpuset.Parse

functests/utils/nodes/nodes.go
```diff
@@ -201,3 +201,31 @@ func BannedCPUs(node corev1.Node) (banned cpuset.CPUSet, err error) {
...
 }
+
+// GetDefaultSmpAffinitySet returns the default smp affinity mask for the node
+func GetDefaultSmpAffinitySet(node *corev1.Node) (cpuset.CPUSet, error) {
+	command := []string{"cat", "/proc/irq/default_smp_affinity"}
+	defaultSmpAffinity, err := ExecCommandOnNode(command, node)
+	if err != nil {
+		return cpuset.NewCPUSet(), err
+	}
+	defaultSmpAffinitySet, err := components.CPUMaskToCPUSet(defaultSmpAffinity)
```
marcel-apf, line 212: You can just return the output of components.CPUMaskToCPUSet

functests/1_performance/cpu_management.go
```diff
@@ -380,3 +412,44 @@ func getStressPod(nodeName string) *corev1.Pod {
...
+	}
+	err := testclient.Client.Get(context.TODO(), key, testpod)
+	if errors.IsNotFound(err) {
+		return
+	}
+
+	err = testclient.Client.Delete(context.TODO(), testpod)
+	Expect(err).ToNot(HaveOccurred())
+
+	err = pods.WaitForDeletion(testpod, 60*time.Second)
```
marcel-apf, line 453: It happened to me I had to wait more :(, maybe my cluster is really slow)

functests/1_performance/cpu_management.go
```diff
@@ -349,6 +311,76 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+			By("Running pod with annotations that disable specific CPU from IRQ balancer")
+			annotations := map[string]string{
+				"irq-load-balancing.crio.io": "true",
+				"cpu-quota.crio.io":          "true",
+			}
+			testpod = getTestPodWithAnnotations(annotations)
+
+			err = testclient.Client.Create(context.TODO(), testpod)
+			Expect(err).ToNot(HaveOccurred())
+			err = pods.WaitForCondition(testpod, corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
```
marcel-apf, line 347: So up to 10 minutes to create the pod, but up to  1 minute to delete it?

functests/1_performance/cpu_management.go
```diff
@@ -349,6 +311,76 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			runTop := []string{"/bin/bash", "-c", "tmux new -d 'top -b > /dev/null'"}
```
marcel-apf, line 349: Can you please why tmux is needed?

functests/1_performance/cpu_management.go
```diff
@@ -349,6 +311,76 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+			defaultSmpAffinitySet, err = nodes.GetDefaultSmpAffinitySet(workerRTNode)
+			Expect(err).ToNot(HaveOccurred())
+
+			getPsr := []string{"/bin/bash", "-c", "ps -o psr $(pgrep top) | tail -1"}
+			psr, err := nodes.ExecCommandOnNode(getPsr, workerRTNode)
+			Expect(err).ToNot(HaveOccurred())
+			psrSet, err := cpuset.Parse(strings.Trim(psr, " "))
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(defaultSmpAffinitySet).To(Equal(onlineCPUsSet.Difference(psrSet)), fmt.Sprintf("Default SMP affinity should not contain isolated CPU %s", psr))
```
marcel-apf, line 363: Is there any other way to get the cpus assigned to a pod other than running some process inside?
Running  sudo crictl inspectp <podid> will return a json with a "cpu": {  "shares": 2 } construct which seems to be the assigned cpus

functests/1_performance/cpu_management.go
```diff
@@ -349,6 +311,76 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+
+			By("Checking that there are no any active IRQ on isolated CPU")
+			// It may takes some time for the system to reschedule active IRQs
+			Eventually(func() string {
+				getActiveIrq := []string{"/bin/bash", "-c", "for n in $(find /proc/irq/ -name smp_affinity_list); do echo $(cat $n); done"}
+				activeIrqList, err := nodes.ExecCommandOnNode(getActiveIrq, workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+				re := regexp.MustCompile(fmt.Sprintf(`(?m)^%s$`, psrSet))
+				return re.FindString(activeIrqList)
```
marcel-apf, line 372: Nice! have you seen this fail for a pod without the annotations?

functests/1_performance/cpu_management.go
```diff
@@ -380,3 +412,44 @@ func getStressPod(nodeName string) *corev1.Pod {
 		},
 	}
 }
+
+func getTestPodWithAnnotations(annotations map[string]string) *corev1.Pod {
```
gsr-shanks, line 416: Should this be in utils/pods ?

functests/1_performance/cpu_management.go
dshchedr: It's existing logic from other test.. Creating pod takes more time (for example for downloading image)

functests/1_performance/cpu_management.go
dshchedr: tmux allows us to run top and exit from console - same logic was used for hugepages,go tests

functests/1_performance/cpu_management.go
dshchedr: yes, here the output for the pod without annitation:


```
    Timed out after 30.000s.
    IRQ still active on CPU  1
    Expected
        <string>: 1
    to be empty

```

functests/1_performance/cpu_management.go
dshchedr: Hm, I don't think so.. here we creates specific pod used only in these tests..

functests/1_performance/cpu_management.go
dshchedr: It is existing logic from other tests, not sure we need to update it (at least for now)

functests/utils/nodes/nodes.go
dshchedr: ah, right! thanks

functests/utils/nodes/nodes.go
dshchedr: right, thanks!

functests/1_performance/cpu_management.go
dshchedr: hm, I think "crictl inspectp" does not show the right cpu, I have 2 pods running and both of them has same "shares":

[root@worker-0 ~]# crictl inspectp 7ed0a6a3b6a4b | grep shares
            "shares": 2
[root@worker-0 ~]# crictl inspectp 1964ca23ccf0e | grep shares
            "shares": 2

I tried some other ways (e.g. parse "sleep 10h" since we anyway have this process running on the test pod) but all of them were a bit more complicated..

functests/1_performance/cpu_management.go
marcel-apf: ack

functests/1_performance/cpu_management.go
marcel-apf: ack

functests/1_performance/cpu_management.go
marcel-apf: yes, hacky but needed

functests/1_performance/cpu_management.go
marcel-apf: I wonder if there may be a way to get this info, the cpus the pod is running on, using k8s. @fromanirh @cynepco3hahue do you know anything about it?

functests/1_performance/cpu_management.go
marcel-apf: Nice!

functests/1_performance/cpu_management.go
dshchedr: I would be happy to hear some advice but as I know the pod itself uses full cpu affinity mask and only running processes have the right mask.

fromanirh: Are we looking to get the cpu affinity mask of  any running process? If so I'd just use
```
grep Cpus_allowed_list /proc/self/status
```
Or do we need something else?

marcel-apf: Thanks @fromanirh this is what we were looking for. @dshchedr, instead of running tmux, get its psr while is running and so on, just run the above command in the pod, is cleaner.

functests/1_performance/cpu_management.go
gsr-shanks: Ack

functests/1_performance/cpu_management.go
dshchedr: Nice! thanks @fromanirh !

functests/1_performance/cpu_management.go
```diff
@@ -349,6 +311,73 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+
+			onlineCPUsSet, err := nodes.GetOnlineCPUsSet(workerRTNode)
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(defaultSmpAffinitySet).To(Equal(onlineCPUsSet), fmt.Sprintf("Default SMP Affinity %s should be equal to all active CPUs %s", defaultSmpAffinitySet, onlineCPUsSet))
+
+			By("Running pod with annotations that disable specific CPU from IRQ balancer")
+			annotations := map[string]string{
+				"irq-load-balancing.crio.io": "true",
```
marcel-apf, line 340: @cynepco3hahue updated cri-o to use "disable" instead of true, we will need to update these eventually

functests/1_performance/cpu_management.go
```diff
@@ -349,6 +311,73 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+			defaultSmpAffinitySet, err = nodes.GetDefaultSmpAffinitySet(workerRTNode)
+			Expect(err).ToNot(HaveOccurred())
+
+			getPsr := []string{"/bin/bash", "-c", "grep Cpus_allowed_list /proc/self/status | awk '{print $2}'"}
+			psr, err := pods.ExecCommandOnPod(testpod, getPsr)
+			Expect(err).ToNot(HaveOccurred())
+			psrSet, err := cpuset.Parse(strings.Trim(string(psr), "\n"))
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(defaultSmpAffinitySet).To(Equal(onlineCPUsSet.Difference(psrSet)), fmt.Sprintf("Default SMP affinity should not contain isolated CPU %s", psr))
```
marcel-apf, line 360: Nice and simple!

marcel-apf:
Looks good to me
/lgtm

functests/1_performance/cpu_management.go
```diff
@@ -349,6 +311,73 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+
+			By("Checking that there are no any active IRQ on isolated CPU")
+			// It may takes some time for the system to reschedule active IRQs
+			Eventually(func() string {
+				getActiveIrq := []string{"/bin/bash", "-c", "for n in $(find /proc/irq/ -name smp_affinity_list); do echo $(cat $n); done"}
```
marcel-apf, line 365: Can be simplified with "find /proc/irq/ -name smp_affinity_list -exec cat {} \;"
I am not requesting a re-spin, only pointing out

functests/1_performance/cpu_management.go
```diff
@@ -349,6 +311,73 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+				activeIrqList, err := nodes.ExecCommandOnNode(getActiveIrq, workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+				re := regexp.MustCompile(fmt.Sprintf(`(?m)^%s$`, psrSet))
```
marcel-apf, line 368: Can you please explain this regex?
The psrSet may be  2,3, and the result of activeIrqList may be 0-5,1,4,5
0-5 includes 2,3 would the regex find it?

PR #454, opened 2020-11-27
Don't check for errors while waiting for profile to be applied.

fedepaol:
In order to make the tests work with a single node cluster, every time there is a chance the node is rebooted (after applying / deleting a profile), we just need to wait for conditions without checking the errors. This is because if polling happens when the cluster is down, the api calls will fail.

### DIFF
--- a/functests/0_config/config.go
+++ b/functests/0_config/config.go
@@ -42,26 +42,13 @@ var _ = Describe("[performance][config] Performance configuration", func() {
 		profileAlreadyExists := false
 
 		performanceManifest, foundOverride := os.LookupEnv("PERFORMANCE_PROFILE_MANIFEST_OVERRIDE")
+		var err error
 		if foundOverride {
-			var err error
 			performanceProfile, err = externalPerformanceProfile(performanceManifest)
 			Expect(err).ToNot(HaveOccurred(), "Failed overriding performance profile", performanceManifest)
 			klog.Warning("Consuming performance profile from ", performanceManifest)
 		}
-		if !discovery.Enabled() || foundOverride {
-			By("Creating the PerformanceProfile")
-			// this might fail while the operator is still being deployed and the CRD does not exist yet
-			Eventually(func() error {
-				err := testclient.Client.Create(context.TODO(), performanceProfile)
-				if errors.IsAlreadyExists(err) {
-					klog.Warning(fmt.Sprintf("A PerformanceProfile with name %s already exists! If created externally, tests might have unexpected behaviour", performanceProfile.Name))
-					profileAlreadyExists = true
-					return nil
-				}
-				return err
-			}, 15*time.Minute, 15*time.Second).ShouldNot(HaveOccurred(), "Failed creating the performance profile")
-		} else if !foundOverride {
-			var err error
+		if discovery.Enabled() {
 			performanceProfile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
 			Expect(err).ToNot(HaveOccurred(), "Failed finding a performance profile in discovery mode")
 			klog.Info("Discovery mode: consuming a deployed performance profile from the cluster")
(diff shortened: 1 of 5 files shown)


fedepaol:
/hold
needs validation against a single node cluster, but it's ready for reviews :-)

cynepco3hahue:
@fedepaol Can you fix CI error
```
hack/verify-generated.sh
 M go.mod
```

cynepco3hahue:
@fedepaol Did you validate it?

fedepaol:
> @fedepaol Did you validate it?

nope, sorry. It's on my radar but current tests are happepning with 4.6, so I need to find a hole when I can test this.

fedepaol:
@cynepco3hahue I updated it after the (failing) validation.
Now it works (despite a couple of failures), but I'd keep the hold until we branch 4.8 out

fedepaol:
@cynepco3hahue forgot to add, if you can take another look so I address any eventual comment, that would be great.

functests/utils/mcps/mcps.go
```diff
@@ -117,18 +154,33 @@ func GetConditionReason(mcpName string, conditionType machineconfigv1.MachineCon
...
+
+		selector := labels.NewSelector()
+		selector = selector.Add(*req)
+		cnfNodes, err = nodes.GetBySelector(selector)
+		if err != nil {
+			return errors.Wrap(err, "Failed getting nodes by selector")
+		}
+		return nil
+	}, 10*time.Minute, 5*time.Second).ShouldNot(HaveOccurred())
+	fmt.Println("FEDE", "exited")
```
cynepco3hahue, line 182: :) please remove

functests/utils/mcps/mcps.go
```diff
@@ -144,7 +196,10 @@ func WaitForCondition(mcpName string, conditionType machineconfigv1.MachineConfi
 func WaitForProfilePickedUp(mcpName string, profileName string) {
 	EventuallyWithOffset(1, func() bool {
 		mcp, err := GetByName(mcpName)
-		Expect(err).ToNot(HaveOccurred(), "Failed getting MCP by name")
+		// we ignore the error and just retry in case of single node cluster
+		if err != nil {
```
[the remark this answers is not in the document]
fedepaol: If I log, it will dump every 30 seconds, not sure it makes sense.
I'll put a log before and one after, I think it would serve the same purpose but it will be less intrusive

functests/utils/mcps/mcps.go
fedepaol: Done

fedepaol:
@cynepco3hahue I updated the pr.
While testing the dpdk tests, I realized it wasn't fully working because the kubelet config was not being applied.
I am treating the master as a special case since the mcp is labelled differently.

See https://github.com/openshift-kni/performance-addon-operators/compare/030e35d85690d139d7fa8eaa5419bf130cda4d8f..da8d030da596285829886da3b2174c694381ed24#diff-da19decbf419f8814aa046da1b2d5b09154f08a54445b9377964b7a01fe52db6R178

functests/0_config/config.go
```diff
@@ -173,4 +175,13 @@ func testProfile() *performancev2.PerformanceProfile {
 			},
 		},
 	}
+	// If the machineConfigPool is master, the automatic selector from PAO won't work
+	// since the machineconfiguration.openshift.io/role label is not applied to the
+	// master pool, hence we put an explicit selector here.
+	if utils.RoleWorkerCNF == "master" {
```
cynepco3hahue, line 181: works for me, I hope it just will not create us additional problems in the future

functests/0_config/config.go
```diff
@@ -72,10 +59,25 @@ var _ = Describe("[performance][config] Performance configuration", func() {
 		mcpLabel := profile.GetMachineConfigLabel(performanceProfile)
 		key, value := components.GetFirstKeyAndValue(mcpLabel)
 		mcpsByLabel, err := mcps.GetByLabel(key, value)
+		fmt.Println("FEDE", key, value, mcpsByLabel, mcpLabel)
```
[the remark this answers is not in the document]
fedepaol: dang

functests/0_config/config.go
fedepaol: The other option was to handle this special case in a generic way, doing way more parsing with higher risk of mistake.
The master case is really a special one, so this felt a lot safer.

fedepaol:
@cynepco3hahue saw the branching happened, mind re-approving this?

cynepco3hahue:
@fedepaol Do you need this one for 4.7?

PR #459, opened 2020-12-02
Align CPU load balancing functionality with the latest CRI-O changes

cynepco3hahue:
- add allowed_annotations under the runtime handler configuration
- change the runtime handler name to the low-latency
- use disable value for the CPU load balancing annotation
- update the documentation

### DIFF
--- a/build/assets/configs/99-runtimes.conf
+++ b/build/assets/configs/99-runtimes.conf
@@ -4,11 +4,12 @@ runtime_path = ""
 runtime_type = "oci"
 runtime_root = "/run/runc"
 
-# The CRI-O will check the runtime handler name under the code and will activate high-performance features,
-# like CPU load balancing.
+# The CRI-O will check the allowed_annotations under the runtime handler and apply low-latency hooks when one of
+# low-latency annotations presents under it.
 # We should provide the runtime_path because we need to inform that we want to re-use runc binary and we
-# do not have high-performance binary under the $PATH that will point to it.
-[crio.runtime.runtimes.high-performance]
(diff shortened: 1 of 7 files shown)


cynepco3hahue:
The CI should run with the RHCoS Build 47.83.202012020056-0

fromanirh:
@cynepco3hahue this is relevant for 4.7 only, right? (meaning the patched crio will never land in 4.6 and before)
IOW, can we assume
- 4.6 and before: old annotations
- 4.7 and beyond: new annotations (the one we use here)

cynepco3hahue:
@fromanirh 

> @cynepco3hahue this is relevant for 4.7 only, right? (meaning the patched crio will never land in 4.6 and before)
> IOW, can we assume
> 
> * 4.6 and before: old annotations
> * 4.7 and beyond: new annotations (the one we use here)

Absolutely, the old annotation value will also continue to work for 4.7(as part of the deprecation cycle) and will be removed under the 4.8(cri-o 1.21)

fromanirh:
> @fromanirh
> 
> > @cynepco3hahue this is relevant for 4.7 only, right? (meaning the patched crio will never land in 4.6 and before)
> > IOW, can we assume
> > 
> > * 4.6 and before: old annotations
> > * 4.7 and beyond: new annotations (the one we use here)
> 
> Absolutely, the old annotation value will also continue to work for 4.7(as part of the deprecation cycle) and will be removed under the 4.8(cri-o 1.21)

great, I was worried about the upgrade flow but it seems we are in the clean

PR #461, opened 2020-12-03
Fix the expected kubeletconfig failure condition status

cynepco3hahue:
Remove the kubeletconfig condition status check because
kubeletconfig failure condition has the False status instead of True and I do
not really know if it some use cases when it can be True.

```
Status:
  Conditions:
    Last Transition Time:  2020-12-03T13:37:06Z
    Message:               Error: could not find any MachineConfigPool set for KubeletConfig
    Status:                False
    Type:                  Failure
```

### DIFF
--- a/controllers/status.go
+++ b/controllers/status.go
@@ -234,10 +234,6 @@ func (r *PerformanceProfileReconciler) getKubeletConditionsByProfile(profile *pe
 		return nil, nil
 	}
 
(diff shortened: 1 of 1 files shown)


fromanirh:
yuck! this could be a MCO quirk :\ but let's stop the bleeding first and investigate better later on

cynepco3hahue:
According to https://github.com/openshift/machine-config-operator/pull/2369 the state of the failure will stay False

PR #462, opened 2020-12-08
Bug 1905492: Update the tuned profile to set higher scheduler priority

cynepco3hahue:
The stalld service has higher priority than `ksoftirqd`, `rcub` and `rcuc` that
can lead to the system freeze under the load pressure. To avoid it we should
raise the scheduler priority of these threads to be higher than the stalld service.

### DIFF
--- a/build/assets/tuned/openshift-node-performance
+++ b/build/assets/tuned/openshift-node-performance
@@ -35,8 +35,12 @@ transparent_hugepages=never                   #  network-latency
 [irqbalance]
 # Override the value set by cpu-partitioning with an empty one
 banned_cpus=""
+{{end}}
 
 [scheduler]
+group.ksoftirqd=0:f:11:*:ksoftirqd.*
+group.rcuc=0:f:11:*:rcuc.*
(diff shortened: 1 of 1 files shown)


build/assets/tuned/openshift-node-performance
```diff
@@ -35,8 +35,14 @@ transparent_hugepages=never                   #  network-latency
 [irqbalance]
 # Override the value set by cpu-partitioning with an empty one
 banned_cpus=""
+{{end}}
 
 [scheduler]
+group.ksoftirqd=0:f:11:*:ksoftirqd.*
+group.rcuc=0:f:13:*:rcuc.*
```
yanirq, line 42: ksoftirqd and rcuc should run with the same rt priority (11)

yanirq (changes requested):
Added comments per rt-kernel team feedback on a different thread

cynepco3hahue:
@yanirq Thanks for the answer, I updated the PR

fromanirh:
/lgtm
/hold
in case @yanirq wants to have another look

PR #463, opened 2020-12-08
Make test "tuned profile validation" more flexible

dshchedr:
This PR improves the test in scenario when we have some additional Tuned based on PAO profile:

```
    [test_id:28467][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Should contain configuration injected through the openshift-node-performance profile [It]
    /home/dshchedr/job/cnf/sno/performance-addon-operators/functests/1_performance/performance.go:226
    Timed out after 480.000s.
    active_profile is not set to openshift-node-performance-performance. <nil>
    Expected
        <string>: openshift-configuration-hotfixes
    to equal
        <string>: openshift-node-performance-performance

``` 
This happens because the file /etc/tuned/active_profile on the node contains new Tuned name (openshift-configuration-hotfixes), but we are trying to compare it with name in Performance Profile

Small improvement in this PR does verification if some another Tuned is based on PAO profile and changes name accordingly.

### DIFF
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 	"k8s.io/utils/pointer"
 
@@ -789,6 +790,21 @@ func validatTunedActiveProfile(nodes []corev1.Node) {
 	var err error
 	var out []byte
 	activeProfileName := components.GetComponentName(testutils.PerformanceProfileName, components.ProfileNamePerformance)
+
+	// check if some another Tuned profile overwrites PAO profile
+	tunedList := &tunedv1.TunedList{}
+	err = testclient.Client.List(context.TODO(), tunedList)
+	Expect(err).NotTo(HaveOccurred())
+
+	for _, t := range tunedList.Items {
+		if len(t.Spec.Profile) > 0 && t.Spec.Profile[0].Data != nil && strings.Contains(*t.Spec.Profile[0].Data, fmt.Sprintf("include=%s", activeProfileName)) {
+			klog.Warning(fmt.Sprintf("PAO tuned profile amended by '%s' profile, test may fail", t.Name))
+			if t.Spec.Profile[0].Name != nil {
+				activeProfileName = *t.Spec.Profile[0].Name
+			}
+		}
+	}
+
 	for _, node := range nodes {
 		tuned := tunedForNode(&node)
 		tunedName := tuned.ObjectMeta.Name


fromanirh:
/lgtm
/hold
looks good to me, but I'd like Artyom's review as well

functests/1_performance/performance.go
```diff
@@ -781,6 +782,19 @@ func validatTunedActiveProfile(nodes []corev1.Node) {
 	var err error
 	var out []byte
 	activeProfileName := components.GetComponentName(testutils.PerformanceProfileName, components.ProfileNamePerformance)
+
+	// check if some another Tuned profile overwrites PAO profile
+	tunedList := &tunedv1.TunedList{}
+	if err := testclient.Client.List(context.TODO(), tunedList); err != nil {
+		println(err)
```
cynepco3hahue, line 789: What does `println` is doing and where does it define?

functests/1_performance/performance.go
```diff
@@ -781,6 +782,19 @@ func validatTunedActiveProfile(nodes []corev1.Node) {
 	var err error
 	var out []byte
 	activeProfileName := components.GetComponentName(testutils.PerformanceProfileName, components.ProfileNamePerformance)
+
+	// check if some another Tuned profile overwrites PAO profile
```
cynepco3hahue, line 786: Why not skip the test in this case?

functests/1_performance/performance.go
fromanirh: and why the test line passed if this is undefined?

fromanirh: TIL: https://golang.org/pkg/builtin/#println https://play.golang.org/p/8ltQKd03z9X

cynepco3hahue: hm `Println is useful for bootstrapping and debugging; it is not guaranteed to stay in the language.`, let's just use klog it anyway already used under the method

fromanirh: agreed :) at least now we knw why it compiled fine :)

dshchedr: to be honest, now I can't find from a which corner of my head I've taken this... 

changed to `klog.Error(err)`

functests/1_performance/performance.go
dshchedr: this function _validatTunedActiveProfile()_ is a part of several tests where we are also verifying some system configurations. Most of users should not change these configurations, so I'm not sure skipping the whole test is a good idea.. User will be warned that this test may fail, but in most of scenarios it should be passed.

functests/1_performance/performance.go
```diff
@@ -781,6 +782,19 @@ func validatTunedActiveProfile(nodes []corev1.Node) {
...
+	tunedList := &tunedv1.TunedList{}
+	if err := testclient.Client.List(context.TODO(), tunedList); err != nil {
+		klog.Error(err)
+	}
+	for _, t := range tunedList.Items {
+		if strings.Contains(*t.Spec.Profile[0].Data, fmt.Sprintf("include=%s", activeProfileName)) {
```
cynepco3hahue, line 792: we should check:
- `len(t.Spec.Profile) > 0`
- `t.Spec.Profile[0].Data != nil`

functests/1_performance/performance.go
```diff
@@ -781,6 +782,19 @@ func validatTunedActiveProfile(nodes []corev1.Node) {
...
+			klog.Warning(fmt.Sprintf("PAO tuned profile amended by '%s' profile, test may fail", t.Name))
+			activeProfileName = *t.Spec.Profile[0].Name
```
cynepco3hahue, line 794: the same check that `t.Spec.Profile[0].Name != nil` before dereferece

functests/1_performance/performance.go
```diff
@@ -781,6 +782,19 @@ func validatTunedActiveProfile(nodes []corev1.Node) {
 	var err error
 	var out []byte
 	activeProfileName := components.GetComponentName(testutils.PerformanceProfileName, components.ProfileNamePerformance)
+
+	// check if some another Tuned profile overwrites PAO profile
+	tunedList := &tunedv1.TunedList{}
+	if err := testclient.Client.List(context.TODO(), tunedList); err != nil {
+		klog.Error(err)
+	}
+	for _, t := range tunedList.Items {
```
cynepco3hahue, line 791: if you get an error, probably tunedList will be nil

dshchedr: error should not happen, added: 	_Expect(err).NotTo(HaveOccurred())_

functests/1_performance/performance.go
dshchedr: added this checking

functests/1_performance/performance.go
dshchedr: hm, not sure if this object can be without _Name_ field, but added checking that.. thanks

PR #466, opened 2020-12-15
machineconfig: fix the RPS boot failures

marcel-apf:
It turns out the veth devices are renamed as soon as they are created,
resulting in a race between that time and the time the update-rps
service handles the interface.

If the net device directory does not exist assume the net device
has already been renamed and trace it using the systemd device
dependencies.

### DIFF
--- a/build/assets/scripts/set-rps-mask.sh
+++ b/build/assets/scripts/set-rps-mask.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+dev=$1
+[ -n "${dev}" ] || { echo "The device argument is missing" >&2 ; exit 1; }
+
+mask=$2
+[ -n "${mask}" ] || { echo "The mask argument is missing" >&2 ; exit 1; }
+
+dev_dir="/sys/class/net/${dev}"
+if [ ! -d "${dev_dir}" ]; then      # the net device was renamed, find the new name
+    systemd_devs=$(systemctl list-units -t device | grep sys-subsystem-net-devices | cut -d' ' -f1)
+
+    for systemd_dev in ${systemd_devs}; do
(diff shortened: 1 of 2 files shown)


build/assets/scripts/set-rps-mask.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+dev=$1
+[ -n "${dev}" ] || { echo "The device argument is missing" >&2 ; exit 1; }
+
+mask=$2
+[ -n "${mask}" ] || { echo "The mask argument is missing" >&2 ; exit 1; }
+
+dev_dir="/sys/class/net/${dev}"
+if [ ! -d "${dev_dir}" ]; then      # the net device was renamed, find the new name
```
fromanirh, line 10: isn't this still racy? it should be a little safer to try the clean way first and only if the clean way fails do this fallback.
If this can't be done for any reason please just add a code comment to document it.

fromanirh (commented):
a question inside, besides that LGTM.
you most likely need to rebase on top of https://github.com/openshift-kni/performance-addon-operators/pull/469

build/assets/scripts/set-rps-mask.sh
marcel-apf: Is not racy because the system renames the device or it doesn't. If the kernel decided to rename the device, it is done at kernel level way before the systemd service runs  (by observation and event flow). That was the point of the comment "the device was renamed".

marcel-apf:
> a question inside, besides that LGTM.
> you most likely need to rebase on top of #469

No problem, I will do it if necessary

build/assets/scripts/set-rps-mask.sh
fromanirh: fine, then we're good to go.

PR #467, opened 2020-12-16
Implement validation webhook

cynepco3hahue:
The validation logic will prevent creating incorrect profiles or
updating the existing profile to the incorrect one.

It will include all logic that was placed under the reconcile loop and also
will prevent the creation of the performance profile when the cluster
already has a performance profile with the same node selector.


api/v2/performanceprofile_validation.go
```diff
@@ -0,0 +1,278 @@
...
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *PerformanceProfile) ValidateDelete() error {
+	klog.Infof("Delete validation for the performance profile %q", r.Name)
+
+	// TODO(user): fill in your validation logic upon object deletion.
+	return nil
+}
+
+func (r *PerformanceProfile) validateNodeSelectorDuplication(ppList *PerformanceProfileList) field.ErrorList {
```
fromanirh, line 109: next time please split the new code from the code movement (two separate commits)

fromanirh (approved):
/approve
/lgtm

marcel-apf:
/hold
I have a review in progress, sorry for being slow. I will submit it ASAP.

cynepco3hahue:
@marcel-apf no worries take your time, probably we anyway need to wait for 4.8 branch

marcel-apf (commented):
Looks great. I have only a few minor comments and questions.
Are you planning any E2E validation test?

api/v2/performanceprofile_validation.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func (r *PerformanceProfile) validateCPUs() field.ErrorList {
+	var allErrs field.ErrorList
+
+	if r.Spec.CPU == nil {
+		allErrs = append(allErrs, field.Required(field.NewPath("spec.cpu"), "cpu section required"))
+	} else {
+		if r.Spec.CPU.Isolated == nil {
+			allErrs = append(allErrs, field.Required(field.NewPath("spec.cpu.isolated"), "isolated CPUs required"))
```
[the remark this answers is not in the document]
fromanirh: good catch. It was me in https://github.com/openshift-kni/performance-addon-operators/commit/dbec255b975aa60ac6ee2b25ec6f07a29fca5ba3

deploy/olm-catalog/performance-addon-operator/4.7.0/performance-addon-operator.v4.7.0.clusterserviceversion.yaml
```diff
@@ -264,3 +264,23 @@ spec:
     targetPort: 4343
     type: ConversionWebhook
     webhookPath: /convert
+  - admissionReviewVersions:
+    - v1beta1
+    containerPort: 443
+    deploymentName: performance-operator
+    failurePolicy: Fail
+    generateName: vperformanceprofile.kb.io
```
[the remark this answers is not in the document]
cynepco3hahue: the name is auto-generated under the controller-runtime and it the name that was generated.

deploy/olm-catalog/performance-addon-operator/4.7.0/performance-addon-operator.v4.7.0.clusterserviceversion.yaml
```diff
@@ -264,3 +264,23 @@ spec:
...
+      - v2
+      operations:
+      - CREATE
+      - UPDATE
+      resources:
+      - performanceprofiles
+    sideEffects: None
+    targetPort: 4343
+    type: ValidatingAdmissionWebhook
+    webhookPath: /validate-performance-openshift-io-v2-performanceprofile
```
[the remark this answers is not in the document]
cynepco3hahue: again it auto-generated during registration

tools/csv-processor/csv-processor.go
```diff
@@ -157,9 +159,33 @@ Performance Addon Operator provides the ability to enable advanced node performa
...
 			ConversionCRDs:          []string{"performanceprofiles.performance.openshift.io"},
 		},
+		{
+			GenerateName:            "vperformanceprofile.kb.io",
+			Type:                    csvv1.ValidatingAdmissionWebhook,
+			DeploymentName:          strategySpec.DeploymentSpecs[0].Name,
+			ContainerPort:           443,
+			TargetPort:              &targetPort,
+			SideEffects:             &sideEffects,
+			AdmissionReviewVersions: []string{"v1beta1"},
```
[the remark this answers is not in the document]
cynepco3hahue: I unsure how does it work for the conversion webhook probably it does not return objects of the AdmissionReview at all, but v1 does not work because of the issue https://github.com/kubernetes-sigs/controller-runtime/issues/1161

deploy/olm-catalog/performance-addon-operator/4.7.0/performance-addon-operator.v4.7.0.clusterserviceversion.yaml
```diff
@@ -264,3 +264,23 @@ spec:
     targetPort: 4343
     type: ConversionWebhook
     webhookPath: /convert
+  - admissionReviewVersions:
+    - v1beta1
```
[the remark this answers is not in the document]
cynepco3hahue: see my answer above

api/v2/performanceprofile_validation.go
```diff
@@ -0,0 +1,278 @@
...
+
+// we need this variable only because our validate methods should have access to the client
+var validatorClient client.Client
+
+// InitValidatorClient defines the validator client that should be used under validator methods
+func InitValidatorClient(mgr ctrl.Manager) {
+	if validatorClient != nil {
+		return
+	}
+	validatorClient = mgr.GetClient()
```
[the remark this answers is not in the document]
cynepco3hahue: I can do it as part of the `SetupWebhookWithManager`

api/v2/performanceprofile_validation.go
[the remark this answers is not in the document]
cynepco3hahue: in general, we had this logic under the previous validation
```
components.NewCPULists(string(*r.Spec.CPU.Reserved), string(*r.Spec.CPU.Isolated))
```
just by luck, we did not get here the nil pointer exception during dereference
so we do not allow it to be nil but we allow it to be an empty string

api/v2/performanceprofile_validation.go
```diff
@@ -0,0 +1,278 @@
...
+	if len(r.Spec.NodeSelector) > 1 {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec.nodeSelector"), r.Spec.NodeSelector, "you should provide ony 1 NodeSelector"))
+	}
+
+	// in case MachineConfigLabels or MachineConfigPoolSelector are not set, we expect a certain format (domain/role)
+	// on the NodeSelector in order to be able to calculate the default values for the former metioned fields.
+	if r.Spec.MachineConfigLabel == nil || r.Spec.MachineConfigPoolSelector == nil {
+		k, _ := components.GetFirstKeyAndValue(r.Spec.NodeSelector)
+		if _, _, err := components.SplitLabelKey(k); err != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec.nodeSelector"), r.Spec.NodeSelector, "invalid NodeSelector label key, can't be split into domain/role"))
```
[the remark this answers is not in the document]
cynepco3hahue: I will update the message

cynepco3hahue:
> Looks great. I have only a few minor comments and questions.
> Are you planning any E2E validation test?

Yep I have an intention to add some basic e2e tests.

deploy/olm-catalog/performance-addon-operator/4.7.0/performance-addon-operator.v4.7.0.clusterserviceversion.yaml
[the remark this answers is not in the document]
marcel-apf: ack

tools/csv-processor/csv-processor.go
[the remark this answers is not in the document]
marcel-apf: hmm, so maybe add both?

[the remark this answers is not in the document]
marcel-apf: one of them should work

api/v2/performanceprofile_validation.go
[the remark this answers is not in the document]
marcel-apf: ok

api/v2/performanceprofile_validation.go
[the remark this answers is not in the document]
marcel-apf: that is really really strange!

tools/csv-processor/csv-processor.go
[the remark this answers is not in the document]
cynepco3hahue: I unsure we should add things that will not work, once the controller-runtime will support v1 we can update it.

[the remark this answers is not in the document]
marcel-apf: as long as we remember :)

api/v2/performanceprofile_validation.go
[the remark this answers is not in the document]
cynepco3hahue: We can discuss it, but I want to avoid change the behavior under the PR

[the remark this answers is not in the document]
fromanirh: IIRC there it was a (indeed too subtle) catch which prevendted the NPE, or maybe we just got lucky indeed. In any case, better to be explicit.
But per https://github.com/openshift-kni/performance-addon-operators/commit/dbec255b975aa60ac6ee2b25ec6f07a29fca5ba3 the existing behaviour is indeed NOT to required to supply the "reserved" list - albeit this is most liley not what people expect.

api/v2/performanceprofile_validation.go
```diff
@@ -0,0 +1,277 @@
...
+		return allErrs
+	}
+
+	// validate NUMA topology policy matches allowed values
+	if r.Spec.NUMA.TopologyPolicy != nil {
+		policy := *r.Spec.NUMA.TopologyPolicy
+		if policy != kubeletconfigv1beta1.NoneTopologyManagerPolicy &&
+			policy != kubeletconfigv1beta1.BestEffortTopologyManagerPolicy &&
+			policy != kubeletconfigv1beta1.RestrictedTopologyManagerPolicy &&
+			policy != kubeletconfigv1beta1.SingleNumaNodeTopologyManager {
```
fromanirh, line 271: in a future PR we may want to avoid duplicating this validation.

api/v2/performanceprofile_validation.go
```diff
@@ -0,0 +1,277 @@
...
+		// node selectors have different length and it can work on the same set of nodes
+		if reflect.DeepEqual(pp.Spec.NodeSelector, r.Spec.NodeSelector) {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec.nodeSelector"), r.Spec.NodeSelector, fmt.Sprintf("the profile has the same node selector as the performance profile %q", pp.Name)))
+		}
+	}
+
+	return allErrs
+}
+
+func (r *PerformanceProfile) validateFields() field.ErrorList {
```
fromanirh, line 124: are we turning what formerly where free functions into methods, right? What's the benefit of this specific move?
Moving the validation into the webhook is fine and welcome, I'm interested about functions vs methods

fromanirh (commented):
not a fan of turning functions into methods, but I don't hate it either. Mostly some questions inside and then we can move forward and merge.

api/v2/performanceprofile_validation.go
cynepco3hahue: The controller-runtime requires that the object should implement a validation interface to activate the validation webhook for it, so upper methods should be methods. And instead of passing the object, I changed all other functions to methods as well, IMHO it placed under the same file and used only for validation so it no reason to make them functions. If you have some strong objectives I can roll back this change.

api/v2/performanceprofile_validation.go
cynepco3hahue: Can you please clarify?

api/v2/performanceprofile_validation.go
fromanirh: I don't, both approaches have pros and cons. I don't have strong opinions in either direction so let's go like you did, no point in wasting this work.

api/v2/performanceprofile_validation.go
```diff
@@ -0,0 +1,277 @@
...
+	return apierrors.NewInvalid(
+		schema.GroupKind{Group: "performance.openshift.io", Kind: "PerformanceProfile"},
+		r.Name, allErrs)
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *PerformanceProfile) ValidateDelete() error {
+	klog.Infof("Delete validation for the performance profile %q", r.Name)
+
+	// TODO(user): fill in your validation logic upon object deletion.
```
marcel-apf, line 101: I suppose I will see it later :)

main.go
```diff
@@ -130,9 +136,21 @@ func main() {
...
+	// configure webhook server
+	webHookServer := mgr.GetWebhookServer()
+	webHookServer.Port = webhookPort
+	webHookServer.CertDir = webhookCertDir
+	webHookServer.CertName = webhookCertName
+	webHookServer.KeyName = webhookKeyName
+
 	if err = (&performancev1.PerformanceProfile{}).SetupWebhookWithManager(mgr); err != nil {
-		klog.Exitf("unable to create PerformanceProfile webhook : %v", err)
+		klog.Exitf("unable to create PerformanceProfile v1 webhook : %v", err)
```
marcel-apf, line 148: What about v1alpha1?

deploy/olm-catalog/performance-addon-operator/4.7.0/performance-addon-operator.v4.7.0.clusterserviceversion.yaml
```diff
@@ -264,3 +264,23 @@ spec:
...
+    containerPort: 443
+    deploymentName: performance-operator
+    failurePolicy: Fail
+    generateName: vwb.performance.openshift.io
+    rules:
+    - apiGroups:
+      - performance.openshift.io
+      apiVersions:
+      - v2
```
marcel-apf, line 277: We only validate v2? We don't validate v1 and v1alpha1? They can be passed as well to PAO

cynepco3hahue: @marcel-apf We should block the creation of the new profile with the old version, no?

api/v2/performanceprofile_validation.go
cynepco3hahue: nope it just a comment after the generation

main.go
cynepco3hahue: it should be registered only for hub

deploy/olm-catalog/performance-addon-operator/4.7.0/performance-addon-operator.v4.7.0.clusterserviceversion.yaml
marcel-apf: I don't get it, you want to be able to create or modify only v2 profiles ? And fail editing / creation of v1 /v1alpha1 profiles?

cynepco3hahue: I mean that once we have 4.8, do we want to allow a user to create profiles with old versions? I believe we want to deprecate the old API version after one or two release versions.

main.go
marcel-apf: Aha, so  why do we need the one below? performancev2 ?

cynepco3hahue: hm, I need to refresh my memory, so what I talked about before is wrong(does not relate to hub)
so the validation webhook implemented only for v2, so we call it for v1 to register conversion webhook and for v2 to register validating webhook

marcel-apf: got it, thanks

deploy/olm-catalog/performance-addon-operator/4.7.0/performance-addon-operator.v4.7.0.clusterserviceversion.yaml
marcel-apf: But we want them to be able to edit. As long as the user can edit a V1  profile we are fine. The reason is the profile promotion from v1 to v2 is lazy. It is done as soon as the user changes something in the profile.

cynepco3hahue: I see, it makes sense, so my plan for 4.8:
1) block the creation of v1 and v1alpha1 
2) provide validation during the update for all versions v2, v1, v1alpha1 WDT?

cynepco3hahue:
@marcel-apf After an additional check, I found that adding validation webhook only for v2 will be enough. The validation webhook configuration has a field `matchPolicy` and it configured to `matchPolicy: Equivalent`.

```
Equivalent means a request should be intercepted if modifies a resource listed in rules, even via another API group or version.
```

So the request will be intercepted also for v1 and v1alpha1.

See https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchpolicy for more details.

deploy/olm-catalog/performance-addon-operator/4.7.0/performance-addon-operator.v4.7.0.clusterserviceversion.yaml
marcel-apf: ack

marcel-apf:
> @marcel-apf After an additional check, I found that adding validation webhook only for v2 will be enough. The validation webhook configuration has a field `matchPolicy` and it configured to `matchPolicy: Equivalent`.
> 
> ```
> Equivalent means a request should be intercepted if modifies a resource listed in rules, even via another API group or version.
> ```
> 
> So the request will be intercepted also for v1 and v1alpha1.
> 

Perfect, so I am ok with ti

> See https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchpolicy for more details.

PR #470, opened 2020-12-17
Revert the change of the generated runtime class name

cynepco3hahue:
It was not a good idea to change the name of the generate runtime class
because the name is immutable field and during the upgrade you will have
failure:

```
2020-12-17T07:07:31.743Z	ERROR	controller	Reconciler error	{"reconcilerGroup": "performance.openshift.io", "reconcilerKind": "PerformanceProfile", "controller": "performanceprofile", "name": "ci-upgrade-test", "namespace": "", "error": "RuntimeClass.node.k8s.io \"performance-ci-upgrade-test\" is invalid: handler: Invalid value: \"low-latency\": field is immutable"}
```

To make it possible to change the name, we should first delete the runtime class and after the
performance-addon-operator will re-create it with the updated name.

I will check under a separate PR if it is possible, but to fix the upgrade-test I will revert the runtime-class name now.

### DIFF
--- a/build/assets/configs/99-runtimes.conf
+++ b/build/assets/configs/99-runtimes.conf
@@ -4,11 +4,11 @@ runtime_path = ""
 runtime_type = "oci"
 runtime_root = "/run/runc"
 
-# The CRI-O will check the allowed_annotations under the runtime handler and apply low-latency hooks when one of
-# low-latency annotations presents under it.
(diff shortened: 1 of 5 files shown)


fromanirh:
/lgtm
ok, let's try again to make the upgrade flow more robust

fromanirh:
/hold
let's check first the changes on the `openshift/release` side

PR #472, opened 2020-12-22
e2e: fix nil pointer errors during dereference of the enable field

cynepco3hahue:

### DIFF
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -603,6 +603,10 @@ func verifyV1alpha1Conversion(v1alpha1Profile *performancev1alpha1.PerformancePr
 		return fmt.Errorf("spec RealTimeKernel field is different")
 	}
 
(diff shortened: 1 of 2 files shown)


cynepco3hahue:
@marcel-apf FYI

functests/1_performance/performance.go
```diff
@@ -603,6 +603,10 @@ func verifyV1alpha1Conversion(v1alpha1Profile *performancev1alpha1.PerformancePr
 		return fmt.Errorf("spec RealTimeKernel field is different")
 	}
 
+	if (specRealTimeKernel.Enabled == nil) != (v1Profile.Spec.RealTimeKernel.Enabled == nil) {
```
marcel-apf, line 606: Thanks for finding it!
You may want to add the same check to the verifyV2Conversion, other than that lgtm

cynepco3hahue: Good catch, fixed.

PR #473, opened 2020-12-22
scripts: Yet another improvement to the set-rps-mask.sh script

marcel-apf:
A few service failures were spotted in a long running BM node.
It seems there may still be a race between the interface renaming
and the systemd service running the script.

Improve it by re-trying to find the renamed network device
after a few seconds.

While at it:
 - do not consider not finding the device a service error, sometimes
   the interfaces vanish fast.
 - consider the improbable case when going over the systemd services
   we find first the original device and not the new one.


cynepco3hahue:
For the best-effort case, the PR looks good to me
/lgtm

PR #475, opened 2020-12-23
Remove Duplicate Polarion test case id:

mrniranjan:
1. Remove Duplicate Polarion id for test case: Verify CPU affinity mask, CPU reservation and CPU isolation on worker node
2. Add test Case id for Test case:  Check CPU Affinity and Isolation on Not-Guaranteed QoS class pod


mrniranjan:
@gsr-shanks @dshchedr @Avielyo10  Please review .

PR #479, opened 2020-12-29
Bug 1903820: watch machine config pools not owned by the performance profile

cynepco3hahue:
Our logic to watch for machine config pools icorrect, because
the machine config pools not owned by the performance profile.

To fix it our logic will watch all machine config pools and will map
it to the performance profile.


yanirq:
will this cover the case of exiting profile ?  (e.g. discovery mode)

cynepco3hahue:
> will this cover the case of exiting profile ? (e.g. discovery mode)

Sure, why not :)

PR #480, opened 2020-12-29
e2e: enable back update tests

cynepco3hahue:
The bug with the rpm-ostree should be solved under the latest
4.7 RHCoS.

### DIFF
--- a/hack/run-functests.sh
+++ b/hack/run-functests.sh
@@ -1,8 +1,6 @@
 #!/bin/bash
 
-# All test suits should be enabled once the bug https://bugzilla.redhat.com/show_bug.cgi?id=1865839 fixed
-# GINKGO_SUITS=${GINKGO_SUITS:-functests}
-GINKGO_SUITS=${GINKGO_SUITS:-"functests/0_config functests/1_performance functests/3_performance_status"}
+GINKGO_SUITS=${GINKGO_SUITS:-functests}
 LATENCY_TEST_RUN=${LATENCY_TEST_RUN:-"false"}
 
 which ginkgo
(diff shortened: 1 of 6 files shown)


hack/run-functests.sh
```diff
@@ -1,8 +1,7 @@
 #!/bin/bash
 
-# All test suits should be enabled once the bug https://bugzilla.redhat.com/show_bug.cgi?id=1865839 fixed
-# GINKGO_SUITS=${GINKGO_SUITS:-functests}
-GINKGO_SUITS=${GINKGO_SUITS:-"functests/0_config functests/1_performance functests/3_performance_status"}
+GINKGO_SUITS=${GINKGO_SUITS:-functests}
+GINKGO_SUITS="functests/0_config functests/2_performance_update"
```
yanirq, line 4: why did you remove 1 and 3 in favour of 2 ?

fromanirh: same question. Everything else LGTM!

cynepco3hahue: my bad added it to test locally

cynepco3hahue: done

hack/run-functests.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 GINKGO_SUITS=${GINKGO_SUITS:-functests}
-GINKGO_SUITS="functests/0_config functests/2_performance_update"
 LATENCY_TEST_RUN=${LATENCY_TEST_RUN:-"false"}
```
fromanirh, line 4: can we please move this change in the parent patch? even though the end result is the same, it is better to avoid introducing this change in the first place.

cynepco3hahue: done

cynepco3hahue:
@yanirq @fromanirh Please review.

PR #481, opened 2021-01-03
Bug 1912118: does not set RT parameters for non-RT kernel

cynepco3hahue:
Avoid setting real-time related parameters for the non-RT kernel.
Setting these parameters can hang the system.

- the most problematic parameter is `sched_rt_runtime_us` that can prevent to get the CPU time for some tasks with a small priority.
- the stalld service does not start without `sched_rt_runtime_us=-1`


cynepco3hahue:
I think the problem exists also under the release-4.5, but it was a tech preview so it is less critical.

marcel-apf:
Nice! stalld and sched_rt_runtime_us = -1 . Makes a lot of sense
/lgtm

PR #489, opened 2021-01-07
upgrade test: fix initial operatorgroup

fromanirh:
when upgrading from 4.6 to 4.7, we need to make sure
to deploy the STARTING (not the ENDING) operator group,
because the operator group spec itself is one of the things
which changed during this transition, the 4.7 being more
restrictive (OwnNamespace and SingleNamespace installModes no longer
allowed).

Please note this is relevant ONLY for this transition; all the other
transitions are perfectly fine using either OperatorGroup definition.

IOW:
- 4.4 -> 4.5
- 4.5 -> 4.6
- ...
- 4.7 -> 4.8 (expected)

all use the *same* definition for both source and target OperatorGroup,
so using either to deploy first is fine.

Please note all of this is relevant only for upgrade tests.


cluster-setup/olm-pre-4.6/upgrade-test-cluster/performance/operator_subscription.patch.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: performance-addon-operator
+  namespace: openshift-performance-addon-operator
+spec:
+  channel: "4.5"
```
fromanirh, line 7: this is out of date (also in the parent directory)

cynepco3hahue:
@fromanirh All configuration is done under the CI, including the creation of the operator group https://github.com/openshift/release/blob/5a2393e9b57d70cdb5c15ecf40b37652a70c5055/ci-operator/config/openshift-kni/performance-addon-operators/openshift-kni-performance-addon-operators-release-4.6.yaml#L110 you can specify the targeted namespace here.

So IIUC you should
- update the config manifest to deploy under the specific targeted namespace
- update the upgrade test script to update the operator group before starting the upgrade process

Here the CI script that will deploy the operator group - https://github.com/openshift/release/blob/5a2393e9b57d70cdb5c15ecf40b37652a70c5055/ci-operator/step-registry/optional-operators/ci/subscribe/optional-operators-ci-subscribe-commands.sh#L78

fromanirh:
we will resume later if needed

PR #493, opened 2021-01-11
must-gather: compress kubelet logs

fromanirh:
must-gather want to collect kubelet logs. Problem is: these logs can get quite big (>= 1 G each) quite fast. To save bandwidth, we compress them, saving roughly 90% of the space (e.g. from ~1G to ~100M). Using gzip inside must-gather is not unheard of, at least audit logs and network logs gatherer already do it.

To save more bandwidth:
- we restrain ourself to collect the last 8 hours of kubelet logs
- we collect kubelet logs from worker nodes only

### DIFF
--- a/must-gather/collection-scripts/gather_nodes
+++ b/must-gather/collection-scripts/gather_nodes
@@ -70,7 +70,7 @@ for NODE in $(oc get nodes --no-headers -o custom-columns=':metadata.name'); do
     NODE_PATH=${NODES_PATH}/$NODE
     mkdir -p ${NODE_PATH}
     for UNIT in ${NODE_UNITS[@]}; do
-        timeout -k 5m 30m bash -c "oc adm node-logs $NODE -u $UNIT | gzip" > ${NODE_PATH}/${NODE}_logs_$UNIT.gz &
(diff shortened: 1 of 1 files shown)


fromanirh:
/retest
it seems we hit a glitch downloading the golang tarball

must-gather/collection-scripts/gather_nodes
```diff
@@ -79,3 +81,6 @@ wait "${ADM_PIDS[@]}"
 oc delete -f $DAEMONSET_MANIFEST
 oc delete -f $SERVICEACCOUNT_MANIFEST
 oc delete -f $NAMESPACE_MANIFEST
+
+# force disk flush to ensure that all data gathered is accessible in the copy container
+sync
```
MarSik, line 86: Is this really needed? I though sync just makes sure the underlying storage has it all. Applications go through kernel and get the storage+caches version.. So I am curious

fromanirh: Not sure, it's done pretty frequently in the (OCP) must-gather scripts though. Can be some cargo cult or a workaround of an obscure bug.

yanirq: that's the only thing I was wondering about too.  everything else looks fine

fromanirh: ok, I'll remove the `sync`s until proven needed (with a "why" attached)

fromanirh:
/retest
quota limits again

PR #508, opened 2021-01-26
gather-sysinfo: support collecting from chroot

fromanirh:
add `-root` option to be able to collect data from chroot-mounted
sysfs/procfs, as long as the chroot is common, which is usually
the case inside the containers.

As side effect, the list of collected files changes, but we
files we omit now are not relevant for the KNI use case.


fromanirh:
replaced by https://github.com/openshift-kni/performance-addon-operators/pull/514

PR #511, opened 2021-01-27
Skeleton code for the performance profile creator tool

swatisehgal:
To build the performance profile creator tool run: make create-performance-profile
    The tool can be tested by running:
    podman run --entrypoint /usr/local/bin/performance-profile-creator  -v ./must-gather:/must-gather:z quay.io/swsehgal/perf-profile-creator:latest > my-profile.yaml


cmd/performance-profile-creator/cmd/root.go
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Copyright 2021 Red Hat, Inc.
+ */
+
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/spf13/cobra"
```
fromanirh, line 22: better to group the imports like
- stdlib
- extra packages
- local (same project) packages

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -0,0 +1,109 @@
...
+	"os"
+
+	"k8s.io/klog"
+	"k8s.io/utils/pointer"
+
+	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
```
fromanirh, line 33: do we really really really need globals?

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func init() {
+	rootCmd.PersistentFlags().IntVarP(&reservedCPUCount, "reserved-cpu-count", "R", 0, "Number of reserved CPUs")
+	rootCmd.PersistentFlags().BoolVarP(&splitCPUsAcrossNUMA, "split-cpus-across-numa", "S", true, "Split the CPUs across NUMA nodes")
+	rootCmd.PersistentFlags().BoolVarP(&disableHT, "disable-ht", "H", false, "Disable Hyperthreading")
+	rootCmd.PersistentFlags().BoolVarP(&rtKernel, "rt-kernel", "K", true, "Enable Real Time Kernel")
+	rootCmd.PersistentFlags().BoolVarP(&userLevelNetworking, "user-level-networking", "U", false, "Run with User level Networking(DPDK) enabled")
+	rootCmd.PersistentFlags().StringVarP(&powerConsumptionMode, "power-consumption-mode", "P", "cstate", "The power consumption mode")
+	rootCmd.PersistentFlags().StringVarP(&outputFilePath, "output-file-path", "O", "/output/sample-profile.json", "Performance Profile file output path")
```
fromanirh, line 62: no absolute path please, always relative unless there is a really strong reason to use absolute (there is?)

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -0,0 +1,109 @@
...
+	rootCmd.PersistentFlags().StringVarP(&mustGatherDirPath, "must-gather-dir-path", "M", "/must-gather", "Must gather directory path")
```
fromanirh, line 63: ditto

swatisehgal:
cc: @marcel-apf

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -0,0 +1,109 @@
...
+	var performanceProfileData []byte
+	var err error
+
+	if performanceProfileData, err = json.Marshal(&profile); err != nil {
+		klog.Errorf("Unable to Marshal sample performance profile: %v", err)
+	}
+
+	klog.Infof("The performance profile data is %q", performanceProfileData)
+
+	err = ioutil.WriteFile(outputFilePath, performanceProfileData, 0777)
```
fromanirh, line 105: 0777 seems too much :) I'd start with 0644

fromanirh (commented):
nice start!

cmd/performance-profile-creator/cmd/root.go
marcel-apf: We need to be able to extract the output perf-prof yaml file after the tool outputs it to some dir.
We are open to suggestions

cmd/performance-profile-creator/cmd/root.go
marcel-apf: This is a directory we want to mount and the tool needs to know about. So we get the dir path  at commadline.
How can we do it better?

fromanirh:
there are multiple ways to fix the CI failures, which have the following root cause (besides lack of committed `go.sum`):
```
 STEP 6: COPY ${ASSETS_DIR} /assets
STEP 7: COPY ${BIN_DIR}performance-addon-operators /usr/local/bin/performance-operator
STEP 8: COPY ${BIN_DIR}performance-profile-creator /usr/local/bin/performance-profile-creator
error: build error: error building at STEP "COPY ${BIN_DIR}performance-profile-creator /usr/local/bin/performance-profile-creator": error adding sources [/tmp/build/inputs/performance-profile-creator]: error checking on source /tmp/build/inputs/performance-profile-creator under "/tmp/build/inputs": copier: stat: "/performance-profile-creator": no such file or directory
2021/01/27 14:27:15 No custom metadata found and prow metadata already exists. Not updating the metadata.
2021/01/27 14:27:15 Ran for 7m58s
error: some steps failed:
  * could not run steps: step performance-addon-operator failed: the build performance-addon-operator failed after 57s with reason DockerBuildFailed: Dockerfile build strategy has failed. 
```
Possibly the nicest is:
1. refactor this PR such as it only adds the binary - without adding to any container image. It should pass with green CI.
2. file a new PR top `openshift/release`, like this one: https://github.com/openshift/release/pull/14777. It should pass cleanly
3. file a new PR against PAO adding the binary to the relevant container image. It should now pass CI smoothly.

cmd/performance-profile-creator/cmd/root.go
fromanirh: Which is fine, but why the default path needs to be absolute (and why the subdirectory btw?) 
IOW, can't we just use `sample-profile.json` as default?

cmd/performance-profile-creator/cmd/root.go
fromanirh: This is fine. The default is less fine: `/must-gather` is outside of the file system hierarchy standard. Can't we just use a relative path? `must-gather` (not lack of heading '/')

cmd/performance-profile-creator/cmd/root.go
marcel-apf: I am not sure we can, /usr/local/bin is ro.
The /output aims to be a designated directory for all the artifacts. We use a tmpfs volume mount to make it rw
Any other wayto fo it?

cmd/performance-profile-creator/cmd/root.go
marcel-apf: We can, this will be used to mount the actual must-gather dir from the host. The default value can be anything. A dir inside /usr/local/bin is a good place like any other. We are going to override it anyway.

cmd/performance-profile-creator/cmd/root.go
marcel-apf: +1

Makefile
```diff
@@ -144,6 +144,14 @@ operator-sdk:
 		echo "Using operator-sdk cached at $(OPERATOR_SDK)";\
 	fi
 
+.PHONY: create-performance-profile
+create-performance-profile:  build-output-dir
+	@echo "Creating performance profile"
+	mkdir -p $(TOOLS_BIN_DIR); \
+    LDFLAGS="-s -w "; \
```
marcel-apf, line 151: Indentation issue ?

marcel-apf:
> ```
> 
> Possibly the nicest is:
> 
> 1. refactor this PR such as it only adds the binary - without adding to any container image. It should pass with green CI.
> 2. file a new PR top `openshift/release`, like this one: [openshift/release#14777](https://github.com/openshift/release/pull/14777). It should pass cleanly
> 3. file a new PR against PAO adding the binary to the relevant container image. It should now pass CI smoothly.



> there are multiple ways to fix the CI failures, which have the following root cause (besides lack of committed `go.sum`):
> 
> ```
> STEP 6: COPY ${ASSETS_DIR} /assets
> STEP 7: COPY ${BIN_DIR}performance-addon-operators /usr/local/bin/performance-operator
> STEP 8: COPY ${BIN_DIR}performance-profile-creator /usr/local/bin/performance-profile-creator
> error: build error: error building at STEP "COPY ${BIN_DIR}performance-profile-creator /usr/local/bin/performance-profile-creator": error adding sources [/tmp/build/inputs/performance-profile-creator]: error checking on source /tmp/build/inputs/performance-profile-creator under "/tmp/build/inputs": copier: stat: "/performance-profile-creator": no such file or directory
> 2021/01/27 14:27:15 No custom metadata found and prow metadata already exists. Not updating the metadata.
> 2021/01/27 14:27:15 Ran for 7m58s
> error: some steps failed:
>   * could not run steps: step performance-addon-operator failed: the build performance-addon-operator failed after 57s with reason DockerBuildFailed: Dockerfile build strategy has failed. 
> ```
> 
> Possibly the nicest is:
> 
> 1. refactor this PR such as it only adds the binary - without adding to any container image. It should pass with green CI.

OK

> 2. file a new PR top `openshift/release`, like this one: [openshift/release#14777](https://github.com/openshift/release/pull/14777). It should pass cleanly

Do we need to configure some yaml file for CI? We are "only" adding a new binary to the same PAO image.
I wasn't aware of this need.

> 3. file a new PR against PAO adding the binary to the relevant container image. It should now pass CI smoothly.

OK

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	var performanceProfileData []byte
+	var err error
+
+	if performanceProfileData, err = json.Marshal(&profile); err != nil {
+		klog.Errorf("Unable to Marshal sample performance profile: %v", err)
+	}
+
+	klog.Infof("The performance profile data is %q", performanceProfileData)
```
marcel-apf, line 103: I tired to use it, the output doesn't look user-friendly. Can you please send the output to STDOUT instead?
Then the user can podman run --entrypoint=/usr/local/bin/performance-profile-creator -v /home/marcel/Workspace/must-gather:/must-gather:z --mount=type=tmpfs,tmpfs-size=10M,destination=/output quay.io/swsehgal/perf-profile-creator:latest > my-profile.yaml
If this work we would not need the output-file-path arg or event the first tmpfs mount

fromanirh:
> > 1. file a new PR top `openshift/release`, like this one: [openshift/release#14777](https://github.com/openshift/release/pull/14777). It should pass cleanly
> 
> Do we need to configure some yaml file for CI? We are "only" adding a new binary to the same PAO image.
> I wasn't aware of this need.

Yes we do. The OCP CI only makes available from the build image the files we explicitely allow to.
Not sure there are other ways, I never checked so far (and changing the config will require a change to the CI configuration anyway...)

marcel-apf:
> > > 1. file a new PR top `openshift/release`, like this one: [openshift/release#14777](https://github.com/openshift/release/pull/14777). It should pass cleanly
> > 
> > 
> > Do we need to configure some yaml file for CI? We are "only" adding a new binary to the same PAO image.
> > I wasn't aware of this need.
> 
> Yes we do. The OCP CI only makes available from the build image the files we explicitely allow to.
> Not sure there are other ways, I never checked so far (and changing the config will require a change to the CI configuration anyway...)

Thanks for the info!

cmd/performance-profile-creator/cmd/root.go
marcel-apf: But maybe we can start sending the output to STDOUT? No arg, no mount?

fromanirh: well, if we want to set defaults to be super-specific to the containerized run use case and thus to make the podman invocation (somehow) shorter, well is not a big deal and we can move forward.
BTW, if the profile will be sent to stdout (see comment below) which output do we expect?

cmd/performance-profile-creator/cmd/root.go
fromanirh: yeah, if we stricly focus on the containerized use case it makes sense. Not a big fan fan but not a big deal either.

cmd/performance-profile-creator/cmd/root.go
fromanirh: btw is this meant for logging purposes only?

cmd/performance-profile-creator/cmd/root.go
marcel-apf: Nothing for now, let's send to STDOUT for now, we will need to send some debug info later on. 
But let's start simple.

swatisehgal: Sure, will update to send the output to STDOUT.

fromanirh: Debug only on stderr please. (all the logging packages already do that by default). Never pollute stdout with non-output (e.g. logging) data.

cmd/performance-profile-creator/cmd/root.go
swatisehgal: I can change it to `must-gather` if that looks better.

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Sure, will fix this!

cmd/performance-profile-creator/cmd/root.go
swatisehgal: We were thinking of running the tool something like :

marcel-apf: But the profile IS the data
The debug message will go to stderr

marcel-apf: We send the profile to stdout, all debug /logs to stderr
So we can run:  podman run --entrypoint performance-profile-creator  -v ./must-gather:/must-gather:z quay.io/swsehgal/perf-profile-creator:latest > my-profile.yaml
Then
oc apply -f my-profile.yaml

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Good point, I can move this to init.

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Ack, will do!

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Well, now that I think again, we are planning to send the data just to STDOUT and hence we won't need to write to the file.

cmd/performance-profile-creator/cmd/root.go
fromanirh: We should use local variables as much as we can (possibly packed into a struct). Not sure if moving them to init will work. They can be local in `init()` can work. I was thinking to something like https://github.com/openshift-kni/debug-tools/blob/main/cmd/knit/cmd/root.go - note we still use a global struct, which is still bad. I Didn't have time to check how to make it local to the `root` command and pass it to subcommands.

If we don't plan to use subcommands (for now) we can do like https://github.com/openshift-kni/debug-tools/blob/main/cmd/knit/cmd/cpuaff.go#L35

any way, the less (and more compact, e.g. at least packed into a struct) globals we use, the better.

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Will push the code to STDOUT rather than using Klog.

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Ack, will update to a struct for now.

marcel-apf (commented):
Hi Swati,
Looks good to me.
Please split the PR in 3 parts as @fromanirh mentioned in order to pass the CI.
Please let me know if the split is not clear.
We are going to merge them one by one.
Thanks!
Marcel

marcel-apf:
/lgtm
/hold
Let's let @fromanirh / others to have another look

Makefile
```diff
@@ -144,6 +144,14 @@ operator-sdk:
 		echo "Using operator-sdk cached at $(OPERATOR_SDK)";\
 	fi
 
+.PHONY: create-performance-profile
+create-performance-profile:  build-output-dir
+	echo "Creating performance profile"
+	mkdir -p $(TOOLS_BIN_DIR); \
+	LDFLAGS="-s -w "; \
+	LDFLAGS+="-X github.com/openshift-kni/performance-addon-operators/cmd/performance-profile-creator "; \
```
fromanirh, line 152: in a future PR, let's make sure we need these LDFLAGS. We copy stuff that works around too often (yes, I'm guilty of this too and it recently burned me of, exactly the reason why I'm raising this).

fromanirh (commented):
/lgtm
nice start

Makefile
swatisehgal: Yep, I am guilty of copying from the other target:) Will double check and update in a subsequent PR.

PR #512, opened 2021-01-27
Bug 1918788: Avoid unneeded object initialization during profile conversion 

cynepco3hahue:
It possible the situation when the source conversion object has nil for
some field, but we initialize this field under the destination object.

### DIFF
--- a/api/v1alpha1/performanceprofile_conversion.go
+++ b/api/v1alpha1/performanceprofile_conversion.go
@@ -3,7 +3,7 @@ package v1alpha1
 import (
 	"k8s.io/utils/pointer"
 
-	"github.com/openshift-kni/performance-addon-operators/api/v1"
+	v1 "github.com/openshift-kni/performance-addon-operators/api/v1"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
@@ -17,8 +17,9 @@ func (curr *PerformanceProfile) ConvertTo(dstRaw conversion.Hub) error {
 	dst.ObjectMeta = curr.ObjectMeta
 
 	// Spec
-	dst.Spec.CPU = new(v1.CPU)
 	if curr.Spec.CPU != nil {
+		dst.Spec.CPU = new(v1.CPU)
+
 		if curr.Spec.CPU.Reserved != nil {
 			reserved := v1.CPUSet(*curr.Spec.CPU.Reserved)
 			dst.Spec.CPU.Reserved = &reserved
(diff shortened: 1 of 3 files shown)


api/v1alpha1/performanceprofile_conversion.go
```diff
@@ -17,8 +17,9 @@ func (curr *PerformanceProfile) ConvertTo(dstRaw conversion.Hub) error {
 	dst.ObjectMeta = curr.ObjectMeta
 
 	// Spec
-	dst.Spec.CPU = new(v1.CPU)
 	if curr.Spec.CPU != nil {
+		dst.Spec.CPU = new(v1.CPU)
```
fromanirh, line 21: I don't follow. We know `curr.Spec.CPU` is non-nil so... we initialize it again? (here and in a bunch of other places)

fromanirh (commented):
I don't really understand why the logic is the reverse of what I'd expect

api/v1alpha1/performanceprofile_conversion.go
cynepco3hahue: we check the `curr` one, and if not nil we initialize the destination one.

fromanirh: ah right, stupid me. Reading again.

functests/1_performance/performance.go
```diff
@@ -474,41 +507,49 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
-			v1Profile.ResourceVersion = ""
-			v1Profile.Spec.NodeSelector = map[string]string{"test": "test"}
-			Expect(testclient.Client.Create(context.TODO(), v1Profile)).ToNot(HaveOccurred())
+		}
 
-			key = types.NamespacedName{
-				Name:      v1Profile.Name,
-				Namespace: v1Profile.Namespace,
-			}
+		When("the performance profile does not NUMA field", func() {
```
fromanirh, line 512: does not contain?

fromanirh (commented):
just a minor comment about a comment, and looks good.

api/v1alpha1/performanceprofile_conversion.go
cynepco3hahue: end of the week :)

functests/1_performance/performance.go
cynepco3hahue: done

fromanirh (commented):
/lgtm
/hold
to give other reviewers time

Avielyo10:
How didn't we catch it sooner?

cynepco3hahue:
Because we used the profile with the NUMA field.

MarSik (approved):

PR #519, opened 2021-01-29
bundle: make annotations consistent

fromanirh:
let's avoid the same error we made in 4.6: elide Z version from channel
name, use only "$MAJOR.$MINOR' version.

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #520, opened 2021-01-29
hack: add helper to bump the zversion

fromanirh:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,13 @@ BUILD_DATE=$$(date --utc -Iseconds)
 # Export GO111MODULE=on to enable project to be built from within GOPATH/src
 export GO111MODULE=on
 
+.PHONY: new-zversion
+new-zversion: bump-zversion generate
+
+.PHONY: bump-zversion
+bump-zversion:
+	./hack/bump-zversion.sh
+
 .PHONY: build
 build: gofmt golint govet dist-gather-sysinfo dist create-performance-profile generate-manifests-tree
 
--- a/hack/bump-zversion.sh
+++ b/hack/bump-zversion.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+
+set -e
+
+BASEDIR="$(dirname $0 )"
+
+# shellcheck source=common.sh
+source "$(dirname $0)/common.sh"
+
+CSV_VERSION_A=( ${CSV_VERSION//./ } ) # replace points, split into array
+ZVERSION_NOW="${CSV_VERSION_A[2]}"
+ZVERSION_NEXT=$( expr ${ZVERSION_NOW} + 1)
+CSV_VERSION_NEXT="${CSV_VERSION_A[0]}.${CSV_VERSION_A[1]}.${ZVERSION_NEXT}"
+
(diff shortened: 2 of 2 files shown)


fromanirh:
/hold
depends on https://github.com/openshift-kni/performance-addon-operators/pull/519

fromanirh:
@MarSik @cynepco3hahue @yanirq

yanirq:
helper as in manually invoked ? do we need an instruction section somewhere ?

fromanirh:
> helper as in manually invoked ? do we need an instruction section somewhere ? (readme mentioning ?)

yep, manually invoked. I'll add help text and explanation to the commit message.

fromanirh:
/hold cancel
PR 519 merged

swatisehgal:
Looks like this is to bump to the next minor version. A rookie question: Should this also take care of scenarios where Z stream major version has to be updated?

fromanirh:
> Looks like this is to bump to the next minor version. A rookie question: Should this also take care of scenarios where Z stream major version has to be updated?

Versions we use have this form: X.Y.Z
X= major
Y= minor
Z= patch

This in theory. In practice we inherit X from OCP, which rarely changes. So We actually change only Y and Z. Changing Y and Z carries different meanings:
Y = relevant new features.
Z = fixes only.

Each time we bump Z, we promise as simple and as painless as possible upgrade experience.
This script wants only to concern with Z upgrades fixed (X and) Y versions.

IOW, this script concerns about handling bumps from, say, 4.7.2 to 4.7.3 ("4.7" being fixed)

hack/bump-zversions.sh
```diff
@@ -0,0 +1,51 @@
...
+BASEDIR="$(dirname $0 )"
+
+# shellcheck source=common.sh
+source "$(dirname $0)/common.sh"
+
+CSV_VERSION_A=( ${CSV_VERSION//./ } ) # replace points, split into array
+ZVERSION_NOW="${CSV_VERSION_A[2]}"
+ZVERSION_NEXT=$( expr ${ZVERSION_NOW} + 1)
+CSV_VERSION_NEXT="${CSV_VERSION_A[0]}.${CSV_VERSION_A[1]}.${ZVERSION_NEXT}"
+
```
cynepco3hahue, line 14: I expect that the PR will update the variable under the `hack/common.sh` and after will call to relevant scripts
- call https://github.com/openshift-kni/performance-addon-operators/pull/520/files#diff-cb3ab62f5a92ed6c60ec87fc8fff45dc8ef7f17287ad3a24c6ebf7384cd91fe9R46
- call to `make generate` (I do not see an update of the skip range)
- update bundle images

fromanirh: sure, why not? let me make this script smarter
with one question though: how we can update the bundle images outside CI?

cynepco3hahue: ah sorry for misleading you I meant to update bundle Dockerfiles

fromanirh:
(yep, pastdd wrong comment, deleted)

hack/bump-zversions.sh
fromanirh: ok, so we want to update just one value in one place and autogenerate everything else, right? This will require much more changes.

PR #527, opened 2021-02-02
Update tools image to use ubi8

cynepco3hahue:
We have troubles with glibc library under the ubi8, so it can be a good reason to advance
tools image version to ubi8.

ubi8 does not have skopeo package, I can not find a single place where it should be used.


fromanirh:
/lgtm
we can afford some experiments in master branch

PR #538, opened 2021-02-07
Generate a machine config name with the numerical prefix

### LINKED ISSUE #490: Prepend numeral index when generating performance profile MachineConfig
Currently, the operator generates a MachineConfig with the name of `performance-<pao-name>` and does not follows a convention of pretending a numerical value to the name so when the rendered version of the MCP is generated it is applied in whatever alphanumerical order it shows in the list of `oc get mc`. Now, when a cluster-admin wants to do further customizations, it cannot longer create MC with a prepend index like `99-my-custom-mc-fix` as it will not be applied correctly when the MCP is rendered.

My proposal is to at least pre-pend something like a 50 for having something like`50-performance-<pao-name>`

cynepco3hahue:
Currently, the operator generates a MachineConfig with the name of performance-<pao-name> and
does not follows a convention of pretending a numerical value to the name so when the rendered
version of the MCP is generated it is applied in whatever alphanumerical order it shows in the
list of oc get mc.
Now, when a cluster-admin wants to do further customizations, it cannot longer create MC with
a prepend index like 99-my-custom-mc-fix as it will not be applied correctly when the MCP is rendered.

Fixes #490

### DIFF
--- a/controllers/performanceprofile_controller.go
+++ b/controllers/performanceprofile_controller.go
@@ -246,6 +246,11 @@ func (r *PerformanceProfileReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
 		return reconcile.Result{}, nil
 	}
 
+	// remove components with the old name after the upgrade
+	if err := r.deleteDeprecatedComponents(instance); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	// apply components
 	result, err := r.applyComponents(instance)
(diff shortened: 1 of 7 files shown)


MarSik:
/hold

I believe simply changing the generated name might break upgrade so before we merge this, please make sure you describe the flow and test it!

fromanirh (approved):
/lgtm

PR #539, opened 2021-02-08
[release-4.6] comment the stalld section

cynepco3hahue:
Under 4.6 stalld installed without systemd overlay file, so it can not be disabled via
service plugin. Remove the service section to make sure that stalld does not run under
the system.

Related to https://github.com/openshift-kni/performance-addon-operators/pull/524

### DIFF
--- a/build/assets/tuned/openshift-node-performance
+++ b/build/assets/tuned/openshift-node-performance
@@ -27,8 +27,8 @@ min_perf_pct=100                              #  latency-performance
 
 # Disable the stalld service until bugs https://bugzilla.redhat.com/show_bug.cgi?id=1912118 and
 # https://bugzilla.redhat.com/show_bug.cgi?id=1903302 will be fixed
-[service]
-service.stalld=stop,disable
+# [service]
+# service.stalld=start,enable
 
 [vm]
 transparent_hugepages=never                   #  network-latency


build/assets/tuned/openshift-node-performance
```diff
@@ -27,8 +27,8 @@ min_perf_pct=100                              #  latency-performance
 
 # Disable the stalld service until bugs https://bugzilla.redhat.com/show_bug.cgi?id=1912118 and
 # https://bugzilla.redhat.com/show_bug.cgi?id=1903302 will be fixed
-[service]
-service.stalld=stop,disable
+# [service]
+# service.stalld=start,enable
```
yanirq, line 31: I assume the comment is for us to remember to re enabled it ?

cynepco3hahue: yep

fromanirh:
I only have one comment/question: say someone wants to enable stalld for some reasons. How we can do that with this patch merged?

cynepco3hahue:
> I only have one comment/question: say someone wants to enable stalld for some reasons. How we can do that with this patch merged?

The same thing we did to disable it, you will need to create new tuned resource that will include the tuned generated by PAO and will have the service section.

fromanirh:
> > I only have one comment/question: say someone wants to enable stalld for some reasons. How we can do that with this patch merged?
> 
> The same thing we did to disable it, you will need to create new tuned resource that will include the tuned generated by PAO and will have the service section.

good enough for me.

PR #540, opened 2021-02-09
Add MCP as an argument and mark certain flags as required

swatisehgal:
- This patch marks reserved-cpu-count, mcp-name and rt-kernel
  as required fields and are expected to be specified. In case
  the user does not specify them, PPC returns an error


fromanirh (approved):
/lgtm

PR #545, opened 2021-02-10
Update discovery mode performance profile selector to choose by NODE_SELECTOR environment variable

sabinaaledort:
Currently when running cnf-tests in discovery mode the performance
profile selector is choosing the profile in the cluster with the
most nodes using it.
When running in a cluster with more than one performance profile,
the profile selector should choose the performance profile by
NODES_SELECTOR env var if it is set.

### DIFF
--- a/functests/utils/consts.go
+++ b/functests/utils/consts.go
@@ -42,7 +42,7 @@ func init() {
 	}
 
 	if discovery.Enabled() {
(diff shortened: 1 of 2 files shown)


cynepco3hahue:
/lgtm

Just be sure it reflected under the cnf-tests documentaiton.

PR #548, opened 2021-02-11
Bug 1885317: Provide a log package that will write to GinkgoWriter

cynepco3hahue:
We want to reduce the verbosity of logs output in the case when the test succeeded and provide more information only in case when the test failed.

```
Ginkgo provides a globally available io.Writer called GinkgoWriter that you can write to. GinkgoWriter aggregates input while a test is running and only dumps it to stdout if the test fails.
```

### DIFF
--- a/functests/0_config/config.go
+++ b/functests/0_config/config.go
@@ -16,7 +16,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/klog"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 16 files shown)


fromanirh (commented):
/approve
/lgtm
I think this is a good improvement

fromanirh:
/hold
give time to other reviewers

cynepco3hahue:
@fedepaol @fromanirh Can you please review again?

fromanirh:
/approve
/lgtm
/hold
I'm ok with this change in master, but backporting to stable branches is a different story (but we can retarget the related bug and fix only 4.8 and onwards).
Let's wait for other reviewers

fromanirh:
> /hold cancel

for the record: I agree. Wait for more comments should be timeboxed. PRs should not wait forever.

fedepaol:
Sorry for the delay, just went over it
/lgtm as well (and thanks)

PR #549, opened 2021-02-15
Find the nodes targeted by an MCP

swatisehgal:
Find the nodes targeted by an MCP

- In this patch we obtain node-selector from the MCP
  specified by the user and then return the nodes targetted
  by that MCP.
- In order to obtain node and mcp objects, we use yaml
   objects collected by must-gather.
 
    An example of how the tool was run to test this patch:
    ./build/_output/bin/performance-profile-creator \
    --mcp-name worker-cnf --reserved-cpu-count 2 --rt-kernel true \
    --must-gather-dir-path /root/must-gather.local.9039793491651665753 > performance-profile.yaml


cmd/performance-profile-creator/cmd/root.go
```diff
@@ -17,28 +17,159 @@
...
+	v1 "k8s.io/api/core/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/component-helpers/scheduling/corev1"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type ClientSet struct {
```
cynepco3hahue, line 44: why do not use the controller-runtime client?

swatisehgal: I needed to access the machineconfig client to be able to get mcp so created a user defined clientset something like was done [here](https://github.com/openshift/machine-config-operator/blob/master/test/e2e/framework/clientset.go). Anyway, based on discussion with @fromanirh  we are going to obtain the node/mcp info from the must gather rather than by querying the cluster so we won't need the client. Will add a commit for that.

cynepco3hahue: Just for the future, we do not use any clientset under the PAO, we only register the relevant scheme(https://github.com/openshift-kni/performance-addon-operators/blob/6c9616d44117b10f7df1fd799c2dff9a5821f083/main.go#L123), and after use the controller-runtime to get objects(https://github.com/openshift-kni/performance-addon-operators/blob/6c9616d44117b10f7df1fd799c2dff9a5821f083/controllers/resources.go#L29).

swatisehgal: Great, thanks!!

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -52,25 +59,22 @@ var rootCmd = &cobra.Command{
...
 		if err != nil {
 			return fmt.Errorf("Error obtaining Nods %s: %v", mcpName, err)
 		}
+
 		matchedNodes := make([]*v1.Node, 0)
 		for _, node := range nodes {
 			matches, _ := corev1.MatchNodeSelectorTerms(node, getNodeSelectorFromLabelSelector(labelSelector))
 			if matches {
-				fmt.Printf("Matched Nodes are: %v\n", node.GetName())
+				log.Println("Matched Node:", node.GetName())
```
fromanirh: minor: why Println and not Printf?

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -52,25 +59,22 @@ var rootCmd = &cobra.Command{
 	Short: "A tool that automates creation of Performance Profiles",
 	PreRunE: func(cmd *cobra.Command, args []string) error {
 		mcpName := cmd.Flag("mcp-name").Value.String()
-		kubeConfig := cmd.Flag("kube-config").Value.String()
-		clientset := NewClientSet(kubeConfig)
-		mcp, err := getMCPByName(mcpName, clientset)
+		mustGatherDirPath := cmd.Flag("must-gather-dir-path").Value.String()
```
fromanirh, line 58: if the preferred way is to read from must-gather, could you please squash this commit with the previous one in the next upload? No point in adding code and removing it in the very next commit.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -114,60 +118,100 @@ func getNodeSelectorFromLabelSelector(labelSelector *metav1.LabelSelector) *v1.N
 	return nodeSelector
 
 }
-func getNodes(mustGatherPath string, clientset *ClientSet) ([]*v1.Node, error) {
-	workerNodes := make([]*v1.Node, 0)
 
-	dirs, err := filepath.Glob(mustGatherPath + "/quay-io-openshift-kni-performance-addon-operator-must-gather*/" + "nodes")
+func getMustGatherFullPaths(mustGatherPath string, suffix string) (string, error) {
+	// The regular expression below depends on the must gather image name. It is assumed here
```
fromanirh, line 118: regular expression -> glob pattern

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -114,60 +118,100 @@ func getNodeSelectorFromLabelSelector(labelSelector *metav1.LabelSelector) *v1.N
...
+	// that the image would have "performance-addon-operator-must-gather" substring in the name.
+	paths, err := filepath.Glob(mustGatherPath + "/*performance-addon-operator-must-gather*/" + suffix)
+	// we return only one path here, we could add a validation
+	// 1) If len(paths) == 1 i.e. there is only one possible subdirectory under must gather prefix with the above prefix, we return the path
+	// 2) If len(paths) != 1 we choose one, one option could be to check the timestamp of the directory and choose the most fresh one
+	return paths[0], err
```
fromanirh, line 124: please note the order on which glob returns is not defined

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -114,60 +118,100 @@ func getNodeSelectorFromLabelSelector(labelSelector *metav1.LabelSelector) *v1.N
...
-	} else {
-		glog.V(4).Infof("Using in-cluster kube client config")
-		config, err = rest.InClusterConfig()
+	mcpPathSuffix := path.Join(ClusterScopedResources, MCPools, mcpName+YAMLSuffix)
+	mcpPath, err := getMustGatherFullPaths(mustGatherDirPath, mcpPathSuffix)
+	if err != nil {
+		return nil, fmt.Errorf("Error obtaining MachineConfigPool %s: %v", mcpName, err)
 	}
+
+	mcp, err = getMCP(mcpPath)
```
fromanirh, line 190: i'd just do `return getMCP(mcpPath)`. Which value we add in the return below?
Additionally: do we get MCPs by any other mean different than "by name"? If not we should squash together the two functions

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -34,25 +32,23 @@ import (
...
 )
 
 const (
+	// ClusterScopedResources defines the cluster scope resources under
+	// must-gather-dir/quay-io-openshift-kni-performance-addon-operator-must-gather*
 	ClusterScopedResources = "cluster-scoped-resources"
-	CoreNodes              = "core/nodes"
-	MCPools                = "machineconfiguration.openshift.io/machineconfigpools"
-	YAMLSuffix             = ".yaml"
+	// CoreNodes defines substring in the path to obtain node data
```
fromanirh, line 44: `defines the subpath, relative to ClusterScopedResources, on which we find node-specific data`

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -34,25 +32,23 @@ import (
...
+	CoreNodes = "core/nodes"
+	// MCPools defines substring in the path to obtain mcp data
```
fromanirh, line 46: `defines the subpath, relative to $DONT_REMEMBER_RIGHT_NOW_PLEASE_CHECK, on which we find the machine config pool definitions`

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -34,25 +32,23 @@ import (
...
+	MCPools = "machineconfiguration.openshift.io/machineconfigpools"
+	// YAMLSuffix defines yaml suffix
```
fromanirh, line 48: `YAMLSuffix is the extension of the yaml files saved by must-gather`

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -34,25 +32,23 @@ import (
...
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
-	clientmachineconfigv1 "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1"
 	v1 "k8s.io/api/core/v1"
-	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
+	// ClusterScopedResources defines the cluster scope resources under
```
fromanirh, line 41: `ClusterScopedResources defines the subpath, relative to the top-level must-gather directory[1] on which we find the cluster-scoped definitions saved by must-gather`
[1] please check

fromanirh (commented):
In general looks reasonnable, but we need to start adding tests to check the MCP/node matching logic. To facilitate the tests, we may want to create some package moving the code away from the `cmd/root`.go file

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -18,27 +18,197 @@ package cmd
...
+const (
+	// ClusterScopedResources defines the cluster scope resources under
+	// must-gather-dir/quay-io-openshift-kni-performance-addon-operator-must-gather*
+	ClusterScopedResources = "cluster-scoped-resources"
+	// CoreNodes defines substring in the path to obtain node data
+	CoreNodes = "core/nodes"
+	// MCPools defines substring in the path to obtain mcp data
+	MCPools = "machineconfiguration.openshift.io/machineconfigpools"
+	// YAMLSuffix defines yaml suffix
+	YAMLSuffix = ".yaml"
```
yanirq, line 49: can we possibly get also .yml files there ?  (minus the `a`)

swatisehgal: @fromanirh As per Yanir's question, Does must gather create YAMLs with .yml extension as well? Is it okay to assume that the extension is yaml only for now?

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Sure, will do

cmd/performance-profile-creator/cmd/root.go
fromanirh: well I took that the other way around, are we expecting extensions different from `.yaml` at all?

fromanirh: yep

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Ack, will make this more sophisticated by checking for the most fresh path

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Good point, will squash together.

swatisehgal:
> In general looks reasonnable, but we need to start adding tests to check the MCP/node matching logic. To facilitate the tests, we may want to create some package moving the code away from the `cmd/root`.go file

Tests and the code restructuring will be handled in a separate PR. Regarding tests to check the matching logic, we are using   [corev1.MatchNodeSelectorTerms](https://pkg.go.dev/k8s.io/component-helpers@v0.20.2/scheduling/corev1#MatchNodeSelectorTerms) for matching  which has been tested [here](https://github.com/kubernetes/component-helpers/blob/v0.20.2/scheduling/corev1/helpers_test.go#L63).

fromanirh:
> > In general looks reasonnable, but we need to start adding tests to check the MCP/node matching logic. To facilitate the tests, we may want to create some package moving the code away from the `cmd/root`.go file
> 
> Tests and the code restructuring will be handled in a separate PR. Regarding tests to check the matching logic, we are using [corev1.MatchNodeSelectorTerms](https://pkg.go.dev/k8s.io/component-helpers@v0.20.2/scheduling/corev1#MatchNodeSelectorTerms) for matching which has been tested [here](https://github.com/kubernetes/component-helpers/blob/v0.20.2/scheduling/corev1/helpers_test.go#L63).

Sure: but let's test how we use the matching logic.
Having e2e tests in a separate PR is ok, but not having any kind of tests in the PRs adding so much logic is not ideal.

swatisehgal:
> Sure: but let's test how we use the matching logic.
Okay, working now to add some tests!

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -18,27 +18,202 @@ package cmd
...
+			MatchExpressions: matchExpressions,
+			MatchFields:      matchFields,
+		},
+	}
+	nodeSelector := &v1.NodeSelector{
+		NodeSelectorTerms: nodeSelectorTerms,
+	}
+
+	return nodeSelector
+
```
marcel-apf: Nice!
(Last line maybe is not needed)

cmd/performance-profile-creator/cmd/root.go
marcel-apf: I think we should return error if more than one directory is found. We don't know which one the user really wants. Chronological order is just a guess. The error message should be clear: "Expected only one performance-addon-operator-must-gather* directory, please check the must-gather tarball..."

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -18,27 +18,202 @@ package cmd
...
+func getNode(path string) (*v1.Node, error) {
+	var node v1.Node
+	src, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("Error opening %q: %v", path, err)
+	}
+	defer src.Close()
+
+	dec := k8syaml.NewYAMLOrJSONDecoder(src, 1024)
```
marcel-apf: How do we know 1024 bytes are enough?

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -18,27 +18,202 @@ package cmd
...
+}
+func getMCP(path string) (*machineconfigv1.MachineConfigPool, error) {
+	var mcp machineconfigv1.MachineConfigPool
+
+	src, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("Error opening %q: %v", path, err)
+	}
+	defer src.Close()
+	dec := k8syaml.NewYAMLOrJSONDecoder(src, 1024)
```
marcel-apf: same question about the buffer size

marcel-apf (commented):
Nice PR!

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -18,27 +18,202 @@ package cmd
...
+type ClientSet struct {
+	corev1client.CoreV1Interface
+	clientmachineconfigv1.MachineconfigurationV1Interface
+}
+
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
 	Use:   "performance-profile-creator",
 	Short: "A tool that automates creation of Performance Profiles",
+	PreRunE: func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
swatisehgal: PreRunE can return an error as opposed to Run.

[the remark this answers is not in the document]
fromanirh: I guess the question was (also?) about if we can use `RunE`

cmd/performance-profile-creator/cmd/root.go
swatisehgal: 1024 is the bufferSize value which determines how far into the stream the decoder will look to figure out whether this is a JSON stream (has whitespace followed by an open brace), so in this case we know that we son't have whitespaces and the yaml is present right from the beginning of the file so 1024 is a safe choice.

pkg/profilecreator/profilecreator.go
```diff
@@ -49,6 +48,10 @@ const (
 	YAMLSuffix = ".yaml"
 )
 
+func init() {
+	log.SetOutput(os.Stderr)
```
fromanirh, line 52: this is not wrong, but seems redundant: https://github.com/sirupsen/logrus/blob/d59e5619da7b597f81da5b56ba4f315daba0a667/logger.go#L84

We can keep this to make it very explicit what we want, though.

pkg/profilecreator/helper.go
```diff
@@ -0,0 +1,24 @@
...
+}
+func newTestNodeList(nodes ...*v1.Node) []*v1.Node {
+	nodeList := make([]*v1.Node, 0)
+	for _, node := range nodes {
+		nodeList = append(nodeList, node)
+	}
+	return nodeList
+}
+
+func getAbsolutePath(mustGatherDirPath string) (string, error) {
```
fromanirh, line 22: this doesn't seem to add much value. But no biggie, we can delete in a later PR if we agree to.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -84,6 +84,7 @@ type profileCreatorArgs struct {
 
 func init() {
 	args := &profileCreatorArgs{}
+	log.SetOutput(os.Stderr)
```
fromanirh, line 87: see comment below

fromanirh (commented):
at glance it seems good to go, will have another review later

pkg/profilecreator/profilecreator.go
swatisehgal: Yep, we wanted to have it explicitly to show that we are logging to STDERR.

pkg/profilecreator/helper.go
swatisehgal: Okay, will keep this in mind and remove in a subsequent PR!

pkg/profilecreator/profilecreator.go
swatisehgal: @marcel-apf If you'd like me to remove it please let me know and I can remove it

pkg/profilecreator/helper.go
swatisehgal: fixed!

pkg/profilecreator/profilecreator.go
marcel-apf: I prefer it as "documentation", to show we want it that way on purpose

marcel-apf: But indeed, why use it twice?

swatisehgal: Okay, perfect!

marcel-apf: got it

marcel-apf: 2 different packages

PR #551, opened 2021-02-17
must gather: more sysinfo debug logs

fromanirh:
improve sysinfo debuggability. We had cases on which the toolsilently failed to retrieve the sysinfo data which must be investigated.
Worth pointing out that according to logs the collection on the worker nodes seems to have happened.
We need to narrow down where the issue is.

### DIFF
--- a/must-gather/collection-scripts/gather_nodes
+++ b/must-gather/collection-scripts/gather_nodes
@@ -52,7 +52,6 @@ do
 done
 
 WORKER_NODES=()
-SYSINFO_PIDS=()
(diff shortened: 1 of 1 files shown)


fromanirh:
/hold
we need to validate using custom builds

fromanirh:
/cc @cynepco3hahue @yanirq 
FYI @marcel-apf @MarSik

fromanirh:
replaced by https://github.com/openshift-kni/performance-addon-operators/pull/552

PR #552, opened 2021-02-17
must-gather: sysinfo: direct copy using stdout

fromanirh:
`gather-sysinfo` can emit the tar file directly to stdout.
Avoiding the local file, we can simplify our gathering script and
make the process a bit faster.


fromanirh:
pre-built images with this PR in: quay.io/fromani/pao-must-gather:sysinfo-stdout-$NUM (0, 1, 2...)

PR #553, opened 2021-02-17
tests: clarify failure reasons

fromanirh:
Add explanation for Expect() failures when the expecations is not
obvious.

Note:
in general, we not change expectations like
```
Expect(err).NotTo(HaveOccurred())
```
because the ginkgo default behaviour is expressive enough.

We can use format strings freely in the gomega annotations. See:
https://onsi.github.io/gomega/ - "Annotating assertions"


swatisehgal (approved):
Looks very pretty straight forward!

PR #555, opened 2021-02-18
must-gather: sysinfo: make sure we copy ghw files

fromanirh:
Make sure we clone all the files ghw (not just gather_sysinfo) cares about.

### DIFF
--- a/tools/gather-sysinfo/gather-sysinfo.go
+++ b/tools/gather-sysinfo/gather-sysinfo.go
@@ -2,10 +2,12 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/jaypipes/ghw/pkg/snapshot"
 )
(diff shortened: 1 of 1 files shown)


fromanirh:
image at `quay.io/fromani/pao-must-gather:sysinfo-stdout-1`

tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -101,3 +83,37 @@ func chrootFileSpecs(fileSpecs []string, root string) []string {
...
+	for retSpec := range specSet {
+		retSpecs = append(retSpecs, retSpec)
+	}
+	return retSpecs
+}
+
+func kniExpectedCloneContent() []string {
+	return []string{
+		// generic information
+		"/proc/cmdline",
```
marcel-apf, line 106: I see the list is much smaller than the original one. I assume you already "de-duplicated" offline the files?
Maybe we should consider listing here all the files we need and let the code do the de-duplication?
It will be like a documentation of the files we need, otherwise we will have the list in more than one place.
Only a thought.

fromanirh: The list now contains only the files we (KNI) need on top of ghw's, that's why is much smaller.
because of net and PCI topologies, which are host-specific, it's unpractical to have a static list; the only feasible option is to create the list dinamically, and to do so - and to be able to inspect that list, we have the `dump` option.

The code which creates the list and manages the snapshot is vendored, so we can inspect it easily.

fromanirh: re: net and PCI, we have unmerged PRs to address this, see https://github.com/jaypipes/ghw/pull/224

marcel-apf: got it, thanks

PR #562, opened 2021-02-22
Consuming ghw snapshot for the targeted nodes

swatisehgal:
This patch loads the ghw snapshot for the targeted nodes

-  shows examples on how these snapshots can be used to obtain CPU and Topology information.
-  In addition, we add unit tests to validate the patch


pkg/profilecreator/profilecreator.go
```diff
@@ -189,3 +195,21 @@ func GetMCP(mustGatherDirPath, mcpName string) (*machineconfigv1.MachineConfigPo
 	}
 	return &mcp, nil
 }
+
+// LoadSnapshot loads a ghw snapshot corresponding to a node
+func LoadSnapshot(mustGatherDirPath string, node *v1.Node) (*option.Option, error) {
```
fromanirh, line 200: so this doesn't actually load a snapshot, but rather creates the options needed to consume it

pkg/profilecreator/profilecreator_test.go
```diff
@@ -101,3 +102,42 @@ var _ = Describe("PerformanceProfileCreator: Getting Nodes from Must Gather", fu
...
+	var node *v1.Node
+	var err error
+
+	Context("Identifying Nodes Info of the nodes cluster", func() {
+		It("gets the Nodes Info successfully", func() {
+			node = newTestNode("cnfd1-worker-0.fci1.kni.lab.eng.bos.redhat.com")
+			mustGatherDirPath = "../../testdata/must-gather/must-gather.local.directory"
+			mustGatherDirAbsolutePath, err = filepath.Abs(mustGatherDirPath)
+			Expect(err).ToNot(HaveOccurred())
+			snapShotOptions, err := LoadSnapshot(mustGatherDirAbsolutePath, node)
```
fromanirh, line 117: does this require a `sysinfo.tgz` to be added to the source tree? If yes (and I'm not 100% happy with this decision) it should be added in this commit, otherwise the test is broken

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -51,7 +52,23 @@ var rootCmd = &cobra.Command{
 
 		matchedNodes, err := profilecreator.GetMatchedNodes(nodes, labelSelector)
 		for _, node := range matchedNodes {
-			log.Infof("%s is targetted by %s MCP", node.GetName(), mcpName)
+			nodeName := node.GetName()
+			log.Infof("%s is targetted by %s MCP", nodeName, mcpName)
+			snapShotOptions, err := profilecreator.LoadSnapshot(mustGatherDirPath, node)
```
fromanirh, line 57: either we rename `LoadSnapshot` (which is not - it is creating the right options to consume the snapshot) or we add a thin wrapper around ghw to get the API object we care about from snapshot, node, parameters. Something like:

```
handle, err := profilecreator.NewKNI(mustGatherDirPath, node)
...

cpuInfo, err := handle.CPU()
...
topologyInfo, err := handle.Topology()
```
I'm fine with either approaches

pkg/profilecreator/profilecreator_test.go
swatisehgal: We have it already. I had added it in the the testdata in the previous PR as I knew we would be needing it here.

pkg/profilecreator/profilecreator.go
swatisehgal: Ack!

cmd/performance-profile-creator/cmd/root.go
swatisehgal: @marcel-apf Do you have any preference? Is it okay to just update the name to `ConsumeSnapshot`  for now?

pkg/profilecreator/profilecreator_test.go
fromanirh: ok, then I missed it :) sorry!

cmd/performance-profile-creator/cmd/root.go
marcel-apf: Yes, it is, however I prefer the "handle" approach. I am good with both options though.

swatisehgal: Done!

pkg/profilecreator/profilecreator.go
```diff
@@ -196,8 +198,8 @@ func GetMCP(mustGatherDirPath, mcpName string) (*machineconfigv1.MachineConfigPo
 	return &mcp, nil
 }
 
-// LoadSnapshot loads a ghw snapshot corresponding to a node
-func LoadSnapshot(mustGatherDirPath string, node *v1.Node) (*option.Option, error) {
+// NewGHWHandler is a handler to use ghw options corresponding to a node
+func NewGHWHandler(mustGatherDirPath string, node *v1.Node) (*GHWHandler, error) {
```
fromanirh, line 202: nice!

PR #563, opened 2021-02-23
e2e: skip the IRQ load balancing test

cynepco3hahue:
The test constantly fails under d/s CI, the reason that some of the interrupts can not be moved.
It no easy way to fix it, and until we will have some solution we need to skip the test.


fromanirh:
/lgtm
/approve
/hold
better to skip than constantly fail. Let's work in the meantime to the long-term fix. Giving time to other reviewes to join the discussion.

PR #566, opened 2021-02-25
Set the reserved and isolated CPUs in the performance profile

swatisehgal:
Based on the input provided by the user for reserved-cpu-count and
split-reserved-cpus-across-numa we determine the the possible
values for reserved and isolated CPUs in the created performance
profile.
This patch also provides associated Unit tests to validate this feature

### DIFF
--- a/cmd/performance-profile-creator/cmd/root.go
+++ b/cmd/performance-profile-creator/cmd/root.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strconv"
 
 	"github.com/openshift-kni/performance-addon-operators/pkg/profilecreator"
 	log "github.com/sirupsen/logrus"
@@ -31,6 +32,13 @@ import (
 	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+// ProfileData collects and stores all the data needed for profile creation
+type ProfileData struct {
+	isolatedCPUs, reservedCPUs string
+}
+
+var profileData ProfileData
 
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
@@ -48,28 +56,31 @@ var rootCmd = &cobra.Command{
 		if err != nil {
 			return fmt.Errorf("Error obtaining Nodes %s: %v", mcpName, err)
 		}
-
+		splitReservedCPUsAcrossNUMA, err := strconv.ParseBool(cmd.Flag("split-reserved-cpus-across-numa").Value.String())
+		if err != nil {
+			return fmt.Errorf("Error parsing split-reserved-cpus-across-numa flag: %v", err)
+		}
+		reservedCPUCount, err := strconv.Atoi(cmd.Flag("reserved-cpu-count").Value.String())
+		if err != nil {
+			return fmt.Errorf("Error parsing reserved-cpu-count flag: %v", err)
+		}
 		matchedNodes, err := profilecreator.GetMatchedNodes(nodes, labelSelector)
 		for _, node := range matchedNodes {
 			nodeName := node.GetName()
 			log.Infof("%s is targetted by %s MCP", nodeName, mcpName)
 			handle, err := profilecreator.NewGHWHandler(mustGatherDirPath, node)
-			cpuInfo, err := handle.CPU()
+			reservedCPUs, isolatedCPUs, err := handle.GetReservedAndIsolatedCPUs(reservedCPUCount, splitReservedCPUsAcrossNUMA)
 			if err != nil {
-				return fmt.Errorf("Error obtaining CPU Info from GHW snapshot for %s: %v", nodeName, err)
+				return fmt.Errorf("Error obtaining Reserved and Isolated CPUs for %s: %v", nodeName, err)
 			}
-			log.Infof("CPU Info: %v\n", cpuInfo)
-			topologyInfo, err := handle.Topology()
-			if err != nil {
-				return fmt.Errorf("Error obtaining Topology Info from GHW snapshot for %s: %v", nodeName, err)
-			}
-			log.Infof("Topology Info: %v\n", topologyInfo)
+			profileData.reservedCPUs = reservedCPUs
+			profileData.isolatedCPUs = isolatedCPUs
 		}
 		return nil
 	},
 	Run: func(cmd *cobra.Command, args []string) {
 		profileName := cmd.Flag("profile-name").Value.String()
-		createProfile(profileName)
+		createProfile(profileName, profileData)
 	},
 }
 
@@ -83,15 +94,15 @@ func Execute() {
 }
 
 type profileCreatorArgs struct {
-	powerConsumptionMode string
-	mustGatherDirPath    string
-	profileName          string
-	reservedCPUCount     int
-	splitCPUsAcrossNUMA  bool
-	disableHT            bool
-	rtKernel             bool
-	userLevelNetworking  bool
-	mcpName              string
+	powerConsumptionMode        string
+	mustGatherDirPath           string
+	profileName                 string
+	reservedCPUCount            int
+	splitReservedCPUsAcrossNUMA bool
+	disableHT                   bool
+	rtKernel                    bool
+	userLevelNetworking         bool
+	mcpName                     string
 }
 
 func init() {
@@ -99,9 +110,9 @@ func init() {
 	log.SetOutput(os.Stderr)
 	rootCmd.PersistentFlags().IntVarP(&args.reservedCPUCount, "reserved-cpu-count", "R", 0, "Number of reserved CPUs (required)")
 	rootCmd.MarkPersistentFlagRequired("reserved-cpu-count")
+	rootCmd.PersistentFlags().BoolVarP(&args.splitReservedCPUsAcrossNUMA, "split-reserved-cpus-across-numa", "S", true, "Split the Reserved CPUs across NUMA nodes")
 	rootCmd.PersistentFlags().StringVarP(&args.mcpName, "mcp-name", "T", "worker-cnf", "MCP name corresponding to the target machines (required)")
 	rootCmd.MarkPersistentFlagRequired("mcp-name")
-	rootCmd.PersistentFlags().BoolVarP(&args.splitCPUsAcrossNUMA, "split-cpus-across-numa", "S", true, "Split the CPUs across NUMA nodes")
 	rootCmd.PersistentFlags().BoolVarP(&args.disableHT, "disable-ht", "H", false, "Disable Hyperthreading")
 	rootCmd.PersistentFlags().BoolVarP(&args.rtKernel, "rt-kernel", "K", true, "Enable Real Time Kernel (required)")
 	rootCmd.MarkPersistentFlagRequired("rt-kernel")
@@ -116,8 +127,10 @@ func init() {
 	// 2) e.g.check to make sure that power consumption string is in {CSTATE NO_CSTATE IDLE_POLL}
 }
 
-func createProfile(profileName string) {
+func createProfile(profileName string, profileData ProfileData) {
 
+	reserved := performancev2.CPUSet(profileData.reservedCPUs)
+	isolated := performancev2.CPUSet(profileData.isolatedCPUs)
 	// TODO: Get the name from MCP if not specified in the command line arguments
 	profile := &performancev2.PerformanceProfile{
 		TypeMeta: metav1.TypeMeta{
@@ -128,6 +141,10 @@ func createProfile(profileName string) {
 			Name: profileName,
 		},
 		Spec: performancev2.PerformanceProfileSpec{
+			CPU: &performancev2.CPU{
+				Isolated: &isolated,
+				Reserved: &reserved,
+			},
 			RealTimeKernel: &performancev2.RealTimeKernel{
 				Enabled: pointer.BoolPtr(true),
 			},
--- a/pkg/profilecreator/profilecreator.go
+++ b/pkg/profilecreator/profilecreator.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sort"
 
 	"github.com/jaypipes/ghw"
 	"github.com/jaypipes/ghw/pkg/cpu"
@@ -31,6 +32,7 @@ import (
 
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/component-helpers/scheduling/corev1"
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	v1 "k8s.io/api/core/v1"
@@ -230,4 +232,65 @@ func (ghwHandler GHWHandler) CPU() (*cpu.Info, error) {
 // Topology returns a TopologyInfo struct that contains information about the Topology on the host system
 func (ghwHandler GHWHandler) Topology() (*topology.Info, error) {
 	return ghw.Topology(ghwHandler.snapShotOptions)
+}
+
+// GetReservedAndIsolatedCPUs returns Reserved and Isolated CPUs
+func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, splitReservedCPUsAcrossNUMA bool) (string, string, error) {
+	if reservedCPUCount < 0 {
+		return "", "", fmt.Errorf("Specified eservered CPU count is negative, please specify it correctly")
+	}
+	topologyInfo, err := ghwHandler.Topology()
+	if err != nil {
+		return "", "", fmt.Errorf("Error obtaining Topology Info from GHW snapshot: %v", err)
+	}
+	reservedCPUSet := cpuset.NewBuilder()
+	isolatedCPUSet := cpuset.NewBuilder()
+	var numaNodeNum int
+	if splitReservedCPUsAcrossNUMA {
+		numaNodeNum = len(topologyInfo.Nodes)
+	} else {
+		numaNodeNum = 1
+	}
+
+	var max = 0
+	reservedPerNuma := reservedCPUCount / numaNodeNum
+	remainder := reservedCPUCount % numaNodeNum
+	for numaID, node := range topologyInfo.Nodes {
+		if splitReservedCPUsAcrossNUMA {
+			if remainder != 0 {
+				max = (numaID+1)*reservedPerNuma + (numaNodeNum - remainder)
+				remainder--
+			} else {
+				max = max + reservedPerNuma
+			}
+		} else {
+			max = reservedCPUCount
+		}
+		//sorting by comparing the first logical core of processor cores
+		sort.Slice(node.Cores, func(i, j int) bool {
+			//sorting the logical cores a processor
+			sort.Slice(node.Cores[i].LogicalProcessors, func(x, y int) bool {
+				return node.Cores[i].LogicalProcessors[x] < node.Cores[i].LogicalProcessors[y]
+			})
+			sort.Slice(node.Cores[j].LogicalProcessors, func(x, y int) bool {
+				return node.Cores[j].LogicalProcessors[x] < node.Cores[j].LogicalProcessors[y]
+			})
+			return node.Cores[i].LogicalProcessors[0] < node.Cores[j].LogicalProcessors[0]
+		})
+		for _, cores := range node.Cores {
+			for _, core := range cores.LogicalProcessors {
+				if reservedCPUSet.Result().Size() < max {
+					reservedCPUSet.Add(core)
+				} else {
+					isolatedCPUSet.Add(core)
+				}
+			}
+		}
+	}
+
+	log.Infof("reservedCPUs: %v len(reservedCPUs): %d \n", reservedCPUSet.Result().String(), reservedCPUSet.Result().Size())
+	log.Infof("isolatedCPUs: %v len(isolatedCPUs): %d \n", isolatedCPUSet.Result().String(), isolatedCPUSet.Result().Size())
+
+	return reservedCPUSet.Result().String(), isolatedCPUSet.Result().String(), nil
+
 }
--- a/pkg/profilecreator/profilecreator_test.go
+++ b/pkg/profilecreator/profilecreator_test.go
@@ -140,3 +140,54 @@ var _ = Describe("PerformanceProfileCreator: Consuming GHW Snapshot from Must Ga
 
 	})
 })
+
+var _ = Describe("PerformanceProfileCreator: Populating Reserved and Isolated CPUs in Performance Profile", func() {
+	var mustGatherDirPath, mustGatherDirAbsolutePath string
+	var node *v1.Node
+	var handle *GHWHandler
+	var splitReservedCPUsAcrossNUMA bool
+	var reservedCPUCount int
+	var err error
+
+	BeforeEach(func() {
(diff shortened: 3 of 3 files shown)


cmd/performance-profile-creator/cmd/root.go
```diff
@@ -32,6 +33,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// ProfileData collects and stores all the data needed for profile creation
+type ProfileData struct {
+	isolatedCPUs, reservedCPUs string
+}
+
+var profileData ProfileData
```
fromanirh, line 41: why global? can't be moved in the `PreRunE` body?

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -83,25 +94,25 @@ func Execute() {
...
+	userLevelNetworking         bool
+	mcpName                     string
 }
 
 func init() {
 	args := &profileCreatorArgs{}
 	log.SetOutput(os.Stderr)
 	rootCmd.PersistentFlags().IntVarP(&args.reservedCPUCount, "reserved-cpu-count", "R", 0, "Number of reserved CPUs (required)")
 	rootCmd.MarkPersistentFlagRequired("reserved-cpu-count")
+	rootCmd.PersistentFlags().BoolVarP(&args.splitReservedCPUsAcrossNUMA, "split-reserved-cpus-across-numa", "S", true, "Split the Reserved CPUs across NUMA nodes")
```
fromanirh, line 113: is this a synonim for packed/spread reserved CPU distribution?

pkg/profilecreator/profilecreator.go
```diff
@@ -231,3 +233,64 @@ func (ghwHandler GHWHandler) CPU() (*cpu.Info, error) {
...
+		}
+		//sorting by comparing the first logical core of processor cores
+		sort.Slice(node.Cores, func(i, j int) bool {
+			//sorting the logical cores a processor
+			sort.Slice(node.Cores[i].LogicalProcessors, func(x, y int) bool {
+				return node.Cores[i].LogicalProcessors[x] < node.Cores[i].LogicalProcessors[y]
+			})
+			sort.Slice(node.Cores[j].LogicalProcessors, func(x, y int) bool {
+				return node.Cores[j].LogicalProcessors[x] < node.Cores[j].LogicalProcessors[y]
+			})
```
fromanirh, line 277: three sorts look suspicious. Can we squash the two inner sorts together somehow?

pkg/profilecreator/profilecreator.go
```diff
@@ -231,3 +233,64 @@ func (ghwHandler GHWHandler) CPU() (*cpu.Info, error) {
...
+			return node.Cores[i].LogicalProcessors[0] < node.Cores[j].LogicalProcessors[0]
+		})
+		for _, cores := range node.Cores {
+			for _, core := range cores.LogicalProcessors {
+				if reservedCPUSet.Result().Size() < max {
+					reservedCPUSet.Add(core)
+				} else {
+					isolatedCPUSet.Add(core)
```
fromanirh, line 285: I wonder if it would be clearer to only pick the reservedCPUSet and to compute the isolated set once, outside the loop, as difference like `isolatedCPUSet = totalCPUSet - reservedCPUSet`

pkg/profilecreator/profilecreator.go
```diff
@@ -231,3 +233,64 @@ func (ghwHandler GHWHandler) CPU() (*cpu.Info, error) {
...
+				}
+			}
+		}
+	}
+
+	log.Infof("reservedCPUs: %v len(reservedCPUs): %d \n", reservedCPUSet.Result().String(), reservedCPUSet.Result().Size())
+	log.Infof("isolatedCPUs: %v len(isolatedCPUs): %d \n", isolatedCPUSet.Result().String(), isolatedCPUSet.Result().Size())
```
fromanirh, line 292: In general it's better to squash related log lines together, if possible.

pkg/profilecreator/profilecreator_test.go
```diff
@@ -140,3 +140,54 @@ var _ = Describe("PerformanceProfileCreator: Consuming GHW Snapshot from Must Ga
...
+	var reservedCPUCount int
+	var err error
+
+	BeforeEach(func() {
+		node = newTestNode("cnfd1-worker-0.fci1.kni.lab.eng.bos.redhat.com")
+		mustGatherDirPath = "../../testdata/must-gather/must-gather.local.directory"
+	})
+	Context("Check if reserved and isolated CPUs are properly populated in the performance profile", func() {
+		It("Ensure reserved CPUs populated are correctly when splitReservedCPUsAcrossNUMA is disabled", func() {
+			reservedCPUCount = 20
```
fromanirh, line 158: is this a special number of just a random number? (could have been 19, 21, 30...?) If it's just a random number please add a small comment like
```
// random number, no special meaning
```

pkg/profilecreator/profilecreator_test.go
```diff
@@ -140,3 +140,54 @@ var _ = Describe("PerformanceProfileCreator: Consuming GHW Snapshot from Must Ga
...
+			Expect(err).ToNot(HaveOccurred())
+			handle, err = NewGHWHandler(mustGatherDirAbsolutePath, node)
+			Expect(err).ToNot(HaveOccurred())
+			reservedCPUs, isolatedCPUs, err := handle.GetReservedAndIsolatedCPUs(reservedCPUCount, splitReservedCPUsAcrossNUMA)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(reservedCPUs).To(Equal("0,2,4,6,8,10,12,14,16,18,40,42,44,46,48,50,52,54,56,58"))
+			Expect(isolatedCPUs).To(Equal("1,3,5,7,9,11,13,15,17,19-39,41,43,45,47,49,51,53,55,57,59-79"))
+		})
+		It("Ensure reserved CPUs populated are correctly when splitReservedCPUsAcrossNUMA is enabled", func() {
+			reservedCPUCount = 20
```
fromanirh, line 170: same

fromanirh (commented):
initial review. I need to spend some time to digest the allocation logic.

cmd/performance-profile-creator/cmd/root.go
swatisehgal: I would really like to do this in a bit better and am open to suggestions here but reason for the current implementation is becase handle.GetReservedAndIsolatedCPUs is called in PreRunE where we populate the profileData and we need this data in the Run() where call to createProfile(profileName, **profileData**)

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Yes that is correct! Do we want to call the argument `spread-reserved-cpus-across-numa` instead?

cmd/performance-profile-creator/cmd/root.go
fromanirh: OK, silly questions first: can't we merge `PreRunE` and `Run` in just one `RunE`? I get why we need the -E variant, I'm still not sure why we need PreRun + Run.

pkg/profilecreator/profilecreator.go
swatisehgal: I can try but the idea is that we first individually sort the Logical Processors (hence the two inner sort) based on CPUID and then then sort the cores by comparing the smallest logical core ID so I think we need the two inner sort but I can try to figure out better ways of doing this.

pkg/profilecreator/profilecreator.go
swatisehgal: Sure, that is no problem. I can do that. I had a version where I did it that way but thought populating both the CPUSets was more readable. Will update.

pkg/profilecreator/profilecreator.go
swatisehgal: Ack

pkg/profilecreator/profilecreator_test.go
swatisehgal: Sure, makes sense, will do.

pkg/profilecreator/profilecreator_test.go
swatisehgal: Same as above :)

cmd/performance-profile-creator/cmd/root.go
marcel-apf: It looks much better since it returns error making it the perfect place to validate the input.
We could (probably) use only Run, but then why use cobra at all if we don't use their features?

Also while I am against global variables, every piece of software has a "context" that can be passed between different subsystems.  If cobra supports such context, then fine, we should use the cobra context to pass our data between the calls. Otherwise using a global context is also acceptable for a small piece of software as long as the "global context" is private to the package/file and not accessible from all the code.  

On the other hand, if Cobra can't help us to complete our task, it becomes an obstacle rather than help us. I am not against Cobra, it seems cool.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -48,28 +56,31 @@ var rootCmd = &cobra.Command{
...
+		if err != nil {
+			return fmt.Errorf("Error parsing reserved-cpu-count flag: %v", err)
+		}
 		matchedNodes, err := profilecreator.GetMatchedNodes(nodes, labelSelector)
 		for _, node := range matchedNodes {
 			nodeName := node.GetName()
 			log.Infof("%s is targetted by %s MCP", nodeName, mcpName)
 			handle, err := profilecreator.NewGHWHandler(mustGatherDirPath, node)
-			cpuInfo, err := handle.CPU()
+			reservedCPUs, isolatedCPUs, err := handle.GetReservedAndIsolatedCPUs(reservedCPUCount, splitReservedCPUsAcrossNUMA)
```
marcel-apf, line 72: It looks like this is business logic and shouldn't be part of the input "validation", so maybe we don't need the "context"...

cmd/performance-profile-creator/cmd/root.go
marcel-apf: But it seems we may not need the context after all, see bellow

pkg/profilecreator/profilecreator.go
```diff
@@ -231,3 +233,64 @@ func (ghwHandler GHWHandler) CPU() (*cpu.Info, error) {
...
+	var numaNodeNum int
+	if splitReservedCPUsAcrossNUMA {
+		numaNodeNum = len(topologyInfo.Nodes)
+	} else {
+		numaNodeNum = 1
+	}
+
+	var max = 0
+	reservedPerNuma := reservedCPUCount / numaNodeNum
+	remainder := reservedCPUCount % numaNodeNum
```
marcel-apf, line 257: Can't we error out if we have remainder != 0 ? That means we can't equally divide the reserved CPUs between the NUMA nodes, so we can't accomplish the task. I will add this check to the validation.

cmd/performance-profile-creator/cmd/root.go
fromanirh: uhm, good point.

pkg/profilecreator/profilecreator.go
```diff
@@ -231,3 +233,64 @@ func (ghwHandler GHWHandler) CPU() (*cpu.Info, error) {
...
+			} else {
+				max = max + reservedPerNuma
+			}
+		} else {
+			max = reservedCPUCount
+		}
+		//sorting by comparing the first logical core of processor cores
+		sort.Slice(node.Cores, func(i, j int) bool {
+			//sorting the logical cores a processor
+			sort.Slice(node.Cores[i].LogicalProcessors, func(x, y int) bool {
```
marcel-apf, line 272: It seems you are sorting everything "in place".
Maybe you can go first over the all the cores and sort their logical processors, then sort the cores by the first logical processor?
Also theoretically you may have the same problem with the NUMA nodes once you have more than 10 NUMAs.

pkg/profilecreator/profilecreator.go
marcel-apf: +1

pkg/profilecreator/profilecreator.go
swatisehgal: So the what I have introduced in the code is even if the reserved CPUs cannot be divided between the NUMA nodes they can still be spread as much as possible. E.g. In case a system has 4 NUMA Nodes and the user wants 15 reserved CPUs, we would get 4 cpus from 3 NUMA nodes and 3 from 1 NUMA Node. The user might not know how many NUMA nodes are on the system and could request an arbitrary number that is not divisible by the number of NUMA nodes and code would be gracefully handle that case without returning an error.

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Okay, what I am going to do is remove Run and move the profile creation to RunE instead.

pkg/profilecreator/profilecreator.go
marcel-apf: I am usually against the "best-effor": approach. If the user will select a "wrong" number we will error out and explain. Otherwise the user MIGHT think he reserved enough CPUs per NUMA node while he didn't...

swatisehgal: Okay, will update as per your suggestion.

pkg/profilecreator/profilecreator.go
```diff
@@ -229,22 +229,43 @@ func (ghwHandler GHWHandler) CPU() (*cpu.Info, error) {
 	return ghw.CPU(ghwHandler.snapShotOptions)
 }
 
-// Topology returns a TopologyInfo struct that contains information about the Topology on the host system
-func (ghwHandler GHWHandler) Topology() (*topology.Info, error) {
-	return ghw.Topology(ghwHandler.snapShotOptions)
+// SortedTopology returns a TopologyInfo struct that contains information about the Topology on the host system
```
fromanirh, line 232: ...sorted how? Meaning, how the end result should look like? numa nodes sorted by their cell ID, increasing, then...

please add a couple lines to document this - it is part of the function contract.

pkg/profilecreator/profilecreator.go
```diff
@@ -229,22 +229,43 @@ func (ghwHandler GHWHandler) CPU() (*cpu.Info, error) {
...
+func (ghwHandler GHWHandler) SortedTopology() (*topology.Info, error) {
+	topologyInfo, err := ghw.Topology(ghwHandler.snapShotOptions)
+	if err != nil {
+		return nil, fmt.Errorf("Error obtaining Topology Info from GHW snapshot: %v", err)
+	}
+	// Sorting the NUMA nodes
```
fromanirh, line 238: redundant comment (same below for line 243 and 249)

pkg/profilecreator/profilecreator.go
```diff
@@ -229,22 +229,43 @@ func (ghwHandler GHWHandler) CPU() (*cpu.Info, error) {
...
+			sort.Slice(core.LogicalProcessors, func(x, y int) bool {
+				return core.LogicalProcessors[x] < core.LogicalProcessors[y]
+			})
+		}
+		//Sorting by comparing the first logical core of processor cores
+		sort.Slice(node.Cores, func(i, j int) bool {
+			return node.Cores[i].LogicalProcessors[0] < node.Cores[j].LogicalProcessors[0]
+		})
+	}
+	return topologyInfo, nil
```
fromanirh, line 254: this function deserves a (couple of) unit test - can be added later.

pkg/profilecreator/profilecreator.go
```diff
@@ -266,31 +295,38 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
 
-	log.Infof("reservedCPUs: %v len(reservedCPUs): %d \n", reservedCPUSet.Result().String(), reservedCPUSet.Result().Size())
-	log.Infof("isolatedCPUs: %v len(isolatedCPUs): %d \n", isolatedCPUSet.Result().String(), isolatedCPUSet.Result().Size())
+}
+func (ghwHandler GHWHandler) isHyperthreadingEnabled() (bool, error) {
+	cpuInfo, err := ghwHandler.CPU()
+	if err != nil {
+		return false, fmt.Errorf("Error obtaining CPU Info from GHW snapshot: %v", err)
+	}
+	return contains(cpuInfo.Processors[0].Capabilities, "ht"), nil
```
fromanirh, line 320: this is ok. It however deserves a comment about why we pick the 0th processor, something like

"since there is no way to disable flags per-processor (not system wide) we check the flags of the first available processor"

pkg/profilecreator/profilecreator.go
```diff
@@ -266,31 +295,38 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
-			//sorting the logical cores a processor
-			sort.Slice(node.Cores[i].LogicalProcessors, func(x, y int) bool {
-				return node.Cores[i].LogicalProcessors[x] < node.Cores[i].LogicalProcessors[y]
-			})
-			sort.Slice(node.Cores[j].LogicalProcessors, func(x, y int) bool {
-				return node.Cores[j].LogicalProcessors[x] < node.Cores[j].LogicalProcessors[y]
-			})
-			return node.Cores[i].LogicalProcessors[0] < node.Cores[j].LogicalProcessors[0]
-		})
+		if max%2 != 0 && htEnabled {
```
fromanirh, line 298: the more I read about this allocation loop, the more I wonder if it would'nt be much more readable to just have two completely separate helper functions, one per allocation strategy (splitReservedCPUsAcrossNUMA=false/true -> two strategies), each one with his clear and nice allocation loop.

Can be done later, however.

fromanirh (commented):
I think looks good now, possible improvements inside but pretty much all of them can be postponed.

fromanirh:
to move forward, let's figure out if/how address the comments I added, then from my side it's lgtm

pkg/profilecreator/profilecreator.go
marcel-apf: Yes, sorted by the numa ids and cpu ids, is better to add the info

pkg/profilecreator/profilecreator.go
```diff
@@ -266,31 +295,38 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
+			return "", "", fmt.Errorf("Can't allocatable odd number of CPUs from a NUMA Node")
+		}
 		for _, cores := range node.Cores {
 			for _, core := range cores.LogicalProcessors {
```
marcel-apf, line 302: I see 2 variables with the same name, "cores", while "it works" maybe it worth renaming one of them

pkg/profilecreator/profilecreator.go
marcel-apf: A following implementation will leverage the /sys/devices/system/cpu/smt/active file which is the "standard" way to query HT.

pkg/profilecreator/profilecreator_test.go
```diff
@@ -188,6 +188,26 @@ var _ = Describe("PerformanceProfileCreator: Populating Reserved and Isolated CP
 			_, _, err := handle.GetReservedAndIsolatedCPUs(reservedCPUCount, splitReservedCPUsAcrossNUMA)
 			Expect(err).To(HaveOccurred())
 		})
+		It("Errors out in case splitReservedCPUsAcrossNUMA is enabled and number of reserved CPUs per number of NUMA nodes are odd", func() {
```
marcel-apf, line 191: If HT is enabled, right?

pkg/profilecreator/profilecreator_test.go
```diff
@@ -188,6 +188,26 @@ var _ = Describe("PerformanceProfileCreator: Populating Reserved and Isolated CP
...
+			reservedCPUCount = 21 // random number which results in a CPU split per NUMA node (11 + 10 in this case) such that odd number of reserved CPUs (11) have to be allocated from a NUMA node
+			splitReservedCPUsAcrossNUMA = true
+			mustGatherDirAbsolutePath, err = filepath.Abs(mustGatherDirPath)
+			Expect(err).ToNot(HaveOccurred())
+			handle, err = NewGHWHandler(mustGatherDirAbsolutePath, node)
+			Expect(err).ToNot(HaveOccurred())
+			_, _, err := handle.GetReservedAndIsolatedCPUs(reservedCPUCount, splitReservedCPUsAcrossNUMA)
+			Expect(err).To(HaveOccurred())
+		})
+		It("Errors out in case splitReservedCPUsAcrossNUMA is disabled and number of reserved CPUs are odd", func() {
```
marcel-apf, line 201: Same as above, if HT is enabled

marcel-apf (commented):
Looks good and ready to be merged, thank you!
Marcel

pkg/profilecreator/profilecreator.go
fromanirh: however I was doublechecking this and we need to be at least aware of the fine details between firmware settings (aka HT enabled/disabled in BIOS or UEFI) and kernel settings.
1. the HT flag, if present, is telling us that the HW is *ht capable*
2. the "nosmt" kernel parameters seems (didn't dive deep enough yet) to do something very close to just offline the hyperthread sibilings.

So in practice:
1. if the BIOS and the kernel aligns, checking for the presence of HT flags tells us if the system is running with SMT enabled - or not
2. if the BIOS and the kernel DO NOT align, things get interesting. The only meaningful case is: HT enabled in firmware, `nosmt` given. In this case, this function wrongly reports the HT state. For userspace, HT is enabled only if ALL hardware, firmware and kernel allow + enable SMT. So in this case the computation will be wrong.

But this is actually no different from the corner case on which everything is enabled, yet some cpu cores are offlined for whatever reason.

So reading `/sys/devices/system/cpu/smt/active` seems to be the single action with more ROI, but this will require some changes in how PPC handles the snapshots.

marcel-apf: nosmt kernel arg does the same as editing `echo [on|off] > /sys/devices/system/cpu/smt/control`, enables a single thread per core.
BTW, the option we have in PPC is "disable-ht=true/false" and not "ht=true/false" because of the reasons you mentioned.
We will need to add this file to the ghw snapshot.

fromanirh: done: https://github.com/openshift-kni/performance-addon-operators/pull/567
I'm still not sure reading `/sys/devices/system/cpu/smt/active` alone is everything we need, but I concur that is probably the single best option.

pkg/profilecreator/profilecreator.go
swatisehgal: Will add a comment clarifying this.

pkg/profilecreator/profilecreator.go
swatisehgal: Ack

pkg/profilecreator/profilecreator.go
swatisehgal: Sure, will add the tests shortly.

pkg/profilecreator/profilecreator.go
swatisehgal: Sure, will add a comment.

marcel-apf: @fromanirh so fast! thanks!

pkg/profilecreator/profilecreator.go
swatisehgal: Sure, the allocation logic can certainly be made more readable! I will add a TODO along with the suggested approach and do the refactoring in a separate PR.

pkg/profilecreator/profilecreator.go
swatisehgal: Sure, will fix.

pkg/profilecreator/profilecreator_test.go
swatisehgal: Yep! Sorry that was a mistake, will fix.

fromanirh:
/lgtm
/hold
please remove hold when a second lgtm comes

PR #567, opened 2021-03-01
gather_sysinfo: collect SMT state from sysfs

fromanirh:
Explicitely collect this information from sysfs.
Another option is to check the processor flags: if "ht" is in there,
then SMT is enabled.

### DIFF
--- a/tools/gather-sysinfo/gather-sysinfo.go
+++ b/tools/gather-sysinfo/gather-sysinfo.go
@@ -109,6 +109,8 @@ func kniExpectedCloneContent() []string {
 		"/proc/irq/default_smp_affinity",
 		"/proc/irq/*/*affinity_list",
 		"/proc/irq/*/node",
+		// KNI-specific CPU infos:
+		"/sys/devices/system/cpu/smt/active",
 		// BIOS/firmware versions
 		"/sys/class/dmi/id/bios*",
 		"/sys/class/dmi/id/product_family",


swatisehgal (approved):
/lgtm

fromanirh:
for the record (https://github.com/openshift-kni/performance-addon-operators/pull/566#discussion_r585332692):
I was  was doublechecking the above claim ("Another option is to check the processor flags: if "ht" is in there,
then SMT is enabled.") and we need to be at least aware of the fine details between firmware settings (aka HT enabled/disabled in BIOS or UEFI) and kernel settings.

    the HT flag, if present, is telling us that the HW is ht capable
    the "nosmt" kernel parameters seems (didn't dive deep enough yet) to do something very close to just offline the hyperthread sibilings.

So in practice:

    if the BIOS and the kernel aligns, checking for the presence of HT flags tells us if the system is running with SMT enabled - or not
    if the BIOS and the kernel DO NOT align, things get interesting. The only meaningful case is: HT enabled in firmware, nosmt given. In this case, this function wrongly reports the HT state. For userspace, HT is enabled only if ALL hardware, firmware and kernel allow + enable SMT. So in this case the computation will be wrong.

But this is actually no different from the corner case on which everything is enabled, yet some cpu cores are offlined for whatever reason.

So reading /sys/devices/system/cpu/smt/active seems to be the single action with more ROI, but this will require some changes in how PPC handles the snapshots.

PR #568, opened 2021-03-02
Make the allocation logic more readable with helper functions

swatisehgal:
As per the suggestions in the review of https://github.com/openshift-kni/performance-addon-operators/pull/566
we are refactoring the allocation logic to make it more readable by using
separate helper functions, one per allocation strategy
corresponding to splitReservedCPUsAcrossNUMA=false/true

### DIFF
--- a/pkg/profilecreator/profilecreator.go
+++ b/pkg/profilecreator/profilecreator.go
@@ -260,15 +260,26 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
 	if err != nil {
 		return "", "", fmt.Errorf("Error obtaining Topology Info from GHW snapshot: %v", err)
 	}
-	if splitReservedCPUsAcrossNUMA {
-		return ghwHandler.getCPUsSplitAcrossNUMA(reservedCPUCount, topologyInfo.Nodes)
+	htEnabled, err := ghwHandler.isHyperthreadingEnabled()
+	if err != nil {
+		return "", "", fmt.Errorf("Error determining if Hyperthreading is enabled or not: %v", err)
 	}
-	return ghwHandler.getCPUsSequentially(reservedCPUCount, topologyInfo.Nodes)
+	if splitReservedCPUsAcrossNUMA {
+		return ghwHandler.getCPUsSplitAcrossNUMA(reservedCPUCount, htEnabled, topologyInfo.Nodes)
+	}
+	return ghwHandler.getCPUsSequentially(reservedCPUCount, htEnabled, topologyInfo.Nodes)
 }
 
 // getCPUsSplitAcrossNUMA returns Reserved and Isolated CPUs split across NUMA nodes
-func (ghwHandler GHWHandler) getCPUsSplitAcrossNUMA(reservedCPUCount int, topologyInfoNodes []*topology.Node) (string, string, error) {
-	totalCPUSet := cpuset.NewBuilder()
+// We identify the right number of CPUs that need to be allocated per NUMA node, meaning reservedPerNuma + (the additional number based on the remainder and the NUMA node)
+// E.g. If the user requests 15 reserved cpus and we have 4 numa nodes, we find reservedPerNuma in this case is 3 and remainder = 3.
+// For each numa node we find a max which keeps track of the cumulative resources that should be allocated for each NUMA node:
+// max = (numaID+1)*reservedPerNuma + (numaNodeNum - remainder)
+// For NUMA node 0 max = (0+1)*3 + 4-3 = 4 remainder is decremented => remainder is 2
+// For NUMA node 1 max = (1+1)*3 + 4-2 = 8 remainder is decremented => remainder is 1
+// For NUMA node 2 max = (2+1)*3 + 4-2 = 12 remainder is decremented => remainder is 0
+// For NUMA Node 3 remainder = 0 so max = 12 + 3 = 15.
+func (ghwHandler GHWHandler) getCPUsSplitAcrossNUMA(reservedCPUCount int, htEnabled bool, topologyInfoNodes []*topology.Node) (string, string, error) {
 	reservedCPUSet := cpuset.NewBuilder()
 	numaNodeNum := len(topologyInfoNodes)
 	max := 0
@@ -277,11 +288,6 @@ func (ghwHandler GHWHandler) getCPUsSplitAcrossNUMA(reservedCPUCount int, topolo
 	if remainder != 0 {
 		log.Warnf("The reserved CPUs cannot be split equally across NUMA Nodes")
 	}
-	htEnabled, err := ghwHandler.isHyperthreadingEnabled()
-	if err != nil {
-		return "", "", fmt.Errorf("Error determining if Hyperthreading is enabled or not: %v", err)
-	}
-
 	for numaID, node := range topologyInfoNodes {
 		if remainder != 0 {
 			max = (numaID+1)*reservedPerNuma + (numaNodeNum - remainder)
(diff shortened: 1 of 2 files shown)


pkg/profilecreator/profilecreator.go
```diff
@@ -312,6 +307,34 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
 
+// getCPUsSequentially returns Reserved and Isolated CPUs sequentially
+func (ghwHandler GHWHandler) getCPUsSequentially(reservedCPUCount int, topologyInfoNodes []*topology.Node) (string, string, error) {
+	totalCPUSet := cpuset.NewBuilder()
+	reservedCPUSet := cpuset.NewBuilder()
+	max := reservedCPUCount
+	htEnabled, err := ghwHandler.isHyperthreadingEnabled()
+	if err != nil {
+		return "", "", fmt.Errorf("Error determining if Hyperthreading is enabled or not: %v", err)
+	}
```
fromanirh, line 318: this block should be moved in the calling site and both helper functions should gain a `htEnabled bool` argument

pkg/profilecreator/profilecreator.go
```diff
@@ -312,6 +307,34 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
 
 }
 
+// getCPUsSequentially returns Reserved and Isolated CPUs sequentially
+func (ghwHandler GHWHandler) getCPUsSequentially(reservedCPUCount int, topologyInfoNodes []*topology.Node) (string, string, error) {
+	totalCPUSet := cpuset.NewBuilder()
```
fromanirh, line 312: this path (and PPC in general) isn't a hot path, plus the fist iteration should aim for clarity rather than raw speed so I'd go a step or two down the path of clarity and add another private helper function
```
func totalCPUSetFromTopology(topologyInfoNodes []*topology.Node) cpuset.CPUSet {
    totalCPUSet := cpuset.NewBuilder()
	for _, node := range topologyInfoNodes {
		for _, processorCores := range node.Cores {
			for _, core := range processorCores.LogicalProcessors {
				totalCPUSet.Add(core)
			}
		}
	}
    return totalCPUSet.Result()
}
```
And use it in both helpers. So the explicit loops in the helpers only concern about allocating reserved cores maximizing clarity.

pkg/profilecreator/profilecreator.go
```diff
@@ -281,18 +282,12 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
-	// (splitReservedCPUsAcrossNUMA=false/true -> two strategies) each one with its clear and nice allocation loop
-	for numaID, node := range topologyInfo.Nodes {
-		if splitReservedCPUsAcrossNUMA {
-			if remainder != 0 {
-				max = (numaID+1)*reservedPerNuma + (numaNodeNum - remainder)
-				remainder--
-			} else {
-				max = max + reservedPerNuma
-			}
+	for numaID, node := range topologyInfoNodes {
```
fromanirh, line 285: if we add the `totalCPUSetFromTopology` helper I suggest below, we can reason about the expanded cpu list here and probably use a clearer logic. Let's explore.

pkg/profilecreator/profilecreator.go
```diff
@@ -312,6 +307,34 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
+	reservedCPUSet := cpuset.NewBuilder()
+	max := reservedCPUCount
+	htEnabled, err := ghwHandler.isHyperthreadingEnabled()
+	if err != nil {
+		return "", "", fmt.Errorf("Error determining if Hyperthreading is enabled or not: %v", err)
+	}
+	if max%2 != 0 && htEnabled {
+		return "", "", fmt.Errorf("Can't allocatable odd number of CPUs from a NUMA Node")
+	}
+	for _, node := range topologyInfoNodes {
```
fromanirh, line 322: if we add the `totalCPUSetFromTopology` helper I suggest below, we can reason about the expanded cpu list here and probably use a clearer logic. In this case, if I'm not mistakens, it boils down to just pick the first `reservedCPUSet` items in the sorted slice. Right?

pkg/profilecreator/profilecreator.go
```diff
@@ -281,18 +282,12 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
-			} else {
-				max = max + reservedPerNuma
-			}
+	for numaID, node := range topologyInfoNodes {
+		if remainder != 0 {
+			max = (numaID+1)*reservedPerNuma + (numaNodeNum - remainder)
+			remainder--
 		} else {
-			max = reservedCPUCount
+			max = max + reservedPerNuma
```
fromanirh, line 290: the only remaining possible improvement is this block (286-291) which we may want to simplify reorganizing the body of the function. The core idea should be:

0. sort cores per-numa node by core id, virtual thread id
1. reserve the first `reservedPerNuma` cores from each numa node 
2. we are left with `remainder` cores to allocate. Iterate over numa node, reserving one core from each up until we reach `remainder`. Right?

fromanirh (commented):
left comments about _all_ possible improvements.

Feel free to address only some of them.

Ideally, the allocation logic should really pop out from this code - it's one of the core algorhytms in PPC after all - but we can get there iteratively, it depends on how much time we can/should spend in this PR.

pkg/profilecreator/profilecreator.go
```diff
@@ -312,6 +307,34 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
+		return "", "", fmt.Errorf("Error determining if Hyperthreading is enabled or not: %v", err)
+	}
+	if max%2 != 0 && htEnabled {
+		return "", "", fmt.Errorf("Can't allocatable odd number of CPUs from a NUMA Node")
+	}
+	for _, node := range topologyInfoNodes {
+		for _, processorCores := range node.Cores {
+			for _, core := range processorCores.LogicalProcessors {
+				totalCPUSet.Add(core)
+				if reservedCPUSet.Result().Size() < max {
```
fromanirh, line 326: this can be a head-scratcher. Per `cpuset.Builder` docs, `Result()` should be the last operation performed in a builder, meaning that after a `Result()` call, all following `Add()` calls should be silently ignored.

However please note that the method receivers of the `Builder` type are all by value (https://tour.golang.org/methods/8). This means that:
1. when `Result` is called, a *copy* of `reservedCPUSet` is passed as method receiver of `Result`
2. `Result` marks the *copy* it received (not the original object, `reservedCPUSet` in our case) as done (https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/cpuset/cpuset.go#L60)
3. We call `Size` on the modified copy. This yields the expected results
4. the modified *copy* is discarded
5. the original object (reservedCPUSet) is unchanged, we can safely Add() cores to it.

This probably means the API of cpuset.Builder could be improved, but I also wonder if we should rearchitect our function to better reflect the intended usage of `cpuset.Builder`, on which `Result()` should be the very last operation performed on them.

pkg/profilecreator/profilecreator.go
```diff
@@ -312,6 +307,34 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
 
 }
 
+// getCPUsSequentially returns Reserved and Isolated CPUs sequentially
+func (ghwHandler GHWHandler) getCPUsSequentially(reservedCPUCount int, topologyInfoNodes []*topology.Node) (string, string, error) {
+	totalCPUSet := cpuset.NewBuilder()
+	reservedCPUSet := cpuset.NewBuilder()
+	max := reservedCPUCount
```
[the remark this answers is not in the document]
swatisehgal: Sure! Will fix.

pkg/profilecreator/profilecreator.go
swatisehgal: Sure, can do that, no problem!

pkg/profilecreator/profilecreator.go
swatisehgal: Okay, if you think that maximizes clarity, we can certainly do that.

pkg/profilecreator/profilecreator.go
swatisehgal: Well, slightly different. We identify the right number of CPUs that need to be allocated per numa node, meaning reservedPerNuma + (the additional number based on the remainder and the NUMA node)
So best way to explain this is with the help of example here. If the user requests 15  reserved cpus and we have 4 numa nodes. We find reservedPerNuma which in this case is 3 and remainder = 3. So for each numa node we find a max which keeps track of the cumulative resources that should be allocated for each NUMA node:
max = (numaID+1)*reservedPerNuma + (numaNodeNum - remainder) 
For NUMA node 0 max = (0+1)*3 + 4-3 = 4 remainder is decremented => remainder is 2
For NUMA node 1 max = (1+1)*3 + 4-2 = 8 remainder is decremented => remainder is 1
For NUMA node 2 max = (2+1)*3 + 4-2 = 12 remainder is decremented => remainder is 0
For NUMA Node 3 remainder is =0 so 	max = 12 + 3 = 15.

Hope that helps explain things.

fromanirh: ok, please copypaste this comment as code comment and as illustrative example and let's move on

pkg/profilecreator/profilecreator.go
fromanirh: no pointing in changing the logic if we don't _also_ change the other helper (see comment below), so scratch this.

pkg/profilecreator/profilecreator.go
fromanirh: see below. Scratch this.

pkg/profilecreator/profilecreator.go
swatisehgal: Thanks for the great insight and the detailed explanation here. Given that cpuset.Builder is a mutable (and cpuset.CPUSet is immutable) we had to take the route of using cpuset.Builder. Also we have to check the size of the CPUset inside the Builder (for which we have to access Result()) to be able to allocate the correct amount of CPUs from a NUMA node before calling the Add(). Given that the current implementation works fine, I am going to leave it as is for now but will try to figure out how this can be implemented in a better way but I guess that will be done as a separate KEP when we have a better idea.

pkg/profilecreator/profilecreator.go
swatisehgal: Ack

pkg/profilecreator/profilecreator.go
swatisehgal: Ack

pkg/profilecreator/profilecreator.go
swatisehgal: Ack

fromanirh:
/approve
/lgtm

thanks for the followups.

PR #570, opened 2021-03-08
Validate user specified arguments and populate fields in the created Performance Profile

swatisehgal:
In this PR we:

- Refactor the code to move the code in RunE responsible for obtaining profile data from flags and creating the performance profile into private methods that are then called from RunE.
- Validate and populate Topology Manager Policy and  Real Time Kernel flag in performance profile as specified by the user.
- Populate Real Time Kernel flag in performance profile as specified by the user.
- Validate power-consumption-mode flag value provided by user.
- Ensure that reservedCPUCount is less than total number of CPUs.

### DIFF
--- a/cmd/performance-profile-creator/cmd/root.go
+++ b/cmd/performance-profile-creator/cmd/root.go
@@ -28,8 +28,6 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"k8s.io/utils/pointer"
-
 	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -49,6 +47,7 @@ type ProfileData struct {
 	nodeSelector               *metav1.LabelSelector
 	performanceProfileName     string
 	topologyPoilcy             string
+	rtKernel                   bool
 }
 
 // rootCmd represents the base command when called without any subcommands
@@ -110,7 +109,10 @@ func getDataFromFlags(cmd *cobra.Command) (*profileCreatorArgs, error) {
 		return nil, fmt.Errorf("Invalid value for power-consumption-mode flag specified: %v", err)
 	}
 	//TODO: Use the validated powerConsumptionMode above to be captured in the created performance profile
-
+	rtKernelEnabled, err := strconv.ParseBool(cmd.Flag("rt-kernel").Value.String())
+	if err != nil {
+		return nil, fmt.Errorf("Error parsing rt-kernel flag: %v", err)
+	}
 	flagData := &profileCreatorArgs{
 		mustGatherDirPath:           mustGatherDirPath,
 		profileName:                 profileName,
@@ -118,6 +120,7 @@ func getDataFromFlags(cmd *cobra.Command) (*profileCreatorArgs, error) {
 		splitReservedCPUsAcrossNUMA: splitReservedCPUsAcrossNUMA,
 		mcpName:                     mcpName,
 		tmPolicy:                    tmPolicy,
+		rtKernel:                    rtKernelEnabled,
 	}
 	return flagData, nil
 }
@@ -164,6 +167,7 @@ func getProfileData(args *profileCreatorArgs) (*ProfileData, error) {
 		nodeSelector:           mcp.Spec.NodeSelector,
 		performanceProfileName: args.profileName,
 		topologyPoilcy:         args.tmPolicy,
+		rtKernel:               args.rtKernel,
 	}
 	return profileData, nil
 }
@@ -232,7 +236,7 @@ func createProfile(profileData ProfileData) {
 			},
 			NodeSelector: profileData.nodeSelector.MatchLabels,
 			RealTimeKernel: &performancev2.RealTimeKernel{
-				Enabled: pointer.BoolPtr(true),
+				Enabled: &profileData.rtKernel,
 			},
 			AdditionalKernelArgs: []string{},
 			NUMA: &performancev2.NUMA{
--- a/cmd/performance-profile-creator/cmd/root.go
+++ b/cmd/performance-profile-creator/cmd/root.go
@@ -34,11 +34,20 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	singleNumaNodeTopologyManagerPolicy = "single-numa-node"
+)
+
+var (
+	validTMPolicyValues = []string{"single-numa-node", "best-effort", "restricted", "none", ""}
+)
+
 // ProfileData collects and stores all the data needed for profile creation
 type ProfileData struct {
 	isolatedCPUs, reservedCPUs string
 	nodeSelector               *metav1.LabelSelector
 	performanceProfileName     string
+	topologyPoilcy             string
 }
 
 // rootCmd represents the base command when called without any subcommands
@@ -80,12 +89,24 @@ func getDataFromFlags(cmd *cobra.Command) (*profileCreatorArgs, error) {
 		return nil, fmt.Errorf("Error parsing split-reserved-cpus-across-numa flag: %v", err)
 	}
 	profileName := cmd.Flag("profile-name").Value.String()
+	tmPolicy := cmd.Flag("topology-manager-policy").Value.String()
+	if err != nil {
+		return nil, fmt.Errorf("Error parsing topology-manager-policy flag: %v", err)
+	}
+	err = validateFlag(tmPolicy, validTMPolicyValues)
+	if err != nil {
+		return nil, fmt.Errorf("Invalid value for topology-manager-policy flag specified: %v", err)
+	}
+	if tmPolicy == singleNumaNodeTopologyManagerPolicy && splitReservedCPUsAcrossNUMA {
+		return nil, fmt.Errorf("Error as it is not appropriate to split reserved CPUs in case of topology-manager-policy: %v", tmPolicy)
+	}
 	flagData := &profileCreatorArgs{
 		mustGatherDirPath:           mustGatherDirPath,
 		profileName:                 profileName,
 		reservedCPUCount:            reservedCPUCount,
 		splitReservedCPUsAcrossNUMA: splitReservedCPUsAcrossNUMA,
 		mcpName:                     mcpName,
+		tmPolicy:                    tmPolicy,
 	}
 	return flagData, nil
 }
@@ -131,8 +152,21 @@ func getProfileData(args *profileCreatorArgs) (*ProfileData, error) {
 		isolatedCPUs:           isolatedCPUs,
 		nodeSelector:           mcp.Spec.NodeSelector,
 		performanceProfileName: args.profileName,
+		topologyPoilcy:         args.tmPolicy,
 	}
 	return profileData, nil
+}
+
+func validateFlag(value string, validValues []string) error {
+	for _, validValue := range validValues {
+		if strings.ToLower(value) == validValue {
+			// value is valid
+			return nil
+		}
+	}
+	// if we're here, value is invalid
+	return fmt.Errorf("Value '%s' is invalid for flag 'color'. Valid values "+
+		"come from the set %v", value, validValues)
 }
 
 type profileCreatorArgs struct {
@@ -145,6 +179,7 @@ type profileCreatorArgs struct {
 	rtKernel                    bool
 	userLevelNetworking         bool
 	mcpName                     string
+	tmPolicy                    string
 }
 
 func init() {
@@ -163,6 +198,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&args.mustGatherDirPath, "must-gather-dir-path", "M", "must-gather", "Must gather directory path")
 	rootCmd.MarkPersistentFlagRequired("must-gather-dir-path")
 	rootCmd.PersistentFlags().StringVarP(&args.profileName, "profile-name", "N", "performance", "Name of the performance profile to be created")
+	rootCmd.PersistentFlags().StringVarP(&args.tmPolicy, "topology-manager-policy", "Q", "restricted", "Kubelet Topology Manager Policy of the performance profile to be created. [Valid values: single-numa-node, best-effort, restricted, none]")
 
 	// TODO: Input validation
 	// 1) Make flags required/optional
@@ -193,7 +229,7 @@ func createProfile(profileData ProfileData) {
 			},
 			AdditionalKernelArgs: []string{},
 			NUMA: &performancev2.NUMA{
-				TopologyPolicy: pointer.StringPtr("restricted"),
+				TopologyPolicy: &profileData.topologyPoilcy,
 			},
 		},
 	}
--- a/cmd/performance-profile-creator/cmd/root.go
+++ b/cmd/performance-profile-creator/cmd/root.go
@@ -38,6 +38,7 @@ import (
 type ProfileData struct {
 	isolatedCPUs, reservedCPUs string
 	nodeSelector               *metav1.LabelSelector
+	performanceProfileName     string
 }
 
 // rootCmd represents the base command when called without any subcommands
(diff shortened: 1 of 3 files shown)


cmd/performance-profile-creator/cmd/root.go
```diff
@@ -34,11 +34,20 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	singleNumaNodeTopologyManagerPolicy = "single-numa-node"
+)
+
+var (
+	validTMPolicyValues = []string{"single-numa-node", "best-effort", "restricted", "none", ""}
```
marcel-apf: Why do we need 2 NOP options, "none" and  "" ?
Can we even have a none TM policy? I wasn't aware of it.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -80,12 +89,24 @@ func getDataFromFlags(cmd *cobra.Command) (*profileCreatorArgs, error) {
 		return nil, fmt.Errorf("Error parsing split-reserved-cpus-across-numa flag: %v", err)
 	}
 	profileName := cmd.Flag("profile-name").Value.String()
+	tmPolicy := cmd.Flag("topology-manager-policy").Value.String()
+	if err != nil {
+		return nil, fmt.Errorf("Error parsing topology-manager-policy flag: %v", err)
```
marcel-apf, line 94: We will need to go over all error messages and remove the capital letters and even the "Error" prefix, but is low priority and not part of this PR.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -39,7 +39,8 @@ const (
 )
 
 var (
-	validTMPolicyValues = []string{"single-numa-node", "best-effort", "restricted", "none", ""}
+	validTMPolicyValues        = []string{"single-numa-node", "best-effort", "restricted", "none", ""}
+	validPowerConsumptionModes = []string{"cstate", "no-state", "idle-poll"}
```
marcel-apf, line 41: @MarSik @fromanirh are the names clear enough? 
What we have in design is:
Power consumption mode
   Default (cstate/pstate sleep level limit to 1)
   Ignore power consumption (don’t use c-states to save power)
   No power saving  - (idle=pool))

Maybe we can have
   "limit-cstates" - default
   "ignore"
    "no-power-saving"

Can you come up with better names?

Also do you remember maybe what kernel args *exactly* we should use for it? We can delay this part for when we'll send a PR for it, but maybe you have the answer.

cmd/performance-profile-creator/cmd/root.go
swatisehgal: We don't need "" anymore. In the previous implementation, in case the user wasn't specifying any value it would default to "" but as we default to "restricted" that is no longer relevant.
Yes "none" is one the TM policies. Please refer [here](https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/#topology-manager-policies).

cmd/performance-profile-creator/cmd/root.go
swatisehgal: No problem, can fix this.

cmd/performance-profile-creator/cmd/root.go
fromanirh: My take: I'd abstract the power saving setting from the implementation, hence I'd go with something like

- default: maps to your default
- performance: maps to your "ignore power consumption" (is it actually ignoring?)
- low-latency: maps to "no power saving"

or something like it. The docs should also further clarify that the higher we go (low-latency is higher than performance) the more power the system will consume.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -38,67 +38,23 @@ import (
...
 
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
 	Use:   "performance-profile-creator",
 	Short: "A tool that automates creation of Performance Profiles",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		mcpName := cmd.Flag("mcp-name").Value.String()
-		mustGatherDirPath := cmd.Flag("must-gather-dir-path").Value.String()
-		mcp, err := profilecreator.GetMCP(mustGatherDirPath, mcpName)
+		flagData, err := getDataFromFlags(cmd)
```
fromanirh, line 58: I'd rename like `profileCreatorArgsFromFlags`
Additionally: we can return a value (not a pointer) here, can't we?

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -38,67 +38,23 @@ import (
...
-		handle, err := profilecreator.NewGHWHandler(mustGatherDirPath, matchedNodes[0])
-		reservedCPUs, isolatedCPUs, err := handle.GetReservedAndIsolatedCPUs(reservedCPUCount, splitReservedCPUsAcrossNUMA)
-		if err != nil {
-			return fmt.Errorf("Error obtaining Reserved and Isolated CPUs for %s: %v", nodeName, err)
-		}
-		profileData = ProfileData{
-			reservedCPUs: reservedCPUs,
-			isolatedCPUs: isolatedCPUs,
-			nodeSelector: mcp.Spec.NodeSelector,
+			return fmt.Errorf("Error obtaining profileData %v", err)
```
fromanirh: I'd use `error creating the profile: %v`

cmd/performance-profile-creator/cmd/root.go
fromanirh: can't we just import the k8s/ocp names? duplicating data (and validation) always carries some risk.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -131,10 +152,23 @@ func getProfileData(args *profileCreatorArgs) (*ProfileData, error) {
...
 		nodeSelector:           mcp.Spec.NodeSelector,
 		performanceProfileName: args.profileName,
+		topologyPoilcy:         args.tmPolicy,
 	}
 	return profileData, nil
 }
 
+func validateFlag(value string, validValues []string) error {
+	for _, validValue := range validValues {
+		if strings.ToLower(value) == validValue {
```
fromanirh, line 177: `strings.EqualFold` (https://golang.org/pkg/strings/#EqualFold) will probably come handy here

fromanirh: the string matching should be extracted:

```
func validateFlag(value string, validValues []string) error {
	if isStringInSlice(value, validValues) {
		return nil
       }
	return fmt.Errorf("Value '%s' is invalid for flag 'color'. Valid values "+
		"come from the set %v", value, validValues)
}

func isStringInSlice(value string, candidates []string) bool {
	for _, candidate := range candidates {
		if strings.EqualFold(candidate, value) {
			return true
		}
	}
	return false
}
```

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -163,6 +198,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&args.mustGatherDirPath, "must-gather-dir-path", "M", "must-gather", "Must gather directory path")
 	rootCmd.MarkPersistentFlagRequired("must-gather-dir-path")
 	rootCmd.PersistentFlags().StringVarP(&args.profileName, "profile-name", "N", "performance", "Name of the performance profile to be created")
+	rootCmd.PersistentFlags().StringVarP(&args.tmPolicy, "topology-manager-policy", "Q", "restricted", "Kubelet Topology Manager Policy of the performance profile to be created. [Valid values: single-numa-node, best-effort, restricted, none]")
```
fromanirh, line 216: why "Q"?

pkg/profilecreator/profilecreator.go
```diff
@@ -254,8 +254,12 @@ func (ghwHandler GHWHandler) SortedTopology() (*topology.Info, error) {
...
 // GetReservedAndIsolatedCPUs returns Reserved and Isolated CPUs
 func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, splitReservedCPUsAcrossNUMA bool) (string, string, error) {
-	if reservedCPUCount < 0 {
-		return "", "", fmt.Errorf("Specified eservered CPU count is negative, please specify it correctly")
+	cpuInfo, err := ghwHandler.CPU()
+	if err != nil {
+		return "", "", fmt.Errorf("Error obtaining CPU Info from GHW snapshot: %v", err)
+	}
+	if reservedCPUCount < 0 || reservedCPUCount > int(cpuInfo.TotalThreads) {
+		return "", "", fmt.Errorf("Specified reserved CPU count is invalid, please specify it correctly")
```
fromanirh: this is very nicely worded but not helpful. Let's tell the user what PPC thinks is correct:
```
fmt.Errorf("Specified reserved CPU count is invalid, please specify it in the range [0,%d]", cpuInfo.TotalThreads)

File: cmd/performance-profile-creator/cmd/root.go (hunk 7 above)
Comment (swatisehgal) [reply cmd/performance-profile-creator/cmd/root.go:216]: Ran out of letter :) We used up T, M and P so thought Q for Kubelet (Qbelet) Topology Manager Policy purely for the phonetic reasons

File: cmd/performance-profile-creator/cmd/root.go (hunk 6 above)
Comment (swatisehgal) [reply cmd/performance-profile-creator/cmd/root.go:177]: Thanks, will fix

File: pkg/profilecreator/profilecreator.go (hunk 8 above)
Comment (swatisehgal) [reply pkg/profilecreator/profilecreator.go]: +1, will do!

File: cmd/performance-profile-creator/cmd/root.go [hunk 9]
```diff
@@ -30,15 +30,16 @@ import (
...
 )
 
 var (
-	validTMPolicyValues        = []string{"single-numa-node", "best-effort", "restricted", "none"}
-	validPowerConsumptionModes = []string{"cstate", "no-state", "idle-poll"}
+	validTMPolicyValues = []string{kubeletconfig.SingleNumaNodeTopologyManager, kubeletconfig.BestEffortTopologyManagerPolicy, kubeletconfig.RestrictedTopologyManagerPolicy}
+	// Values part of validPowerConsumptionModes are explained below:
+	// default => Use CPU C-states but limit them to 1, meaning the processor is idle it can sleep but not "deep sleep"
+	// performance => Disable CPU sleep (c-states), processor never sleeps even if is idle
+	// low-latency => processor is never idle, it is in polling mode (cpu=poll)
```
marcel-apf, line 41: performance + idle=poll (not cpu=poll, my mistake)

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -30,15 +30,16 @@ import (
...
+	kubeletconfig "k8s.io/kubelet/config/v1beta1"
 )
 
 var (
-	validTMPolicyValues        = []string{"single-numa-node", "best-effort", "restricted", "none"}
-	validPowerConsumptionModes = []string{"cstate", "no-state", "idle-poll"}
+	validTMPolicyValues = []string{kubeletconfig.SingleNumaNodeTopologyManager, kubeletconfig.BestEffortTopologyManagerPolicy, kubeletconfig.RestrictedTopologyManagerPolicy}
+	// Values part of validPowerConsumptionModes are explained below:
+	// default => Use CPU C-states but limit them to 1, meaning the processor is idle it can sleep but not "deep sleep"
+	// performance => Disable CPU sleep (c-states), processor never sleeps even if is idle
```
marcel-apf, line 40: intel_idle.max_cstate=0 processor.max_cstate=0 intel_pstate=disable

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -30,15 +30,16 @@ import (
...
-	singleNumaNodeTopologyManagerPolicy = "single-numa-node"
+	kubeletconfig "k8s.io/kubelet/config/v1beta1"
 )
 
 var (
-	validTMPolicyValues        = []string{"single-numa-node", "best-effort", "restricted", "none"}
-	validPowerConsumptionModes = []string{"cstate", "no-state", "idle-poll"}
+	validTMPolicyValues = []string{kubeletconfig.SingleNumaNodeTopologyManager, kubeletconfig.BestEffortTopologyManagerPolicy, kubeletconfig.RestrictedTopologyManagerPolicy}
+	// Values part of validPowerConsumptionModes are explained below:
+	// default => Use CPU C-states but limit them to 1, meaning the processor is idle it can sleep but not "deep sleep"
```
marcel-apf, line 39: "processor.max_cstate=1"

cmd/performance-profile-creator/cmd/root.go
marcel-apf: @MarSik can you please confirm the kernel args for each power mode? Thank you

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -203,17 +210,17 @@ func init() {
 	rootCmd.PersistentFlags().IntVarP(&args.reservedCPUCount, "reserved-cpu-count", "R", 0, "Number of reserved CPUs (required)")
 	rootCmd.MarkPersistentFlagRequired("reserved-cpu-count")
 	rootCmd.PersistentFlags().BoolVarP(&args.splitReservedCPUsAcrossNUMA, "split-reserved-cpus-across-numa", "S", false, "Split the Reserved CPUs across NUMA nodes")
-	rootCmd.PersistentFlags().StringVarP(&args.mcpName, "mcp-name", "T", "worker-cnf", "MCP name corresponding to the target machines (required)")
+	rootCmd.PersistentFlags().StringVarP(&args.mcpName, "mcp-name", "n", "worker-cnf", "MCP name corresponding to the target machines (required)")
```
marcel-apf, line 213: Why not capital letter? it may confuse

pkg/profilecreator/profilecreator.go
```diff
@@ -259,7 +259,7 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
 		return "", "", fmt.Errorf("can't obtain CPU info from GHW snapshot: %v", err)
 	}
 	if reservedCPUCount < 0 || reservedCPUCount > int(cpuInfo.TotalThreads) {
```
marcel-apf, line 261: is  reservedCPUCount == int(cpuInfo.TotalThreads)  a valid choice? Not a single cpu for workloads? We may want to error out or to issue a warning at least

cmd/performance-profile-creator/cmd/root.go
MarSik: I wonder what is the difference between performance and low-latency. It is basically the same just using a different method, isn't it?

cmd/performance-profile-creator/cmd/root.go
MarSik: - "processor.max_cstate=1" <- This does not support 0 internally, it converts it to 1 anyway
    - "intel_idle.max_cstate=0"

cmd/performance-profile-creator/cmd/root.go
swatisehgal: I made updates based on @fromanirh suggestion [here](https://github.com/openshift-kni/performance-addon-operators/pull/570#discussion_r590338740).

In the previous patch, we used shorthand for topology-manager-policy flag as "Q" (discussion about this [here](https://github.com/openshift-kni/performance-addon-operators/pull/570#discussion_r590322813)) and updated shorthand for mcp-name to "n" (n for name as "N" was already used for policy-name) so I could use "T" for topology-manager-policy.

Please let me know if that is okay? If not please let me know what shorthand letter you would like me to use for topology-manager-policy.

marcel-apf: I am "lost" a little in letters. I have no special "preference" as long as all are Capital letters (or not)

fromanirh: the best option is just to get rid of single-letter options.

swatisehgal: Sure, can certainly do that if it causes less confusion :)

fromanirh (commented):
/lgtm
/hold
thanks for following through. All my comments are addressed so I'm happy with the state of the current PR.
Please remove the hold when the other reviewers are happy.

marcel-apf:
Looks good, thank you!
/unhold

PR #572, opened 2021-03-09
Net API including setting netqueues count on tuned

yanirq:
PAO introduces a new net object that can hold a object list containing 2 new fields:
**userLevelNetworking** - boolean flag , if true: Set all supported network devices queue count to the reserved CPUs count.   Default is false.  
**devices** - An optional list of devices structure as followed:
  name:
  vendor id:
  model id:

device fields have a syntax of shell-style wildcards which are either positive or negative. 
The negative rules are prefixed with an exclamation mark - e.g. exclude devices `!<device>` (use all devices other than the excluded list) 

**The queues for devices in this list will be set to the reserved CPUs count.**
**When no devices are added - all physical network devices on the node (where the performance profile applies to) will be set to the  reserved CPUs count.**

These values will be propagated to the generated base tuned profile creating for each entry under net an instance in the tuned configuration profile.


api/v2/performanceprofile_validation_test.go
```diff
@@ -28,12 +28,23 @@ const (
...
+	//NetDeviceName defines a net device name for the test profile
+	NetDeviceName = "enp0s4"
+	//NetDeviceVendorID defines a net device vendor ID for the test profile
+	NetDeviceVendorID = "0x1af4"
+	//NetDeviceModelID defines a net device model ID for the test profile
+	NetDeviceModelID = "0x1000"
+	//NetDeviceMAC defines a net device PCI path for the test profile
+	NetDevicePCIpath = "pci-0000:00:04.0"
+	//NetDeviceMAC defines a net device MAC address for the test profile
+	NetDeviceMAC = "enx54ee75491111"
```
fromanirh, line 47: shouldn't MAC look like `52:54:00:0b:b7:fb`?

api/v2/performanceprofile_validation_test.go
```diff
@@ -28,12 +28,23 @@ const (
...
 	MachineConfigPoolLabelValue = "mcpValue"
+
+	//NetDeviceName defines a net device name for the test profile
+	NetDeviceName = "enp0s4"
+	//NetDeviceVendorID defines a net device vendor ID for the test profile
+	NetDeviceVendorID = "0x1af4"
+	//NetDeviceModelID defines a net device model ID for the test profile
+	NetDeviceModelID = "0x1000"
+	//NetDeviceMAC defines a net device PCI path for the test profile
+	NetDevicePCIpath = "pci-0000:00:04.0"
```
fromanirh, line 45: This seems like a plain PCI address in DOMAIN-BUS-DEVICE-FUNCTIOn format. Is it so? If so why do we call it path?

api/v2/performanceprofile_types.go
```diff
@@ -123,6 +126,36 @@ type NUMA struct {
...
+
+// Device defines a way to represent a network device in several options:
+// device name, vendor ID, model ID, PCI path and MAC address
+type Device struct {
+	// Network device name
+	// +optional
+	Name *string `json:"name,omitempty"`
+	// Network device vendor ID represnted as a 16 bit Hexmadecimal number.
+	// +optional
+	VendorID *string `json:"vendor,omitempty"`
```
fromanirh, line 147: On which cases does make it sense to specify only the vendor...

api/v2/performanceprofile_types.go
```diff
@@ -123,6 +126,36 @@ type NUMA struct {
...
+	// If no devices are specified then the default is all devices.
+	Devices []Device `json:"devices,omitempty"`
+}
+
+// Device defines a way to represent a network device in several options:
+// device name, vendor ID, model ID, PCI path and MAC address
+type Device struct {
+	// Network device name
+	// +optional
+	Name *string `json:"name,omitempty"`
```
fromanirh, line 144: Should this be `Interface` or `InterfaceName`?

fromanirh (commented):
initial review

api/v2/performanceprofile_types.go
yanirq: It would probably make sense to change it to InterfaceName for clarity ,yes.

api/v2/performanceprofile_types.go
yanirq: In general all fields under device are acceptable fields by tuned net plugin - there will always be a cost for validation vs generalization.
If you specify only vendor you can cover a case where you want all devices of a specific vendor to be set with the downsized queue count for example.

api/v2/performanceprofile_types.go
```diff
@@ -123,6 +126,36 @@ type NUMA struct {
...
+	// Network device vendor ID represnted as a 16 bit Hexmadecimal number.
+	// +optional
+	VendorID *string `json:"vendor,omitempty"`
+	// Network device model ID (device) represnted as a 16 bit hexmadecimal number.
+	// +optional
+	ModelID *string `json:"model,omitempty"`
```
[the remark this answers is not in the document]
yanirq: Here its a bit more tricky. mentioning only model probably does not make much sense without specifying vendor ID but this is still acceptable in terms of tuned network plugin

api/v2/performanceprofile_types.go
fromanirh: ok, makes sense

api/v2/performanceprofile_types.go
```diff
@@ -123,6 +126,36 @@ type NUMA struct {
...
+	// Network device PCI path
+	// +optional
+	PCIpath *string `json:"pciPath,omitempty"`
```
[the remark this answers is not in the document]
fromanirh: got below, it looks like `pci-0000:00:04.0` which is weird. Can't parse what a "pci path" is :\

api/v2/performanceprofile_validation_test.go
yanirq: We  can match anything udev knows about the device - e.g. udevadm info -p <device path>
The format presented here `ID_NET_NAME_MAC=prefixxAABBCCDDEEFF` matches the way udev maps this MAC address.
What I will probably add will be matching it as a regular expression which means we can e.g. use:
ID_NET_NAME_MAC=.*xAABBCCDDEEFF

I will probably add a validation to accept valid MAC addresses and add an internal functions for conversions from the EUI-48/64  for more simple usage in the tuned base profile we have, e.g.:
[net]
devices_udev_regex=${f:net_eui64_to_udev_regex:${EUI_64_ADDRESS}}
(OR do the translation on our side but the functions already exit in tuned side)

fromanirh: btw, as you also wrote, users will likely expect the simplest format I mentioned, so maybe let's start with this?

api/v2/performanceprofile_validation_test.go
yanirq: following the udev convention here, for example:
udevadm info -p /sys/class/net/ens4
P: /devices/pci0000:00/0000:00:04.0/virtio1/net/ens4
...
E: ID_PATH=pci-0000:00:04.0

yanirq: But maybe I should remove PCI and just call it interfacePath

fromanirh: But isn't this "path" in the form "pci-$PCI_ADDRESS" where PCI_ADDRESS="0000:00:04.0" (DOMAIN:BUS:DEVICE.FUNCTION) ?
meaning: shouldn't it be `NetDevicePCIAddress` ?

api/v2/performanceprofile_types.go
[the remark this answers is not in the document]
fromanirh: fine. Point is: shall we pass through the network plugin? (IOW: should we expose all the options the network plugin allows _in the same format_ the network plugin allows?) This is something we need to carefully evaluate because adding options to the API seems safe and easy, while removing them is painful.

For example: can we start allowing a `vendorid:modelid` pair supporting a simple wildcard mechanism to specfy only `vendorid` ?

api/v2/performanceprofile_validation_test.go
MarSik: Yeah, using the basic format is probably the correct way to do this. Most of the tools show it like that (ip a, ifconfig..).

api/v2/performanceprofile_validation_test.go
MarSik: @fromanirh I wonder if this could be usb path as well (or some other bus). Probably not on a server, but..

fromanirh: fair point, but we have `PCI` hardcoded in the name of the field already (`NetDevicePCIpath`).
My biggest gripe is that "PCI path" is a quite hard concept to parse and seems very udev-specific. We should not expose udev-specific details in our API, especially considering that a plain simple PCI address seems to be equivalent and much simpler concept.

api/v2/performanceprofile_types.go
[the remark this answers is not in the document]
fromanirh: it could look like:

- `vendorid:modelid` - match this exact pair
- `vendorid` - match all models from this vendor

IOW:

1. if `:` is in the string, treat like a pair
2. otherwise, treat it like a `vendorid`

api/v2/performanceprofile_types.go
marcel-apf: I personally don't like this approach, users will want to use the same devices in all the nodes, otherwise they may not have the same hw, for example the same SRIOV VFs count. The way I understand the feature is that we need ways to identify the same Net devices. Is it customary use *the pair* VendorId:DeviceId to identify PCI devices.
We can use PPC to verify the assumptions, but it seems risky. I am OK with whatever you decide.

api/v2/performanceprofile_types.go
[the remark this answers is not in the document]
marcel-apf: Please see my comment above, anyway please use deviceid as is the standard naming.

api/v2/performanceprofile_types.go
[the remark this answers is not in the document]
marcel-apf: Please use the customary [domain]:[bus]:[device]:[function] approach. For ARI you have [domain]:[bus]:[function]

api/v2/performanceprofile_types.go
```diff
@@ -123,6 +126,36 @@ type NUMA struct {
...
+	// Network device MAC address
+	// +optional
+	MAC *string `json:"mac,omitempty"`
```
marcel-apf, line 156: Please ensure syntactic validation for all the fields. I would use the standard "ff":"ff": notation.
I do not understand how we can use MAC addresses to identify similar devices on different nodes? Are we going to use some "pattern" based matching? Or is this a way to enumerate all devices on all nodes if we don;t have other ways to identify them?
Other points:
 - There are net devices that have multiple MAC addresses
 - MAC addresses can be changed by drivers / firmware updates and so on AFAIK

api/v2/performanceprofile_validation_test.go
marcel-apf: Please see my comment above, just use domain:bus:device:function, please don;t use something that is not universally accepted. The same reservations about using PCI addresses

yanirq: Given all the above - since we will probably need a way to translate the PCI address to something udev understands and the risk of PCI addresses not being guaranteed I wonder how badly to we want this field.
@MarSik  thoughts here ?

api/v2/performanceprofile_types.go
```diff
@@ -123,6 +126,36 @@ type NUMA struct {
 	TopologyPolicy *string `json:"topologyPolicy,omitempty"`
 }
 
+// Net defines a set of network related features
+type Net struct {
+	// UserLevelNetworking defines if the specified network devices . Defaults to "false".
+	UserLevelNetworking *bool `json:"userLevelNetworking,omitempty"`
```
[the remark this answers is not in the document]
yanirq: yes, and that would be the default case.
If not devices are listed here and the flag is true - every non virtual device in the node will be set with a netqueue count = reserved cpus count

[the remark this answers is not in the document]
yanirq: We can use yet another flag and not the one in PPC if that poses a risk

api/v2/performanceprofile_validation_test.go
fromanirh: @marcel-apf could you please link docs (kernel or whatever) that explain this behaviour?

api/v2/performanceprofile_types.go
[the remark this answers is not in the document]
marcel-apf: that's actually ok, when the user turns on the UserLevelNetworking  flag in PPC, we are going to set it in the PAO profile

api/v2/performanceprofile_validation.go
```diff
@@ -253,3 +255,39 @@ func (r *PerformanceProfile) validateNUMA() field.ErrorList {
...
 }
+
+func (r *PerformanceProfile) validateNet() field.ErrorList {
+	var allErrs field.ErrorList
+
+	if r.Spec.Net == nil {
+		return allErrs
+	}
+
+	if *r.Spec.Net.UserLevelNetworking && r.Spec.CPU.Reserved == nil {
```
marcel-apf: CPU field may be NULL

api/v2/performanceprofile_types.go
marcel-apf, line 151: DeviceId please

api/v2/performanceprofile_validation.go
```diff
@@ -253,3 +255,39 @@ func (r *PerformanceProfile) validateNUMA() field.ErrorList {
...
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec.net"), r.Spec.Net, "Can not set network devices queues count without specifiying spec.cpu.reserved"))
```
marcel-apf, line 267: as I understand errors should not start with capital letters

api/v2/performanceprofile_validation.go
```diff
@@ -253,3 +255,39 @@ func (r *PerformanceProfile) validateNUMA() field.ErrorList {
...
+	}
+
+	for _, device := range r.Spec.Net.Devices {
```
marcel-apf, line 270: Net can be NULL?

docs/performance_profile.md
```diff
@@ -39,6 +41,20 @@ CPUSet is of type `string`.
...
 
+## Device
+
+Device defines a way to represent a network device in several options: device name, vendor ID, model ID, PCI path and MAC address
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| name | Network device name | *string | false |
+| vendor | Network device vendor ID represnted as a 16 bit Hexmadecimal number. | *string | false |
+| model | Network device model ID (device) represnted as a 16 bit hexmadecimal number. | *string | false |
```
marcel-apf, line 52: device

docs/performance_profile.md
```diff
@@ -80,6 +96,17 @@ NUMA defines parameters related to topology awareness and affinity.
 
 [Back to TOC](#table-of-contents)
 
+## Net
+
+Net defines a set of network related features
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| userLevelNetworking | UserLevelNetworking defines if the specified network devices . Defaults to \"false\". | *bool | false |
```
marcel-apf: I couldn't parse the above

docs/performance_profile.md
```diff
@@ -80,6 +96,17 @@ NUMA defines parameters related to topology awareness and affinity.
...
+| devices | Devices contains a list of network device representations that will be set with a netqueue count equals to CPU.Reserved . If no devices are specified then the default is all devices. | [][Device](#device) | false |
```
marcel-apf, line 104: /s/equals/equal . Also you may want to use link to CPU reserved (optional!)

functests/1_performance/performance.go
```diff
@@ -328,6 +329,58 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 		})
 	})
 
+	Context("Network device queues adjusted by Tuned", func() {
+		//TODO - set test id
+		It("[test_id:XXXXX][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Should be set to the profile's reserverd CPUs count ", func() {
```
marcel-apf, line 334: /s/reserverd/reserved

functests/1_performance/performance.go
```diff
@@ -328,6 +329,58 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+			noDeviceFound := true
+			if profile.Spec.Net != nil {
+				reservedSet, err := cpuset.Parse(string(*profile.Spec.CPU.Reserved))
+				Expect(err).ToNot(HaveOccurred())
+				reserveCPUsCount := reservedSet.Size()
+				if *profile.Spec.Net.UserLevelNetworking && len(profile.Spec.Net.Devices) == 0 {
```
marcel-apf: Net can be NULL?

functests/1_performance/performance.go
```diff
@@ -339,6 +339,7 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				reserveCPUsCount := reservedSet.Size()
 				if profile.Spec.Net.UserLevelNetworking != nil && *profile.Spec.Net.UserLevelNetworking && len(profile.Spec.Net.Devices) == 0 {
 					By("To all non virtual network devices when no devices are specified under profile.Spec.Net.Devices")
+					validatTunedActiveProfile(workerRTNodes)
```
marcel-apf, line 342: /s/validatTunedActiveProfile/validateTunedActiveProfile

api/v2/performanceprofile_validation.go
yanirq: @marcel-apf  if its NULL then we can't set a desired number  to the count no ?  what should be the default behaviour in that case ?  should we just do nothing  ?  maybe issue a warning ?

api/v2/performanceprofile_validation.go
yanirq: there is a check about that returns if Net == NULL

functests/1_performance/performance.go
yanirq: there is a check few line above

functests/1_performance/performance.go
```diff
@@ -323,11 +324,63 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
 			Expect(err).ToNot(HaveOccurred(), "cannot find the Cluster Node Tuning Operator object "+components.ProfileNamePerformance)
-			validatTunedActiveProfile(workerRTNodes)
+			validateTunedActiveProfile(workerRTNodes)
 			execSysctlOnWorkers(workerRTNodes, sysctlMap)
 		})
 	})
 
+	Context("Network device queues adjusted by Tuned", func() {
+		//TODO - set test id
+		It("[test_id:XXXXX][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Should be set to the profile's reserved CPUs count ", func() {
```
yanirq, line 334: @gsr-shanks @Avielyo10   What test ID should I use here ?

api/v2/performanceprofile_validation_test.go
```diff
@@ -69,6 +86,18 @@ func NewPerformanceProfile(name string) *PerformanceProfile {
...
 			},
+			Net: &Net{
+				UserLevelNetworking: pointer.BoolPtr(true),
+				Devices: []Device{
+					{
+						Name:     &netDeviceName,
+						VendorID: &netDeviceVendorID,
+						ModelID:  &netDeviceModelID,
+						PCIpath:  &netDevicePCIpath,
+						MAC:      &netDeviceMAC,
```
fromanirh: what about if these fields refer to different cards? should we error out?

fromanirh (commented):
/lgtm
/hold
All my comments have been addressed, so LGTM, but holding to let other reviewers review.
I'd add a test (or few tests) to document the behaviour when the user set conflicting parameters to identify a device. What should happen in that case?

api/v2/performanceprofile_validation_test.go
marcel-apf: > /lgtm
> /hold
> All my comments have been addressed, so LGTM, but holding to let other reviewers review.
> I'd add a test (or few tests) to document the behaviour when the user set conflicting parameters to identify a device. What should happen in that case?

I think PPC can check for errors, but PAO can't know if vendorid:deviceid makes sense.
BTW, are we allowed to give both vendor/device AND interfaceName for the the same net device?
Is there any other kind of conflict?

fromanirh: yep, close to my point. Do we want to make the ways to identify a device mutually exclusive? do we care at all about possible inconsistencies?

marcel-apf: I like the "more validation" approach, I would make it mutually exclusive.
Any inconsistencies would result on the tuned profile not being applied partially; we will we end up with a cryptic error event  in the "best-case", silent failure otherwise.

yanirq: Please note that you are commenting on outdate code,  I already removed MAC and PCI PATH.
> BTW, are we allowed to give both vendor/device AND interfaceName for the the same net device?
All the relevant validations are here:
https://github.com/openshift-kni/performance-addon-operators/blob/edac4f260867db0f8edb26ac75f9805b6d757771/api/v2/performanceprofile_validation.go#L270 

Let me know if there is anything else missing there.

fromanirh: yes, I seen. The open question still stands: do we want to allow users to specify ALL the parameters or should the ways to identify devices be mutually exclusive? what if interfacename and vendor:device are inconsistent? we just error out?
The answer can totally be "yes, we just error out", but worth discussing anyway.

yanirq: We cant error out on PAO in that case since we cant check in this stage if under the same device object  if interfacename and vendor:device are consistent.  
We can either deny mutual existence of both on PAO  or verify that on PPC side,  not sure what is the best option atm

fromanirh: ok, so I guess we keep things as they are atm.

marcel-apf: > We cant error out on PAO in that case since we cant check in this stage if under the same device object if interfacename and vendor:device are consistent.
> We can either deny mutual existence of both on PAO or verify that on PPC side, not sure what is the best option at the moment.

I vote for denying the use of InterfaceName AND (vendorid-deviceid) pair for the same "netdevice". If you want both, declare 2 netdevices.
PPC will check the semantics of each (not together)

MarSik: Remember the netdevice can be a wildcard. In general my personal experience is to avoid overconstrained systems and checks, it eventually blocks someone from doing something they need.

marcel-apf: What will be the meaning of NetDevice {interfaceName=ens*,deviceid=0x1000} ? All 0x1000 devices mapped to ens* interfaces? Does tuned support that? Or it will be translates into  0x1000 devices + ens*  interfaces?

yanirq: it will be translated to 0x1000 devices + ens* interfaces

marcel-apf: So on our system it will look like an "AND" operation, but would be translated into an "OR". Does it look friendly to anyone?

api/v2/performanceprofile_validation_test.go
```diff
@@ -292,11 +284,34 @@ var _ = Describe("PerformanceProfile", func() {
...
+				invalidVendor := "123"
+				invalidModel := "0x12345"
+				profile.Spec.Net.Devices[0].InterfaceName = pointer.StringPtr("")
+				profile.Spec.Net.Devices[0].VendorID = pointer.StringPtr(invalidVendor)
+				profile.Spec.Net.Devices[0].ModelID = pointer.StringPtr(invalidModel)
+				errors := profile.validateNet()
+				Expect(errors).NotTo(BeEmpty())
+				Expect(errors[0].Error()).To(ContainSubstring(fmt.Sprintf("device name cannot be empty")))
+				Expect(errors[0].Error()).To(ContainSubstring(fmt.Sprintf("device vendor ID %s has an invalid format. Vendor ID should be represented as 0x<4 hexadecimal digits> (16 bit representation)", invalidVendor)))
+				Expect(errors[0].Error()).To(ContainSubstring(fmt.Sprintf("device model ID %s has an invalid format. Model ID should be represented as 0x<4 hexadecimal digits> (16 bit representation)", invalidModel)))
```
fromanirh: should not be checking *at least one* between InterfaceName vs VendorID+ModelID should be given? (this should be an additional test probably)

fromanirh (commented):
/lgtm
/hold
possible followup inside, please check. We're good to move forward.

fromanirh:
/hold cancel
we have all the acks we need

PR #573, opened 2021-03-09
Bug 1936953: remove the psr check from the rcu_nocbs test

cynepco3hahue:
We have flakiness during psr check, because:
1. OS creates rcuo process
2. rcuo process runs on some CPU
3. OS set the affinity for all rcuo processes
4. we are checking rcuo psr, but it did not run again so we will see psr before affinity change

It should be enough to check only the CPU affinity, because the assumption that the process can
not run on the CPU that the process does not pinned to it.

### DIFF
--- a/functests/1_performance/cpu_management.go
+++ b/functests/1_performance/cpu_management.go
@@ -38,8 +38,8 @@ var profile *performancev2.PerformanceProfile
 var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 	var balanceIsolated bool
 	var reservedCPU, isolatedCPU string
-	var listReservedCPU, listIsolatedCPU []int
-	var reservedCPUSet, isolatedCPUSet cpuset.CPUSet
+	var listReservedCPU []int
+	var reservedCPUSet cpuset.CPUSet
 
 	BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


deploy/olm-catalog/performance-addon-operator/4.8.0/performance-addon-operator.v4.8.0.clusterserviceversion.yaml
```diff
@@ -304,23 +304,3 @@ spec:
     targetPort: 4343
     type: ConversionWebhook
     webhookPath: /convert
-  - admissionReviewVersions:
```
fromanirh, line 307: why was this removed? seems unexpected

fromanirh (commented):
the change per se lgtm but it seems we have an unwanted CSV change?

deploy/olm-catalog/performance-addon-operator/4.8.0/performance-addon-operator.v4.8.0.clusterserviceversion.yaml
fromanirh: ack, once you do that, consider this PR as LGTM by me.

PR #575, opened 2021-03-11
Populate the additionalKernelArgs field in the performance profile

swatisehgal:
- based on the power mode specified by the user we populate the kernel arguments
  in the additionalKernelArgs field of the created performance profile.
- Log  values of computed fields to explain how each field in the profile is populated.
- Minor refactoring to print logs corresponding to reservedCPUs
   and isolatedCPUs just once after GetReservedAndIsolatedCPUs is
   called.
- Additional unit tests for GetAdditionalKernelArgs
- Updates to the README

### DIFF
--- a/cmd/performance-profile-creator/cmd/root.go
+++ b/cmd/performance-profile-creator/cmd/root.go
@@ -35,11 +35,6 @@ import (
 
 var (
 	validTMPolicyValues = []string{kubeletconfig.SingleNumaNodeTopologyManager, kubeletconfig.BestEffortTopologyManagerPolicy, kubeletconfig.RestrictedTopologyManagerPolicy}
-	// default => no args
-	// low-latency => "nmi_watchdog=0", "audit=0",  "mce=off"
-	// ultra-low-latency: low-latency values + "processor.max_cstate=1", "intel_idle.max_cstate=0", "idle=poll"
-	validPowerConsumptionModes = []string{"default", "low-latency", "ultra-low-latency"}
-	additionalKernelArgs       = []string{"nmi_watchdog=0", "audit=0", "mce=off", "processor.max_cstate=1", "intel_idle.max_cstate=0", "idle=poll"}
 )
 
 // ProfileData collects and stores all the data needed for profile creation
@@ -83,17 +78,14 @@ func getDataFromFlags(cmd *cobra.Command) (profileCreatorArgs, error) {
 	creatorArgs := profileCreatorArgs{}
 	mustGatherDirPath := cmd.Flag("must-gather-dir-path").Value.String()
 	mcpName := cmd.Flag("mcp-name").Value.String()
-	log.Infof("mcp-name: %s ", mcpName)
 	reservedCPUCount, err := strconv.Atoi(cmd.Flag("reserved-cpu-count").Value.String())
 	if err != nil {
 		return creatorArgs, fmt.Errorf("failed to parse reserved-cpu-count flag: %v", err)
 	}
-	log.Infof("reserved-cpu-count: %d ", reservedCPUCount)
 	splitReservedCPUsAcrossNUMA, err := strconv.ParseBool(cmd.Flag("split-reserved-cpus-across-numa").Value.String())
 	if err != nil {
 		return creatorArgs, fmt.Errorf("failed to parse split-reserved-cpus-across-numa flag: %v", err)
 	}
-	log.Infof("split-reserved-cpus-across-numa = %t", splitReservedCPUsAcrossNUMA)
 	profileName := cmd.Flag("profile-name").Value.String()
 	tmPolicy := cmd.Flag("topology-manager-policy").Value.String()
 	if err != nil {
@@ -106,16 +98,15 @@ func getDataFromFlags(cmd *cobra.Command) (profileCreatorArgs, error) {
 	if tmPolicy == kubeletconfig.SingleNumaNodeTopologyManager && splitReservedCPUsAcrossNUMA {
 		return creatorArgs, fmt.Errorf("not appropriate to split reserved CPUs in case of topology-manager-policy: %v", tmPolicy)
 	}
-	log.Infof("topology-manager-policy: %s ", tmPolicy)
 	powerConsumptionMode := cmd.Flag("power-consumption-mode").Value.String()
 	if err != nil {
 		return creatorArgs, fmt.Errorf("failed to parse power-consumption-mode flag: %v", err)
 	}
-	err = validateFlag(powerConsumptionMode, validPowerConsumptionModes)
+	err = validateFlag(powerConsumptionMode, profilecreator.ValidPowerConsumptionModes)
 	if err != nil {
 		return creatorArgs, fmt.Errorf("invalid value for power-consumption-mode flag specified: %v", err)
 	}
-	log.Infof("power-consumption-mode: %s ", powerConsumptionMode)
+
 	//TODO: Use the validated powerConsumptionMode above to be captured in the created performance profile
 	rtKernelEnabled, err := strconv.ParseBool(cmd.Flag("rt-kernel").Value.String())
 	if err != nil {
@@ -154,7 +145,11 @@ func getProfileData(args profileCreatorArgs) (*ProfileData, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to find matching nodes for %s: %v", args.mcpName, err)
 	}
-
+	var matchedNodeNames []string
+	for _, node := range matchedNodes {
+		matchedNodeNames = append(matchedNodeNames, node.GetName())
+	}
+	log.Infof("Nodes targetted by %s MCP are: %v", args.mcpName, matchedNodeNames)
 	err = profilecreator.EnsureNodesHaveTheSameHardware(args.mustGatherDirPath, matchedNodes)
 	if err != nil {
 		return nil, fmt.Errorf("targeted nodes differ: %v", err)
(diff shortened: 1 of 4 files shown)


cmd/performance-profile-creator/cmd/root.go
```diff
@@ -83,14 +83,17 @@ func getDataFromFlags(cmd *cobra.Command) (profileCreatorArgs, error) {
 	creatorArgs := profileCreatorArgs{}
 	mustGatherDirPath := cmd.Flag("must-gather-dir-path").Value.String()
 	mcpName := cmd.Flag("mcp-name").Value.String()
+	log.Infof("mcp-name: %s ", mcpName)
```
fromanirh, line 86: could you please add in the commit message why/how this change improve the log messages?
Also, should these logs be Info or should they be debug? Or do we want to emit them only if a certain verbosity level is given?

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -160,17 +165,33 @@ func getProfileData(args profileCreatorArgs) (*ProfileData, error) {
...
 
+func getAdditionalKernelArgs(powerMode string) []string {
+	var kernelArgs []string
+	switch powerMode {
+	case validPowerConsumptionModes[0]:
+		kernelArgs = []string{}
+	case validPowerConsumptionModes[1]:
+		kernelArgs = additionalKernelArgs[:3]
+	case validPowerConsumptionModes[2]:
+		kernelArgs = additionalKernelArgs
```
fromanirh, line 194: not a fan of this approach. I prefer to have different sets and to merge them as needed. Also, the kernel args (among other things) are data we treat as code, so it should be inside some pkg, albeit possibly a tiny one.

cmd/performance-profile-creator/cmd/root.go
marcel-apf: We want to send to stderr as INFO the exact reason we created the profile so the created profile would not have "magic" results. We intend to justify each field in the profile we create. E.g. the node selector comes from the mcp node selector.

fromanirh: Sure, and so verbose logging is always warranted? Shouldn't the user opt in?

fromanirh: Additionally: I see some merit in (optionally) logging the compute field. But what do we gain in echoing back the value given as options?

pkg/profilecreator/profilecreator.go
```diff
@@ -338,10 +340,11 @@ func (ghwHandler GHWHandler) getCPUsSequentially(reservedCPUCount int, htEnabled
 	}
 	totalCPUSet := totalCPUSetFromTopology(topologyInfoNodes)
 	isolatedCPUSet := totalCPUSet.Difference(reservedCPUSet.Result())
-	log.Infof("reservedCPUs: %v len(reservedCPUs): %d\n isolatedCPUs: %v len(isolatedCPUs): %d\n", reservedCPUSet.Result().String(), reservedCPUSet.Result().Size(), isolatedCPUSet.String(), isolatedCPUSet.Size())
+	log.Infof("%d reserved CPUs allocated: %v ", reservedCPUSet.Result().Size(), reservedCPUSet.Result().String())
+	log.Infof("%d isolated CPUs allocated: %v", isolatedCPUSet.Size(), isolatedCPUSet.String())
```
fromanirh, line 344: maybe log once in the calling site? this could require some trivial refactoring

pkg/profilecreator/profilecreator.go
```diff
@@ -290,6 +290,8 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
 func (ghwHandler GHWHandler) getCPUsSplitAcrossNUMA(reservedCPUCount int, htEnabled bool, topologyInfoNodes []*topology.Node) (string, string, error) {
 	reservedCPUSet := cpuset.NewBuilder()
 	numaNodeNum := len(topologyInfoNodes)
+	log.Infof("Number of NUMA nodes on the node are: %d", numaNodeNum)
```
fromanirh, line 293: Let's avoid NUMA nodes/node, I'd use NUMA cells, NUMA zones

cmd/performance-profile-creator/cmd/root.go
marcel-apf: Echoing the input value does no help, indeed. Listing the nodes belonging to the mcp will help.
Until now it was to early to have a decision on "logging" levels, I thought "showing" the user how we compute the values is not "verbose mode", but a way to expose the user how we do things.
If you think this should not be the default option, we need a new command line arg.

swatisehgal: Will update to remove the input values and update add log for listing the nodes belonging to the mcp (missed that).

fromanirh: OK, if we just log the computed field we can leave on as default and add a verbose option later.

pkg/profilecreator/profilecreator.go
swatisehgal: Sure, will do that.

pkg/profilecreator/profilecreator.go
swatisehgal: Ack.

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Ack.

cmd/performance-profile-creator/cmd/root.go
swatisehgal: I felt this was much cleaner than joining multiple sets but okay to update if that's the way we want to implement it. Regarding the second part, I am not sure I understand completely what you mean by data being treated as code. You mean that we should be defining additionalKernelArgs in profilecreator pkg or want me to create a separate pkg just for this?

swatisehgal: Had a discussion with @fromanirh about this and got the clarification on why it should be implemented as a set. Implementing as a slice and using the indices to obtain a portion of the slice has risk in the case the master slice expands or shrinks causing the boundary to change leading to unexpected results.

Also got the clarification that its best to move getAdditionalKernelArgs to a pkg.

pkg/profilecreator/profilecreator.go
```diff
@@ -57,6 +57,16 @@ const (
 	SysInfoFileName = "sysinfo.tgz"
 )
 
+var (
+	// ValidPowerConsumptionModes are a set of valid power consumption modes
+	// default => no args
+	// low-latency => "nmi_watchdog=0", "audit=0",  "mce=off"
+	// ultra-low-latency: low-latency values + "processor.max_cstate=1", "intel_idle.max_cstate=0", "idle=poll"
+	ValidPowerConsumptionModes = []string{"default", "low-latency", "ultra-low-latency"}
```
yanirq, line 65: I wonder if we should maintain a "one source of truth" for these settings. tuned has [profiles](https://github.com/redhat-performance/tuned/tree/master/profiles) for various modes but we are still using a somewhat hybrid approach. 
Not for this PR but something we should consider (cc @marcel-apf )

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -35,8 +35,11 @@ import (
 
 var (
 	validTMPolicyValues = []string{kubeletconfig.SingleNumaNodeTopologyManager, kubeletconfig.BestEffortTopologyManagerPolicy, kubeletconfig.RestrictedTopologyManagerPolicy}
-	// TODO: Explain the power-consumption-mode and associated kernel args
+	// default => no args
+	// low-latency => "nmi_watchdog=0", "audit=0",  "mce=off"
+	// ultra-low-latency: low-latency values + "processor.max_cstate=1", "intel_idle.max_cstate=0", "idle=poll"
```
fromanirh: It would be nice to link https://gist.github.com/wmealing/2dd2b543c4d3cff6cab7 using a comment, but we can do in a later PR.

pkg/profilecreator/profilecreator.go
fromanirh: Yes, it should be a separate (sub-) package.
Tuned moves data like this in configuration, but since in the container world we will bundle everything in a container image (or even in a binary) anyway, a pkg is good enough IMO.

pkg/profilecreator/profilecreator.go
```diff
@@ -445,3 +453,36 @@ func ensureSameTopology(topology1, topology2 *topology.Info) error {
 
 	return nil
 }
+
+// GetAdditionalKernelArgs returns a set of kernel parameters based on the power mode
```
fromanirh, line 457: but the API contract doesn't guarantee to return a sorted list (which is fine but then in the tests we should not assume that)

pkg/profilecreator/profilecreator_test.go
```diff
@@ -498,3 +499,45 @@ var _ = Describe("PerformanceProfileCreator: Test Helper Function ensureSameTopo
...
+	})
+	Context("Ensure kernel args are populated correctly", func() {
+		It("Ensure kernel args are populated correctly in case of low-latency ", func() {
+			powerMode = "low-latency"
+			args := []string{"audit=0",
+				"mce=off",
+				"nmi_watchdog=0",
+			}
+			kernelArgs := GetAdditionalKernelArgs(powerMode)
+			sort.Strings(kernelArgs) // sort to avoid inequality due to difference in order
```
fromanirh, line 522: good idea, but the `GetAdditionalKernelArgs` contract doesn't guarantee a sorted output, so we should sort both.
We can do that on a later PR.

fromanirh (commented):
/lgtm
/hold
minor comments, nothing blocking

fromanirh:
/unhold
per offline chat

PR #579, opened 2021-03-16
Turn off Tuned scheduler plugin dynamic tuning.

jmencak:
The cost of using dynamic tuning (runtime=1) by the Tuned daemon is
currenty very high, and especially so in the OCP environment (see
rhbz#1921738).  PAO uses the Tuned scheduler plugin to set scheduling
policy and priority of ksoftirqd and rcuc kernel threads.  These are
per-CPU kernel threads which start very early during system boot and
exist with a unique PID regardless of CPUs being logically shutdown via
the sysfs interface.

It seems that the only reason for the dynamic scheduler configuration
(runtime=1) would be situations to pick up newly added NUMA
hot-pluggable hardware modules.  This unique case, however, very likely
does not justify the increased CPU utilization incurred by using dynamic
tuning by the Tuned scheduler plugin.

### DIFF
--- a/build/assets/tuned/openshift-node-performance
+++ b/build/assets/tuned/openshift-node-performance
@@ -41,6 +41,7 @@ banned_cpus=""
 {{end}}
 
 [scheduler]
+runtime=0
 group.ksoftirqd=0:f:11:*:ksoftirqd.*
 group.rcuc=0:f:11:*:rcuc.*
 {{if not .GloballyDisableIrqLoadBalancing}}


build/assets/tuned/openshift-node-performance
```diff
@@ -41,6 +41,7 @@ banned_cpus=""
 {{end}}
 
 [scheduler]
+runtime=0
```
marcel-apf, line 44: @MarSik @fromanirh  is it possible to ask MCO to reboot the node after a hotplug operation? Basically saying "OCP does not support dynamic HW changes." Or we can simply document this assumption... but we know it may backfire later.

fromanirh: It boils down to: how frequent are hotplug/hotunplug events in the platforms we care about, especially NUMA hotplug events (!) and how we detect them?

jmencak: > @MarSik @fromanirh is it possible to ask MCO to reboot the node after a hotplug operation? Basically saying "OCP does not support dynamic HW changes." Or we can simply document this assumption... but we know it may backfire later.

Restart of the Tuned daemon would be enough.

MarSik: Does kubelet even support cpu hotplug? I am pretty sure it does not update the available resources in such case.

fromanirh: Last time I checked kubelet could not really handle cpu hotplugs (and perhaps not even memory) without a restart, but we need to check that (and possibly add a test to it?)

marcel-apf: Ok, so what we need is:
 - to be able to detect when hw changes
 - execute a tuned-daemon restart when it happens
 - e2e test to verify we know how to do that
I understand why k8s doesn't care about hw changes, but bare metal OCP should. It may not care today though.
In the later case we can just document it and get back to it as soon as the situation changes.

jmencak: > Ok, so what we need is:
> 
>     * to be able to detect when hw changes
> 
>     * execute a tuned-daemon restart when it happens

At this point you can do this already by changing anything (e.g. a comment) in any of the Tuned profiles.  This is not ideal though and a proper fix would be to have rhbz#1921738 resolved.  The rhbz is planned to be resolved in time for RHEL 8.5.

> 
>     * e2e test to verify we know how to do that
>       I understand why k8s doesn't care about hw changes, but bare metal OCP should. It may not care today though.
>       In the later case we can just document it and get back to it as soon as the situation changes.

In my view I'd probably just document this and then revert by removing this fix once rhbz#1921738 is resolved.

fromanirh: please note that we care about detecting hotplug events for a bunch of reasons, (re)configuring tund is "just" one (important) case among the few we have. So we can totally use a taint/label mechanism even after  rhbz#1921738 is properly fixed.

MarSik: I would split this into two pieces:
- Detecting hotplug events - I like that as it would tell us whether anything more than we have is actually needed at all
- Acting on hotplug events - If kubelet does not care then the tuning can't either, because cpu manager will not know about new cpus anyway (for example). If the OCP way is to reboot the machine (= no hotplug) then there really is no reason for us to do better at this moment. I believe we should use runtime=0 and add a comment that explains it is disabled until https://bugzilla.redhat.com/show_bug.cgi?id=1921738 and/or https://github.com/kubernetes/kubernetes/issues/67500 is fixed.

In other words I think that if OCP node sees a hotplug event during runtime, it is a sysadmin's mistake and the whole node is potentially compromised.

jmencak: Just a note that I've tried whether "logical" hotplug CPU events are automatically picked up by the kubelet by `echo 0 > /sys/devices/system/cpu/cpuX/online`.  They are not.  Manual kubelet restart was needed so that the node would report new CPU capacity.

MarSik: As expected...

fromanirh:
/approve
Given the provided rationale, I think we should merge. My only concern is: how do we test for lack of regressions?

fromanirh:
/lgtm
/hold
we need another ack.

MarSik (approved):

PR #582, opened 2021-03-16
WIP: DONTMERGE: debug podman failures

fromanirh:


fromanirh:
turns out it was a breaking change in how the kernel handles the xattrs. The shadow-utils package, which the must-gather dockerfile wants to install, triggers the usage of xattr, leading to buildah, than podman, to fail: https://github.com/containers/buildah/issues/3071

PR #583, opened 2021-03-16
Consume NTO's per-node Tuned profile application status

Tal-or:
This update allows PAO controller to watch for changes in Tuned profile status, and update its own status based on this new information.

e2e test is WIP.


Tal-or:
cc @fromanirh , @yanirq , @cynepco3hahue

controllers/status.go
```diff
@@ -240,6 +241,49 @@ func (r *PerformanceProfileReconciler) getKubeletConditionsByProfile(profile *pe
...
+				isDegraded = true
+				pDegCondition = &condition
+			}
+
+			if (condition.Type == tunedv1.TunedProfileApplied) && condition.Status == corev1.ConditionFalse {
+				isApplied = false
+			}
+		}
+		// We need both condition to exists,
+		// since there is a scenario where both Degraded & Applied condition are ture
```
fromanirh, line 268: typo: true

controllers/status.go
```diff
@@ -240,6 +241,49 @@ func (r *PerformanceProfileReconciler) getKubeletConditionsByProfile(profile *pe
...
+		if isDegraded == true && isApplied == false {
+			if len(pDegCondition.Reason) > 0 {
+				message.WriteString("Tuned " + tunedProfile.GetName() + " Degraded Reason: " + pDegCondition.Reason + ".\n")
```
fromanirh, line 271: why a `bytes.Buffer` and not, say, just a plain `fmt.Sprintf` ?

controllers/status.go
```diff
@@ -240,6 +241,49 @@ func (r *PerformanceProfileReconciler) getKubeletConditionsByProfile(profile *pe
...
+			}
+			if len(pDegCondition.Message) > 0 {
+				message.WriteString("Tuned " + tunedProfile.GetName() + " Degraded Message: " + pDegCondition.Message + ".\n")
```
fromanirh, line 274: same

controllers/performanceprofile_controller_test.go
```diff
@@ -641,6 +641,57 @@ var _ = Describe("Controller", func() {
...
+				Expect(reconcileTimes(r, request, 1)).To(Equal(reconcile.Result{}))
+
+				updatedProfile := &performancev2.PerformanceProfile{}
+				key := types.NamespacedName{
+					Name:      profile.Name,
+					Namespace: metav1.NamespaceNone,
+				}
+				Expect(r.Get(context.TODO(), key, updatedProfile)).ToNot(HaveOccurred())
+
+				// verify performance profile status
```
fromanirh, line 685: typo: performance

fromanirh (commented):
looks nice, minor comments inside

controllers/performanceprofile_controller_test.go
Tal-or: I don't see any typo here

controllers/status.go
Tal-or: No special reason. Just followed the convention: https://github.com/openshift-kni/performance-addon-operators/blob/master/controllers/status.go#L199

controllers/performanceprofile_controller_test.go
fromanirh: yep, turns out I misread :)

controllers/status.go
fromanirh: well, ok, let's change all these in a later PR

controllers/status.go
```diff
@@ -240,6 +241,49 @@ func (r *PerformanceProfileReconciler) getKubeletConditionsByProfile(profile *pe
...
+	if err := r.List(context.TODO(), tunedProfileList); err != nil {
+		klog.Errorf("Cannot list Tuned Profiles to match with profile %q : %v", profile.Name, err)
+		return nil, err
+	}
+
+	message := bytes.Buffer{}
+	for _, tunedProfile := range tunedProfileList.Items {
+		isDegraded := false
+		isApplied := true
+		var pDegCondition *tunedv1.ProfileStatusCondition
```
yanirq, line 255: maybe lets use full names here without shortcuts , e.g.  tunedDegradedCondition

yanirq:
looks nice.  I would start thinking about e2e tests as well , could be as a followup PR but not neglected

controllers/performanceprofile_controller.go
```diff
@@ -278,6 +278,14 @@ func (r *PerformanceProfileReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
 		}
 	}
 
+	// get tuned profile degraded conditions
+	if conditions == nil {
+		conditions, err = r.getTunedConditionsByProfile(instance)
+		if err != nil {
+			return r.updateDegradedCondition(instance, conditionFailedGettingMCPStatus, err)
```
cynepco3hahue, line 285: You should change `conditionFailedGettingMCPStatus` to something like
`conditionFailedGettingTunedProfileStatus`

controllers/performanceprofile_controller_test.go
```diff
@@ -641,6 +641,77 @@ var _ = Describe("Controller", func() {
 				Expect(degradedCondition.Reason).To(Equal(conditionReasonMCPDegraded))
 				Expect(degradedCondition.Message).To(ContainSubstring(mcpMessage))
 			})
+
+			It("should update status when TunedProfile is degraded", func() {
+				tunedReason := "tunedReason"
+				tunedMessage := "Tuned message"
+
+				tuned := &tunedv1.Profile{
+					TypeMeta: metav1.TypeMeta{},
```
cynepco3hahue, line 650: you can omit empty structs, during the initialization it will fill it automatically, so drop
`TypeMeta: metav1.TypeMeta{}` and `Spec: tunedv1.ProfileSpec{}`

controllers/performanceprofile_controller_test.go
```diff
@@ -641,6 +641,77 @@ var _ = Describe("Controller", func() {
...
+								Status:  corev1.ConditionFalse,
+								Reason:  tunedReason,
+								Message: tunedMessage,
+							},
+						},
+					},
+				}
+
+				nodes := &corev1.NodeList{
+					TypeMeta: metav1.TypeMeta{},
```
cynepco3hahue, line 675: The same omit empty structs

controllers/status.go
```diff
@@ -262,3 +316,15 @@ func removeMCPDuplicateEntries(mcps []mcov1.MachineConfigPool) []mcov1.MachineCo
...
 	return filtered
 }
+
+func removeUnMatchedTunedProfiles(nodes []corev1.Node, profiles []tunedv1.Profile) []tunedv1.Profile {
+	filteredProfiles := make([]tunedv1.Profile, 0)
+	for _, profile := range profiles {
+		for _, node := range nodes {
+			if profile.Name == node.Name {
+				filteredProfiles = append(filteredProfiles, profile)
+			}
```
cynepco3hahue, line 326: nit: in general you can add break here, it not possible that you will have two nodes with the same name

controllers/status.go
```diff
@@ -240,6 +242,58 @@ func (r *PerformanceProfileReconciler) getKubeletConditionsByProfile(profile *pe
...
+				isDegraded = true
+				tunedDegradedCondition = &condition
+			}
+
+			if (condition.Type == tunedv1.TunedProfileApplied) && condition.Status == corev1.ConditionFalse {
+				isApplied = false
+			}
+		}
+		// We need both condition to exists,
+		// since there is a scenario where both Degraded & Applied condition are true
```
cynepco3hahue, line 278: I am curious when can you have degraded true and applied true?

Tal-or: You should refer this question to @yanirq .
But as far as I understood there is a situation when tuned configuration applied with errors. then is will be both applied and degraded as true

cynepco3hahue:
You should check performance addon logs under the CI job, please do:
1. Click on details under the CI job, it will bring you to prow job
2. Find artifacts link and click on it
3. Click on the artifacts folder
4. Click on the folder with the job name(hint: in our case, it can be `e2e-gcp`)
5. Go to `gather-extra->artifacts->pods` folder(it can be a good idea to check all folders to see what kind of logs is available)
6. Find the relevant performance addon operator log
7. Check the log for errors

controllers/status.go
yanirq: This is something that should be eventually fixed under tuned. tuned has a buffer that keeps errors and this buffer is read by NTO.  
When NTO detects errors in that buffer it updates the profile state accordingly but it could get to a state where a profile was successfully applied  but the tuned error buffer was not cleaned up , therefore NTO  sets both status as true (degraded + applied)  that is why we should have our status degraded only when degraded = true and applied = false

controllers/performanceprofile_controller.go
```diff
@@ -178,10 +178,10 @@ func validateUpdateEvent(e *event.UpdateEvent) bool {
 	return true
 }
 
-// +kubebuilder:rbac:groups="",resources=events,verbs=*
+// +kubebuilder:rbac:groups="",resources=events;nodes,verbs=*
```
cynepco3hahue, line 181: Please reduce the number of permissions, in general, `get,list,watch` should be enough, I think something like 
```
// +kubebuilder:rbac:groups="",resources=nodes,verbs=get,list,watch
```
should work

controllers/performanceprofile_controller.go
```diff
@@ -178,10 +178,10 @@ func validateUpdateEvent(e *event.UpdateEvent) bool {
...
 // +kubebuilder:rbac:groups=performance.openshift.io,resources=performanceprofiles;performanceprofiles/status;performanceprofiles/finalizers,verbs=*
 // +kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigs;machineconfigpools;kubeletconfigs,verbs=*
-// +kubebuilder:rbac:groups=tuned.openshift.io,resources=tuneds,verbs=*
+// +kubebuilder:rbac:groups=tuned.openshift.io,resources=tuneds;profiles,verbs=*
 // +kubebuilder:rbac:groups=node.k8s.io,resources=runtimeclasses,verbs=*
 // +kubebuilder:rbac:namespace="openshift-performance-addon-operator",groups=core,resources=pods;services;services/finalizers;configmaps,verbs=*
 // +kubebuilder:rbac:namespace="openshift-performance-addon-operator",groups=apps,resources=deployments;daemonsets;replicasets;statefulsets,verbs=*
```
cynepco3hahue, line 187: not related to this PR, but we should remove unrelated to the PAO permissions like we should not need any permissions for `deployments;daemonsets;replicasets;statefulsets`

fromanirh: big giant +1

Tal-or: I can open a separate story for that. Is that works for you?

cynepco3hahue: yep works for me

controllers/status.go
```diff
@@ -240,6 +243,58 @@ func (r *PerformanceProfileReconciler) getKubeletConditionsByProfile(profile *pe
 	return r.getDegradedConditions(conditionKubeletFailed, latestCondition.Message), nil
 }
 
+func (r *PerformanceProfileReconciler) getTunedConditionsByProfile(profile *performancev2.PerformanceProfile) ([]conditionsv1.Condition, error) {
+	tunedProfileList := &tunedv1.ProfileList{}
+	if err := r.List(context.TODO(), tunedProfileList); err != nil {
```
yanirq, line 248: Where are we doing the match between the performance profile tuned name and the existing tuneds in the cluster here?

Tal-or: We don't. We should look for a match between the node's name and tuned's profile name, as you can see [here](https://github.com/openshift-kni/performance-addon-operators/pull/583/files#diff-e8b328ffb5ced885894c80b942846601597e05020f831c2b1442d8fb8a654c8dR325)
The node itself associate with the PP via the ```nodeSelector```

yanirq: right. you are getting all the profiles.

PR #584, opened 2021-03-16
profilecreator: Support SNO clusters

marcel-apf:
must-gather returns the master nodes only if they are schedulable.
Pick up the master MCP, but do not error out on failure.


pkg/profilecreator/profilecreator.go
```diff
@@ -172,13 +172,9 @@ func GetMCPList(mustGatherDirPath string) ([]*machineconfigv1.MachineConfigPool,
...
 
-		// must-gather does not return the master nodes
-		if mcpName == "master" {
-			continue
-		}
-
 		mcp, err := GetMCP(mustGatherDirPath, mcpName)
-		if err != nil {
+		// master pool relevant only when pods can be scheduled on masters, e.g. SNO
+		if mcpName != "master" && err != nil {
```
fromanirh, line 177: this may change soon enough in `control-plane`

fromanirh (commented):
/lgtm
/hold
looks good (lgtm) but (hold) do we have tests to cover this case?

pkg/profilecreator/profilecreator.go
marcel-apf: We only say "is not an error if the MCP hasn't been picked up by the must-gather".
What can we do better? How can we fail ?

fromanirh: Not sure, I guess we can only track k8s/ocp and update the literal string.

marcel-apf: :(

marcel-apf:
> /lgtm
> /hold
> looks good (lgtm) but (hold) do we have tests to cover this case?

It worth a test for sure, thanks! e2e tests + test coverage are scheduled for next sprint.
The easiest way to do e2e test is to add a few tar.gz tarballs as assets, would it be ok?

fromanirh:
> > /lgtm
> > /hold
> > looks good (lgtm) but (hold) do we have tests to cover this case?
> 
> It worth a test for sure, thanks! e2e tests + test coverage are scheduled for next sprint.
> The easiest way to do e2e test is to add a few tar.gz tarballs as assets, would it be ok?

Adding assets as tgz is fine by me.
Adding e2e tests later is not ideal. The sooner the stop the better. But I'm ok not adding e2e tests *now*. Any kind of test coverage is enough - be it unit test, or coverage (even partial) from existing tests.
Adding untested change is not something we need to do unless is the very last resort.

marcel-apf:
> > > /lgtm
> > > /hold
> > > looks good (lgtm) but (hold) do we have tests to cover this case?
> > 
> > 
> > It worth a test for sure, thanks! e2e tests + test coverage are scheduled for next sprint.
> > The easiest way to do e2e test is to add a few tar.gz tarballs as assets, would it be ok?
> 
> Adding assets as tgz is fine by me.
> Adding e2e tests later is not ideal. The sooner the stop the better. But I'm ok not adding e2e tests _now_. Any kind of test coverage is enough - be it unit test, or coverage (even partial) from existing tests.
> Adding untested change is not something we need to do unless is the very last resort.

E2E tests need some thinking. I'll add an SNO must-gather to unit-tests, thanks

fromanirh:
/lgtm
thanks for adding the tests.

PR #585, opened 2021-03-19
profile-creator: Friendlier error messages

marcel-apf:
Is not a server side component, is a cmd-line tool.
Look for errors early and come up with friendly user messages.

### DIFF
--- a/cmd/performance-profile-creator/cmd/root.go
+++ b/cmd/performance-profile-creator/cmd/root.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/openshift-kni/performance-addon-operators/pkg/profilecreator"
 	"github.com/openshift-kni/performance-addon-operators/pkg/utils/csvtools"
+	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cmd/performance-profile-creator/cmd/root.go
```diff
@@ -127,25 +128,48 @@ func getDataFromFlags(cmd *cobra.Command) (profileCreatorArgs, error) {
...
 	mcps, err := profilecreator.GetMCPList(args.mustGatherDirPath)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get the MCP list under %s: %v", args.mustGatherDirPath, err)
 	}
 
+	var mcp *machineconfigv1.MachineConfigPool
+	mcpNames := make([]string, 0)
+	for _, m := range mcps {
+		mcpNames = append(mcpNames, m.Name)
+		if m.Name == args.mcpName {
```
swatisehgal, line 151: Suggestion: once we find a match, we can break from the loop

swatisehgal (commented):
A minor comment, but other than that look good !

cmd/performance-profile-creator/cmd/root.go
marcel-apf: We can't, because we are accumulating all the mcpNames in the cluster in order to help the user:
Couldn't find the the mcp X, try one of [mcpNames]

swatisehgal: Okay, got it! Thanks for the clarification, that makes sense!

marcel-apf: Actually, if we did find the VM we don't need the mcpNames anyway since we have the right MCP. However there may be other reasons we can't use the mcp: no nodes defined and so on. So we better keep it anyway. Also we may re-use the mcp names for other info  logs.

marcel-apf:
> /lgtm

thank you!

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Ack.

PR #586, opened 2021-03-22
profile-creator: Always add the MCP selector to the profile

marcel-apf:
Find the unique MCP labels, if more than one found prefer
the one including the mcp-name, pick a random one otherwise.

Refuse to create the profile if no unique label exists.

### DIFF
--- a/pkg/profilecreator/mcp.go
+++ b/pkg/profilecreator/mcp.go
@@ -2,6 +2,7 @@ package profilecreator
 
 import (
 	"fmt"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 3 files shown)


marcel-apf:
/hold
needs (more) testing

marcel-apf:
Added unit-tests
/unhold

pkg/profilecreator/mcp.go
```diff
@@ -12,6 +13,56 @@ import (
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 )
 
+// GetMCPSelector returns a label that is unique to the target pool, error otherwise
+func GetMCPSelector(pool *mcfgv1.MachineConfigPool, clusterPools []*mcfgv1.MachineConfigPool) (map[string]string, error) {
```
fromanirh, line 17: I wonder if this function should just return `key, value, error` since the expected size of the `map[string]string` it returns now is always 1

pkg/profilecreator/profilecreator_test.go
```diff
@@ -54,6 +54,38 @@ var _ = Describe("PerformanceProfileCreator: MCP and Node Matching", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(mcpSelector)).To(Equal(1))
+
+			for key, value := range mcpSelector {
+				Expect(key).To(Equal("pools.operator.machineconfiguration.openshift.io/worker"))
+				Expect(value).To(Equal(""))
+				break
+			}
+		})
+	})
```
fromanirh, line 88: do we want to add negative tests -  e.g. tests that check the error path, not just the happy path? perhaps in a later PR?

pkg/profilecreator/profilecreator_test.go
```diff
@@ -89,6 +121,37 @@ var _ = Describe("PerformanceProfileCreator: MCP and Node Matching in SNO", func
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(mcpSelector)).To(Equal(1))
+
+			for key, value := range mcpSelector {
+				Expect(key).To(Equal("pools.operator.machineconfiguration.openshift.io/master"))
+				Expect(value).To(Equal(""))
+				break
+			}
+		})
+	})
```
fromanirh, line 154: same

fromanirh (commented):
/lgtm
/hold
because of inline (non blocking) comments

pkg/profilecreator/mcp.go
marcel-apf: I decided to keep it that way since the receiver of this function result is the PerformanceProfile which expects the same type: map[string]string for the MCP selector. I take it is some kind of "pattern".

pkg/profilecreator/profilecreator_test.go
marcel-apf: I am all for adding more tests if it helps. I wonder how a negative test would look like and what would be the added value.  A non existing MCP is already tested before the MCP selector, so not an issue. An interesting "fail" path would be a cluster configuration where we cannot find an unique MCP selector for the target MCP. But I don't have a concrete example... maybe we should add it to the E2E tests if we can come up with something.

swatisehgal (approved):
/lgtm

PR #587, opened 2021-03-23
Enable disable-ht flag and user-level-networking in PPC

swatisehgal:
1. Follow up from PR https://github.com/openshift-kni/performance-addon-operators/pull/575
2. Populate userLevelNetworking field in the performance profile
2. Enable disable-ht flag in PPC
3. Remove unused code and log system info to improve user experience

### DIFF
--- a/cmd/performance-profile-creator/cmd/root.go
+++ b/cmd/performance-profile-creator/cmd/root.go
@@ -36,8 +36,6 @@ import (
 
 var (
 	validTMPolicyValues = []string{kubeletconfig.SingleNumaNodeTopologyManagerPolicy, kubeletconfig.BestEffortTopologyManagerPolicy, kubeletconfig.RestrictedTopologyManagerPolicy}
-	// TODO: Explain the power-consumption-mode and associated kernel args
-	validPowerConsumptionModes = []string{"default", "low-latency", "ultra-low-latency"}
 )
 
 // ProfileData collects and stores all the data needed for profile creation
--- a/cmd/performance-profile-creator/cmd/root.go
+++ b/cmd/performance-profile-creator/cmd/root.go
@@ -49,6 +49,7 @@ type ProfileData struct {
 	rtKernel                   bool
 	additionalKernelArgs       []string
 	userLevelNetworking        bool
+	disableHT                  bool
 }
 
 // rootCmd represents the base command when called without any subcommands
@@ -119,6 +120,11 @@ func getDataFromFlags(cmd *cobra.Command) (profileCreatorArgs, error) {
 	if err != nil {
 		return creatorArgs, fmt.Errorf("failed to parse user-level-networking flag: %v", err)
 	}
+
+	htDisabled, err := strconv.ParseBool(cmd.Flag("disable-ht").Value.String())
+	if err != nil {
+		return creatorArgs, fmt.Errorf("failed to parse disable-ht flag: %v", err)
+	}
 	creatorArgs = profileCreatorArgs{
 		mustGatherDirPath:           mustGatherDirPath,
 		profileName:                 profileName,
@@ -129,6 +135,7 @@ func getDataFromFlags(cmd *cobra.Command) (profileCreatorArgs, error) {
 		rtKernel:                    rtKernelEnabled,
 		powerConsumptionMode:        powerConsumptionMode,
 		userLevelNetworking:         userLevelNetworkingEnabled,
+		disableHT:                   htDisabled,
 	}
 	return creatorArgs, nil
 }
@@ -191,13 +198,13 @@ func getProfileData(args profileCreatorArgs) (*ProfileData, error) {
 	// same from hardware topology point of view
 
 	handle, err := profilecreator.NewGHWHandler(args.mustGatherDirPath, matchedNodes[0])
-	reservedCPUs, isolatedCPUs, err := handle.GetReservedAndIsolatedCPUs(args.reservedCPUCount, args.splitReservedCPUsAcrossNUMA)
+	reservedCPUs, isolatedCPUs, err := handle.GetReservedAndIsolatedCPUs(args.reservedCPUCount, args.splitReservedCPUsAcrossNUMA, args.disableHT)
 	if err != nil {
 		return nil, fmt.Errorf("failed to compute the reserved and isolated CPUs: %v", err)
 	}
 	log.Infof("%d reserved CPUs allocated: %v ", reservedCPUs.Size(), reservedCPUs.String())
 	log.Infof("%d isolated CPUs allocated: %v", isolatedCPUs.Size(), isolatedCPUs.String())
-	kernelArgs := profilecreator.GetAdditionalKernelArgs(args.powerConsumptionMode)
+	kernelArgs := profilecreator.GetAdditionalKernelArgs(args.powerConsumptionMode, args.disableHT)
 	profileData := &ProfileData{
 		reservedCPUs:           reservedCPUs.String(),
 		isolatedCPUs:           isolatedCPUs.String(),
--- a/cmd/performance-profile-creator/cmd/root.go
+++ b/cmd/performance-profile-creator/cmd/root.go
@@ -48,6 +48,7 @@ type ProfileData struct {
 	topologyPoilcy             string
 	rtKernel                   bool
 	additionalKernelArgs       []string
+	userLevelNetworking        bool
 }
 
 // rootCmd represents the base command when called without any subcommands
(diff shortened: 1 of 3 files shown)


pkg/profilecreator/profilecreator.go
```diff
@@ -255,8 +256,37 @@ func (ghwHandler GHWHandler) SortedTopology() (*topology.Info, error) {
 	return topologyInfo, nil
 }
 
+func (ghwHandler GHWHandler) TopologyHTDisabled(info *topology.Info) *topology.Info {
```
fromanirh, line 260: this should be a free function: we never use `ghwHandler`, do we?

pkg/profilecreator/profilecreator.go
```diff
@@ -255,8 +256,37 @@ func (ghwHandler GHWHandler) SortedTopology() (*topology.Info, error) {
...
+		cores := []*cpu.ProcessorCore{}
+		for _, processorCore := range node.Cores {
+			newCore := cpu.ProcessorCore{ID: processorCore.ID,
+				Index:      processorCore.Index,
+				NumThreads: 1,
+			}
+			for id, logicalProcessor := range processorCore.LogicalProcessors {
+				if id == 0 {
+					newCore.LogicalProcessors = []int{logicalProcessor}
+					cores = append(cores, &newCore)
```
fromanirh, line 276: I'm not completely sure of the logic here.
Of course we will need to delete some virtual cores, that's fine, but why we are keeping those whose `id == 0` ?
Do we depend on ghw behaviour, on kernel representation or what? This rationale should be made explicit with a code comment.

pkg/profilecreator/profilecreator.go
```diff
@@ -273,10 +303,17 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
 	if err != nil {
 		return cpuset.CPUSet{}, cpuset.CPUSet{}, fmt.Errorf("can't determine if Hyperthreading is enabled or not: %v", err)
 	}
+	//currently HT is enabled on the system and the user wants to disable HT
+	if htEnabled && disableHTFlag {
+		log.Infof("Currently hyperthreading is enabled and the performance profile will disable it")
+		topologyInfo = ghwHandler.TopologyHTDisabled(topologyInfo)
+
```
fromanirh, line 311: so here inside this `if` we can now do
```
htEnabled =  false
```
and just use `htEnabled` from now on, without the need of carrying with us two separate flags, right?

pkg/profilecreator/profilecreator.go
```diff
@@ -309,6 +310,18 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
 
 	}
+	log.Infof("NUMA cell(s): %d", len(topologyInfo.Nodes))
+	totalCPUs := 0
+	for id, node := range topologyInfo.Nodes {
+		coreList := []int{}
+		for _, core := range node.Cores {
+			coreList = append(coreList, core.LogicalProcessors...)
+		}
+		log.Infof("NUMA cell %d : %v", id, coreList)
```
fromanirh, line 320: why we cannot just use `core.LogicalProcessoers`?

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -36,8 +36,6 @@ import (
 
 var (
 	validTMPolicyValues = []string{kubeletconfig.SingleNumaNodeTopologyManagerPolicy, kubeletconfig.BestEffortTopologyManagerPolicy, kubeletconfig.RestrictedTopologyManagerPolicy}
-	// TODO: Explain the power-consumption-mode and associated kernel args
-	validPowerConsumptionModes = []string{"default", "low-latency", "ultra-low-latency"}
```
fromanirh, line 40: why is this gone?

fromanirh (commented):
few comments inside about the HT disabling logic. Didn't review the related tests yet.

cmd/performance-profile-creator/cmd/root.go
swatisehgal: This was moved to profilecreator pkg.

pkg/profilecreator/profilecreator.go
swatisehgal: You mean why we are using `core.LogicalProcessors...` here?

pkg/profilecreator/profilecreator.go
swatisehgal: We need two separate flags as `htEnabled` represents the current state of hyperthreading on the system and `disableHTFlag` represents the intended state of the HT that the user wants the system to be in.

pkg/profilecreator/profilecreator.go
swatisehgal: In cases the user wants to disable hyperthreading, while accounting reserved and isolated CPUs we are keeping only the first hyperthread sibling. Example in case in the system there are cores such as {0, 40}, {1,41}, {2, 42}, {3, 43} .... we change that to {0}, {1}, {2}, {3}.... and then do the accounting of reserved and isolated CPU. Will add a comment to explain the rationale.

pkg/profilecreator/profilecreator.go
swatisehgal: Right! Will update.

pkg/profilecreator/profilecreator.go
swatisehgal: Also, the topologyinfo received here is already sorted by numa ids and cpu ids on the host system.

fromanirh: ack, please make explicit if we depend on ghw representation/behaviour.

pkg/profilecreator/profilecreator.go
fromanirh: ok, but where this difference materializes? in both `getCPUsSplitAcrossNUMA` and `getCPUsSequentially` I always see the flag being used like `htEnabled && !disableHTFlag`

pkg/profilecreator/profilecreator.go
fromanirh: Nevermind, got it. We want all the logical processors of all the cores in one go.

pkg/profilecreator/profilecreator.go
swatisehgal: I understand now what you mean!
The reason why it is implemented like this is as follows:
1. In case hyperthreading is enabled on the system (htEnabled = true) and the user does not want to disable it (disableHTFlag=false) we can't allow odd number of CPUs to be allocated from a NUMA node.
In rest of the cases (as follows) we can have odd number of CPUs to be allocated from a NUMA node
2. htEnabled = false disableHTFlag= false
3. htEnabled = true disableHTFlag= true
4. htEnabled = false disableHTFlag= true

but what you are saying makes sense now as in case 3 if assign htEnable=false and that would mean that  htEnabled && !disableHTFlag would return false for all case 2, 3 and 4 (achieving the behavior we want).

cmd/performance-profile-creator/cmd/root.go
fromanirh: ok, so this was a dupe I guess.

swatisehgal: Yes, we forgot to remove it in a previous PR. I combined it with a smaller set of changes where we add logs to show system info. But will keep in mind for the future to have changes like this in a completely separate commit.

pkg/profilecreator/profilecreator.go
```diff
@@ -363,13 +365,7 @@ func (ghwHandler GHWHandler) getCPUsSplitAcrossNUMA(reservedCPUCount int, htEnab
...
 			return reservedCPUSet.Result(), isolatedCPUSet, fmt.Errorf("can't allocate odd number of CPUs from a NUMA Node")
 		}
-		for _, processorCores := range node.Cores {
-			for _, core := range processorCores.LogicalProcessors {
-				if reservedCPUSet.Result().Size() < max {
-					reservedCPUSet.Add(core)
-				}
-			}
-		}
+		addCorestoCPUSet(reservedCPUSet, max, node.Cores)
```
fromanirh, line 368: nit: should probably be `addCoresToCPUSet` (captal T)

swatisehgal:
Had an offline conversation with @fromanirh. ARs are:

1. Add comments to explain the rationale behind the implementation of  topologyHTDisabled.
2. In context of addCorestoCPUSetAvoid (to be renamed to addCoresToCPUSetAvoid), rather than using intMax for the case where all the cores need to be added to a CPUSet. We would switch to using "-1" as a value indicative of a case where all the CPUs need to be added to a CPUset.

fromanirh (approved):
/lgtm
thanks for the followups.

PR #592, opened 2021-04-04
Change net api fields to more logical names

yanirq:
Spec.net field names changes:
vendor => vendorId
device => deviceId

### DIFF
--- a/api/v2/performanceprofile_types.go
+++ b/api/v2/performanceprofile_types.go
@@ -144,10 +144,10 @@ type Device struct {
 	InterfaceName *string `json:"interfaceName,omitempty"`
 	// Network device vendor ID represnted as a 16 bit Hexmadecimal number.
 	// +optional
-	VendorID *string `json:"vendor,omitempty"`
+	VendorID *string `json:"vendorId,omitempty"`
 	// Network device ID (model) represnted as a 16 bit hexmadecimal number.
 	// +optional
-	DeviceID *string `json:"device,omitempty"`
+	DeviceID *string `json:"deviceId,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


api/v2/performanceprofile_types.go
```diff
@@ -144,10 +144,10 @@ type Device struct {
 	InterfaceName *string `json:"interfaceName,omitempty"`
 	// Network device vendor ID represnted as a 16 bit Hexmadecimal number.
 	// +optional
-	VendorID *string `json:"vendor,omitempty"`
+	VendorID *string `json:"vendorId,omitempty"`
```
fromanirh, line 147: should be `vendorId` or `vendorID` (lowercase or uppercase 'D') ?
Here and in all the following instances.
Everything else lgtm

yanirq: Actually I did'nt stumble upon a good guideline for this (especially in our code)  so I decided to go with a Capital separator. 
I'm fine either way

fromanirh: > Actually I did'nt stumble upon a good guideline for this (especially in our code) so I decided to go with a Capital separator.
> I'm fine either way

Pretty much the same here. I'd say:
- let's be consistent with our current API - if relevant (can't remember if we have already someIds)
- let's just quickly review what's the latest standard on ocp/k8s, and file an issue should we change the code.

yanirq: looking at some openshift/k8s projects it seems they all comply to using ID/UID in capitals. changing accordingly

fromanirh (commented):
/lgtm

PR #595, opened 2021-04-06
profile-creator: Add E2E regression tests

marcel-apf:
Use expected profiles files, encoding in the file names the arguments.
The approach allows adding use-cases by having more profiles
or must-gather data added to the project, without the need to add code.

Add a few profiles to verify the regression mechanism.
More profiles will be added as part of the QE test plans.


functests-performance-profile-creator/1_performance-profile_creator/ppc.go
```diff
@@ -0,0 +1,159 @@
...
+	It("[test_id:xxxxx] performance profile creator regression tests", func() {
+		Expect(ppcPath).To(BeAnExistingFile())
+
+		// directory base name => full path
+		mustGatherDirs := getMustGatherDirs(mustGatherPath)
+		// full profile path => arguments the profile was created with
+		expectedProfiles := getExpectedProfiles(expectedProfilesPath, mustGatherDirs)
+
+		for expectedProfilePath, args := range expectedProfiles {
+			cmdArgs := fmt.Sprintf("--disable-ht=%v --mcp-name=%s --must-gather-dir-path=%s "+
```
fromanirh, line 40: it seems we never use `cmdArgs`, do we? If so, why not just create the `args` slice straight away? as extra bonus, that would make the format strings easier to follow:
```
cmdArgs = []string{
   fmt.Sprintf("--disable-ht=%v", args.DisableHT),
   fmt.Sprintf("--mcp-name=%s", args.MCPName),
  ...
}
```

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
```diff
@@ -0,0 +1,159 @@
...
+	expectedProfilesPathContent, err := ioutil.ReadDir(expectedProfilesPath)
+	Expect(err).To(BeNil(), fmt.Errorf("can't list '%s' files: %v", expectedProfilesPath, err))
+
+	for _, file := range expectedProfilesPathContent {
+		fullFilePath := filepath.Join(expectedProfilesPath, file.Name())
+		Expect(filepath.Ext(fullFilePath)).To(Equal(".yaml"))
+
+		var ppcArgs cmd.ProfileCreatorArgs
+
+		args := strings.Split(strings.TrimSuffix(file.Name(), filepath.Ext(file.Name())), argsDelimiter)
```
fromanirh, line 97: what's the benefit of encoding the expected values in the file name?

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
marcel-apf: will, do thanks!

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
marcel-apf: Each profile yaml file will by a self contained test (input/output)
I added 2 use-cases in the PR

fromanirh: Sure, but why, for example, we cannot read back these values from the profile themselves? IOW, why should we encode this data *in the file names*? This is pretty much like adding a little DSL in disguise, and I'm trying to understand if we have simpler alternatives. Like just reading the manifests themselves :)

marcel-apf: This would a be a good idea if we would have been able to "easily" map profile values to ppc arguments, however this is not the case. Tracing back a certain "isolated cpus"/"reserved cpus" partition to --reserved-cpus=? -split-reserved-cpus-across-numa-nodes=? ht-disabled=? is far from trivial and the complexity is "almost" as the original code.   The drawback is that a failed e2e test may be caused by the test implementation and not the tool itself.

While I do agree the args-encoded-in-file-name has a "smell", is not really a DSL, is a simple key-value list and will bever go beyond that. The last point is we want to make it easy for the QE to map their testing "matrix" (args combination -> profile) easier.

If you find this encoding as a  "no-go", I can add a file that defines the mappings, it will basically look the same, a list of key-value pairs pointing to a yaml file.

fromanirh: good point. So what about a trivial `ppc-params.json` alongside the profile?
```json
{
   "reserved-cpus": "0-3",
   "split-reserved-cpus-across-numa-nodes": false,
   ...
}
```

fromanirh: Yup, totally. IMO this is both simpler to code (and maintain) and to extend later. We can just add annotations to ProfileCreatorArgs, `json.Unmarshal` it and call it a day.

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
```diff
@@ -0,0 +1,159 @@
...
+const (
+	mustGatherPath       = "../../testdata/must-gather"
+	expectedProfilesPath = "../../testdata/ppc-expected-profiles"
+	ppcPath              = "../../build/_output/bin/performance-profile-creator"
+
+	argsDelimiter     = "_"
+	keyValueDelimiter = ":"
+)
+
+var _ = Describe("[rfe_id:xxxxx][ppc] Performance Profile Creator", func() {
```
swatisehgal, line 29: Adding a note here to make sure that we populate the rfe_id before we merge this PR.

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
swatisehgal: +1 on having args in a separate json file!

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
marcel-apf: done

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
marcel-apf: done

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
marcel-apf: done

testdata/ppc-expected-profiles/profile2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "mg": "must-gather.sno",
```
swatisehgal, line 2: For more clarity here, I would recommend using the flag names here ( as we use when we run PPC) i.e. `must-gather-dir-path` instead of `mg`, `mcp-name` instead of `mcp`,  `reserved-cpu-count` instead of `reserved` and `rt-kernel` instead of  `rt`.

swatisehgal (changes requested):

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
```diff
@@ -0,0 +1,136 @@
...
+	for _, file := range expectedProfilesPathContent {
+		if filepath.Ext(file.Name()) != ".json" {
+			continue
+		}
+
+		fullFilePath := filepath.Join(expectedProfilesPath, file.Name())
+		bytes, err := ioutil.ReadFile(fullFilePath)
+		Expect(err).To(BeNil(), "failed to read the ppc params file for '%s': %v", fullFilePath, err)
+
+		ppcArgs := cmd.ProfileCreatorArgs{
```
swatisehgal, line 103: Let's add a comment here to explain that we are populating non-required creator args with default values. Might be worth explaining that other non-required fields that are not here: disableHT, userLevelNetworking and reservedCPUCount automatically default to false, false and 0 respectively because of zero values in go.

testdata/ppc-expected-profiles/profile2.json
marcel-apf: Sure, I can do that

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
marcel-apf: Right, I am going to not pass the empty string values to the tool (not an interesting scenario anyway)

testdata/ppc-expected-profiles/profile2.json
marcel-apf: done

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
marcel-apf: please see the proposed change

swatisehgal: Looks good, Thanks!

PR #597, opened 2021-04-08
Add management workload annotations

lack:
In support of the workload partitioning feature
(https://github.com/openshift/enhancements/pull/703), we need to add
annotations to all management pods and namespaces so they can be
properly identified and assigned to segregated management cores on
clusters configured to do so.

### DIFF
--- a/deploy/olm-catalog/performance-addon-operator/4.5.0/performance-addon-operator.v4.5.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/performance-addon-operator/4.5.0/performance-addon-operator.v4.5.0.clusterserviceversion.yaml
@@ -109,6 +109,8 @@ spec:
           strategy: {}
           template:
             metadata:
+              annotations:
+                workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
               labels:
                 name: performance-operator
             spec:
--- a/must-gather/node-gather/daemonset.yaml
+++ b/must-gather/node-gather/daemonset.yaml
@@ -11,6 +11,8 @@ spec:
       name: perf-node-gather-daemonset
   template:
     metadata:
+      annotations:
+        workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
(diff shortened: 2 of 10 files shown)


deploy/olm-catalog/performance-addon-operator/4.5.0/performance-addon-operator.v4.5.0.clusterserviceversion.yaml
```diff
@@ -109,6 +109,8 @@ spec:
           strategy: {}
           template:
             metadata:
+              annotations:
+                workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
```
fromanirh, line 113: I don't think we need to add annotations to old (pre 4.7) manifests, but won't hurt either.

must-gather/node-gather/daemonset.yaml
```diff
@@ -11,6 +11,8 @@ spec:
       name: perf-node-gather-daemonset
   template:
     metadata:
+      annotations:
+        workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
```
fromanirh, line 15: AFAIU must-gather is supposed to follow the same conventions as regular infrastructure pods, right?

lack: Correct.  Everything that's "OCP" is intended to be treated in the same workload / class of service.

fromanirh: cool, then it seems we're good.

fromanirh:
/lgtm
/hold
let's have another ack before to merge.

deploy/olm-catalog/performance-addon-operator/4.5.0/performance-addon-operator.v4.5.0.clusterserviceversion.yaml
cynepco3hahue: We should not forget to update d/s manifests.

PR #598, opened 2021-04-11
E2E test for consume NTO's per-node Tuned profile application status

Tal-or:
The test will create a TuneD object with a TuneD profile that doesn't exist.
Tuned daemon will fail to apply it, therefore it will get into a Degraded state which eventually should be reflected in PAO status.
In the end, we remove the bad Tuned object and wait for PAO status condition to be Available again.

### DIFF
--- a/functests/3_performance_status/status.go
+++ b/functests/3_performance_status/status.go
@@ -44,7 +44,7 @@ var _ = Describe("Status testing of performance profile", func() {
 
 	Context("[rfe_id:28881][performance] Performance Addons detailed status", func() {
 
-		It("[test_id:30894] Tuned status field tied to Performance Profile", func() {
+		It("[test_id:30894] Tuned status name tied to Performance Profile", func() {
 			profile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
 			Expect(err).ToNot(HaveOccurred())
 			key := types.NamespacedName{
@@ -103,6 +103,21 @@ var _ = Describe("Status testing of performance profile", func() {
 
 			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
 		})
+
+		It("[test_id:40402] Tuned profile status tied to Performance Profile", func() {
+			// Creating bad Tuned object that leads to degraded state
+			badTuned := createBadTuned("openshift-cause-tuned-failure", "openshift-cluster-node-tuning-operator")
+			err = testclient.Client.Create(context.TODO(), badTuned)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Waiting for performance profile condition to be Degraded")
+			profiles.WaitForCondition(testutils.NodeSelectorLabels, v1.ConditionDegraded, corev1.ConditionTrue)
+
+			By("Deleting bad Tuned and waiting when Degraded state is removed")
+			err = testclient.Client.Delete(context.TODO(), badTuned)
+			profiles.WaitForCondition(testutils.NodeSelectorLabels, v1.ConditionAvailable, corev1.ConditionTrue)
+		})
+
 	})
 })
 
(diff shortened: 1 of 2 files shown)


functests/3_performance_status/status.go
```diff
@@ -139,3 +154,35 @@ func createBadMachineConfig(name string) *machineconfigv1.MachineConfig {
...
+		Spec: tunedv1.TunedSpec{
+			Profile: []tunedv1.TunedProfile{
+				{
+					Name: &name,
+					Data: &data,
+				},
+			},
+			Recommend: []tunedv1.TunedRecommend{
+				{
+					MachineConfigLabels: map[string]string{"machineconfiguration.openshift.io/role": testutils.RoleWorkerCNF},
```
yanirq, line 180: maybe we should explain the connection here (or in the test itself) . its not intuitive that creating a misconfigured tuned CR  with the same `machineconfiguration.openshift.io/role` will lead a degraded state of the tuned PAO creates from the performance profile

functests/3_performance_status/status.go
```diff
@@ -103,6 +103,21 @@ var _ = Describe("Status testing of performance profile", func() {
...
 		})
+
+		It("[test_id:40402] Tuned profile status tied to Performance Profile", func() {
+			// Creating bad Tuned object that leads to degraded state
+			badTuned := createBadTuned("openshift-cause-tuned-failure", "openshift-cluster-node-tuning-operator")
+			err = testclient.Client.Create(context.TODO(), badTuned)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Waiting for performance profile condition to be Degraded")
+			profiles.WaitForCondition(testutils.NodeSelectorLabels, v1.ConditionDegraded, corev1.ConditionTrue)
```
fromanirh, line 114: if this fails, is the bad tuned removed anyway?

functests/3_performance_status/status.go
Tal-or: Because it's not. the tuned that PAO created is perfectly fine.
During this test we're creating additional synthetic tuned CR by invoking the ```createrBadTuned``` function.
This synthetic tuned will look for a tuned profile which doesn't exist. 
see ```include=profile-does-not-exist``` line in [here](https://github.com/openshift-kni/performance-addon-operators/pull/598/files/747cbc5b937af9aded6afda395c8359c747db471#diff-46dd9c2fd680952d6ae2734f9f4b9d42734d386f897dcb79cf030f17eae6f8f4R160)

This tuned CR will be applied on the ```profiles.tuned.openshift.io``` CR (there is such profile per node) which is associate to the node object with the same name.
The connection between the node object and the tuned object is via the ```MachineConfigLables``` in our case ```worker-cnf``` 

The initial idea which was to modify the same tuned CR that PAO creates, turned out to be a bad idea, since the system not recover from such state.

functests/3_performance_status/status.go
Tal-or: No. Do you want to remove it anyway in order to leave the system in a clean state after the test is falling?

fromanirh: Yes please. We should always do all we can to leave the cluster in a clean/consistent state.

functests/3_performance_status/status.go
Tal-or: I'll comment the above explanation in the code

functests/3_performance_status/status.go
Tal-or: So the same problem apply in [mcp]( https://github.com/openshift-kni/performance-addon-operators/blob/747cbc5b937af9aded6afda395c8359c747db471/functests/3_performance_status/status.go#L94) as well

fromanirh: right. We can just add a TODO here and fix on a separate PR.

functests/3_performance_status/status.go
```diff
@@ -103,6 +115,43 @@ var _ = Describe("Status testing of performance profile", func() {
...
+		It("[test_id:40402] Tuned profile status tied to Performance Profile", func() {
+			// During this test we're creating additional synthetic tuned CR by invoking the createrBadTuned function.
+			// This synthetic tuned will look for a tuned profile which doesn't exist.
+			// This tuned CR will be applied on the profiles.tuned.openshift.io CR (there is such profile per node)
+			// which is associate to the node object with the same name.
+			// The connection between the node object and the tuned object is via the MachineConfigLables, worker-cnf in our case.
+			ns := "openshift-cluster-node-tuning-operator"
+			tunedName := "openshift-cause-tuned-failure"
+
+			// Make sure to clean badTuned object even if the It threw an assertion
```
yanirq, line 128: assertion ?

functests/3_performance_status/status.go
```diff
@@ -103,6 +115,43 @@ var _ = Describe("Status testing of performance profile", func() {
...
+			clean = func() error {
```
yanirq, line 129: I'm wondering if an `eventually` block at the end of the function that does the cleanup would be better here.
having after each means something more inclusive (maybe also for future test)

functests/3_performance_status/status.go
Tal-or: I thought the error being thrown by the ```Eventually``` blocks called assertions. Read that wrong. will fix

functests/3_performance_status/status.go
Tal-or: Not sure I understand what you mean.
You suggest to add ```eventually``` block at the end of the ```clean``` function. 
The ```clean()``` call will still be invoked by the ```AfterEach``` func

yanirq: adding eventually at the end of the test with the clean code

Tal-or: If the test failed in the middle, for example the condition won't be updated properly [here](https://github.com/openshift-kni/performance-addon-operators/pull/598/files#diff-46dd9c2fd680952d6ae2734f9f4b9d42734d386f897dcb79cf030f17eae6f8f4R149), the test will exist without invoking any additional function besides ```AfterEach```
IOW if the ```eventually``` which contains the clean code, is after the line in which the test failed, it won't work.

yanirq: ok . works for me

PR #601, opened 2021-04-13
performance-creator: Don't add the 'userLevelNetworking' field by default

marcel-apf:
Do not automatically add the net.userLevelNetworking field to the performance profile output
if the user does not specify the corresponding user-level-networking param.

Update the regression tests to check for both scenarios (param specified or not).

### DIFF
--- a/cmd/performance-profile-creator/cmd/root.go
+++ b/cmd/performance-profile-creator/cmd/root.go
@@ -32,6 +32,7 @@ import (
 	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeletconfig "k8s.io/kubelet/config/v1beta1"
+	"k8s.io/utils/pointer"
 )
 
 var (
@@ -47,7 +48,7 @@ type ProfileData struct {
 	topologyPoilcy             string
 	rtKernel                   bool
 	additionalKernelArgs       []string
-	userLevelNetworking        bool
+	userLevelNetworking        *bool
 	disableHT                  bool
 }
 
@@ -115,11 +116,6 @@ func getDataFromFlags(cmd *cobra.Command) (ProfileCreatorArgs, error) {
 		return creatorArgs, fmt.Errorf("failed to parse rt-kernel flag: %v", err)
 	}
 
-	userLevelNetworkingEnabled, err := strconv.ParseBool(cmd.Flag("user-level-networking").Value.String())
-	if err != nil {
-		return creatorArgs, fmt.Errorf("failed to parse user-level-networking flag: %v", err)
-	}
-
 	htDisabled, err := strconv.ParseBool(cmd.Flag("disable-ht").Value.String())
 	if err != nil {
 		return creatorArgs, fmt.Errorf("failed to parse disable-ht flag: %v", err)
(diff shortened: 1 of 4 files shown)


testdata/ppc-expected-profiles/profile2.json
```diff
@@ -2,6 +2,6 @@
   "must-gather-dir-path": "must-gather.sno",
   "mcp-name": "master",
   "reserved-cpu-count": 1,
-  "rt-kernel": false
+  "rt-kernel": false,
+  "user-level-networking": false
```
swatisehgal, line 6: I see we have added `user-level-networking` argument here. Just to demonstrate exactly what this PR does, lets add a test-case where user-level-networking is not passed as an argument and we should be able to the expected profile where `userLevelNetworking` field does not exist in the generated performance profile.

swatisehgal (changes requested):
A minor suggestion but other than that the PR looks good to me! Thanks!

testdata/ppc-expected-profiles/profile1.yaml
```diff
@@ -8,8 +8,6 @@ spec:
     reserved: 0,2,40,42
   machineConfigPoolSelector:
     machineconfiguration.openshift.io/role: worker-cnf
-  net:
```
marcel-apf, line 11: @swatisehgal I removed the net field completly the profile1.yaml and since there is no user-level-networking argument in profile1.json it describes the exact scenario you were interested in.

testdata/ppc-expected-profiles/profile2.json
marcel-apf: Thanks for the review, I actually test the exact case you mentioned with the other pair: profile1.{yaml,json}
I removed the "net" field from that profile2.yaml so it can check for it.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -249,13 +253,15 @@ type ProfileCreatorArgs struct {
 	SplitReservedCPUsAcrossNUMA bool   `json:"split-reserved-cpus-across-numa"`
 	DisableHT                   bool   `json:"disable-ht"`
 	RTKernel                    bool   `json:"rt-kernel"`
-	UserLevelNetworking         bool   `json:"user-level-networking"`
+	UserLevelNetworking         *bool  `json:"user-level-networking,omitempty"`
```
fromanirh, line 256: Could you please clarify in the commit message what's the benefit of not adding this option at all, rather than adding with value false?

marcel-apf: Sure, I thought is was clear, we don't want to have:
    kind: PerformanceProfile 
    spec:
    ...
    net:
       userLevelNetworking: false
   ...
in the output profile is the user did not pass --user-level-networking to the ppc cmd line.

marcel-apf: In the commit message:
```    Do not automatically add the net.userLevelNetworking field to the performance profile output
    if the user does not specify the corresponding user-level-networking param.
```

marcel-apf: I'll add "the benefit is to have a more concise profile and the user being able to understand what fields were 'touched'".

marcel-apf: done

fromanirh: thanks!

swatisehgal (approved):
/lgtm
Thanks for the PR!

PR #603, opened 2021-04-13
Bug 1945010: Update CPU Section in PAO Doc

Tal-or:
Rewrite the explanations about Isolated & Reserved CPU features to be more percise and accurate

### DIFF
--- a/api/v1/performanceprofile_types.go
+++ b/api/v1/performanceprofile_types.go
@@ -65,7 +65,8 @@ type CPUSet string
 
 // CPU defines a set of CPU related features.
 type CPU struct {
-	// Reserved defines a set of CPUs that will not be used for any container workloads initiated by kubelet.
+	// Reserved defines a set of CPUs that will be used for interrupts, systemd services, OS and kernel processes.
+	// It can also be used for non guraranteed workloads (burstable and best-effort workloads) such as infrastructure and general workloads.
 	Reserved *CPUSet `json:"reserved,omitempty"`
 	// Isolated defines a set of CPUs that will be used to give to application threads the most execution time possible,
 	// which means removing as many extraneous tasks off a CPU as possible.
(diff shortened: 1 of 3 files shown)


api/v1/performanceprofile_types.go
```diff
@@ -65,14 +65,16 @@ type CPUSet string
 
 // CPU defines a set of CPU related features.
 type CPU struct {
-	// Reserved defines a set of CPUs that will not be used for any container workloads initiated by kubelet.
+	// Reserved defines a set of CPUs that will be used for interrupts, systemd services, OS and kernel processes.
+	// It can also be used for non guraranteed workloads (burstable and best-effort workloads) such as infrastructure and general workloads.
```
fromanirh, line 69: this is the only sentence I'm not convincend, everything else LGTM.
Thing is pods/containers will run on isolated pool, in non-pinned CPUs. Maybe just remove the last sentence ("It can... general workloads.")

Tal-or: But if all isolated pools are taken, new pods/containers can run on reserved right ?

fromanirh: We need to very carefully review and use the terms (myself included). "reserved" means kernel + system daemons + kubelet, aka all CPUS - isolated. "isolated" is actually the set of cpus managed by cpumanager. All QoSes (burstable, guaranteed...) gets CPU from the "isolated" pool. Proper isolation is guaranteed, among the "isolated" pool, only with the static policy and guaranteed QoS will get exclusive CPUs.
TL;DR: the terminolgy is a constant source of confusion, but it can hardly be fixed.

Tal-or: The following line is a quote from the bug description:
_"Reserved field explains that defines a set of CPUs only used for non-container workloads. That's not true since non guaranteed pods can run on reserved ones as well."_
According to your explanation above, this quote is wrong. i.e. non-guaranteed pods **can't** run on reserved CPUs

fromanirh: Sure, this why I mentioned "carefully reviewed" :)
https://github.com/kubernetes/kubernetes/blob/release-1.21/pkg/kubelet/cm/cpumanager/policy_static.go#L35
https://github.com/kubernetes/kubernetes/blob/release-1.21/pkg/kubelet/cm/cpumanager/policy_static.go#L150
https://github.com/kubernetes/kubernetes/blob/release-1.21/pkg/kubelet/cm/cpumanager/policy_static.go#L189

fromanirh (approved):
/lgtm
this looks correct to me and reflects what the system is doing.
We will iterate from there - if needed.

PR #607, opened 2021-04-13
profile-creator: do not consider MCPs targeting unschedulable nodes

marcel-apf:
Refuse to create the profile if all the nodes targeted by the input MCP
are not schedulable. This is the case for the master MCPs in "normal" clusters.


swatisehgal (approved):
/lgtm

PR #615, opened 2021-04-15
Ensure PPC binary is created before running PPC functional tests

swatisehgal:
In case the Performance Profile Creator functional tests are executed
without creating PPC binary, the PPC functional tests would fail. In
order to avoid that, we make it a prerequisite to create the binary
prior to executing the functional tests.


fromanirh:
/lgtm
I think this is a good change in general

PR #617, opened 2021-04-19
Advance go version to 1.16 under go.mod

cynepco3hahue:


fromanirh:
/lgtm
if CI likes it, why not?

PR #622, opened 2021-04-29
Render Command for PAO

Tal-or:
Render command gets performance-profile manifest as an input and based on the provided profile,
it creates manifests yaml of the corresponding CR's which should have been created by the profile.

### DIFF
--- a/controllers/performanceprofile_controller.go
+++ b/controllers/performanceprofile_controller.go
@@ -23,18 +23,16 @@ import (
 
 	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
-	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/kubeletconfig"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/machineconfig"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/manifestset"
 	profileutil "github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/profile"
-	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/runtimeclass"
-	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/tuned"
 	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
 	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
 	corev1 "k8s.io/api/core/v1"
 	nodev1beta1 "k8s.io/api/node/v1beta1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8serros "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -200,7 +198,7 @@ func (r *PerformanceProfileReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
 	instance := &performancev2.PerformanceProfile{}
 	err := r.Get(context.TODO(), req.NamespacedName, instance)
 	if err != nil {
-		if errors.IsNotFound(err) {
+		if k8serros.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
@@ -360,52 +358,36 @@ func (r *PerformanceProfileReconciler) applyComponents(profile *performancev2.Pe
 		return nil, nil
 	}
 
-	// get mutated machine config
-	mc, err := machineconfig.New(r.AssetsDir, profile)
+	components, err := manifestset.GetNewComponents(profile, &r.AssetsDir)
 	if err != nil {
 		return nil, err
 	}
-	if err := controllerutil.SetControllerReference(profile, mc, r.Scheme); err != nil {
-		return nil, err
+	for _, componentObj := range components.ToObjects() {
+		if err := controllerutil.SetControllerReference(profile, componentObj, r.Scheme); err != nil {
+			return nil, err
+		}
 	}
-	mcMutated, err := r.getMutatedMachineConfig(mc)
+
+	// get mutated machine config
+	mcMutated, err := r.getMutatedMachineConfig(components.MachineConfig)
 	if err != nil {
 		return nil, err
 	}
 
 	// get mutated kubelet config
-	kc, err := kubeletconfig.New(profile)
-	if err != nil {
-		return nil, err
-	}
-	if err := controllerutil.SetControllerReference(profile, kc, r.Scheme); err != nil {
-		return nil, err
-	}
-	kcMutated, err := r.getMutatedKubeletConfig(kc)
+	kcMutated, err := r.getMutatedKubeletConfig(components.KubeletConfig)
 	if err != nil {
 		return nil, err
 	}
 
 	// get mutated performance tuned
-	performanceTuned, err := tuned.NewNodePerformance(r.AssetsDir, profile)
-	if err != nil {
-		return nil, err
-	}
-
-	if err := controllerutil.SetControllerReference(profile, performanceTuned, r.Scheme); err != nil {
-		return nil, err
-	}
-	performanceTunedMutated, err := r.getMutatedTuned(performanceTuned)
+	performanceTunedMutated, err := r.getMutatedTuned(components.Tuned)
 	if err != nil {
 		return nil, err
 	}
 
 	// get mutated RuntimeClass
-	runtimeClass := runtimeclass.New(profile, machineconfig.HighPerformanceRuntime)
-	if err := controllerutil.SetControllerReference(profile, runtimeClass, r.Scheme); err != nil {
-		return nil, err
-	}
-	runtimeClassMutated, err := r.getMutatedRuntimeClass(runtimeClass)
+	runtimeClassMutated, err := r.getMutatedRuntimeClass(components.RuntimeClass)
 	if err != nil {
 		return nil, err
 	}
@@ -473,23 +455,23 @@ func (r *PerformanceProfileReconciler) deleteComponents(profile *performancev2.P
 
 func (r *PerformanceProfileReconciler) isComponentsExist(profile *performancev2.PerformanceProfile) bool {
 	tunedName := components.GetComponentName(profile.Name, components.ProfileNamePerformance)
-	if _, err := r.getTuned(tunedName, components.NamespaceNodeTuningOperator); !errors.IsNotFound(err) {
+	if _, err := r.getTuned(tunedName, components.NamespaceNodeTuningOperator); !k8serros.IsNotFound(err) {
 		klog.Infof("Tuned %q custom resource is still exists under the namespace %q", tunedName, components.NamespaceNodeTuningOperator)
 		return true
 	}
 
 	name := components.GetComponentName(profile.Name, components.ComponentNamePrefix)
-	if _, err := r.getKubeletConfig(name); !errors.IsNotFound(err) {
+	if _, err := r.getKubeletConfig(name); !k8serros.IsNotFound(err) {
 		klog.Infof("Kubelet Config %q exists under the cluster", name)
 		return true
 	}
 
-	if _, err := r.getRuntimeClass(name); !errors.IsNotFound(err) {
+	if _, err := r.getRuntimeClass(name); !k8serros.IsNotFound(err) {
 		klog.Infof("Runtime class %q exists under the cluster", name)
 		return true
 	}
 
-	if _, err := r.getMachineConfig(machineconfig.GetMachineConfigName(profile)); !errors.IsNotFound(err) {
+	if _, err := r.getMachineConfig(machineconfig.GetMachineConfigName(profile)); !k8serros.IsNotFound(err) {
 		klog.Infof("Machine Config %q exists under the cluster", name)
 		return true
 	}
--- a/go.mod
+++ b/go.mod
@@ -32,6 +32,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.6.0
 	github.com/spf13/cobra v1.1.1
+	github.com/spf13/pflag v1.0.5
 	go.mongodb.org/mongo-driver v1.3.2 // indirect
 	go.uber.org/zap v1.14.1 // indirect
 	golang.org/x/lint v0.0.0-20201208152925-83fdc39ff7b5 // indirect
--- a/main.go
+++ b/main.go
@@ -27,6 +27,7 @@ import (
 	"github.com/openshift-kni/performance-addon-operators/controllers"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
 	"github.com/openshift-kni/performance-addon-operators/version"
+	"github.com/spf13/cobra"
 
 	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
 	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
@@ -37,6 +38,7 @@ import (
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	"github.com/openshift-kni/performance-addon-operators/pkg/cmd/render"
 	apiruntime "k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -83,11 +85,33 @@ func printVersion() {
 }
 
 func main() {
+	// Add klog flags
+	klog.InitFlags(nil)
+
+	command := newRootCommand()
+	if err := command.Execute(); err != nil {
+		klog.Errorf("%v", err)
+	}
+}
+
+func newRootCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "performance-addon-operator",
+		Short: "OpenShift performance addon operator",
+		Run: func(cmd *cobra.Command, args []string) {
+			// if no subcommand just run the usual
+			runPAO()
+		},
+	}
+
+	cmd.AddCommand(render.NewRenderCommand())
+	return cmd
+}
+
+func runPAO() {
 	var metricsAddr string
 	var enableLeaderElection bool
 
-	// Add klog flags
-	klog.InitFlags(nil)
 	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
 
 	flag.StringVar(&metricsAddr, "metrics-addr", fmt.Sprintf("%s:%d", metricsHost, metricsPort),
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -1,0 +1,165 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package render
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/manifestset"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+)
+
+type renderOpts struct {
+	performanceProfileInputFiles performanceProfileFiles
+	assetsInDir                  string
+	assetsOutDir                 string
+}
+
+type performanceProfileFiles []string
+
+func (ppf *performanceProfileFiles) String() string {
+	return fmt.Sprint(*ppf)
+}
+
+func (ppf *performanceProfileFiles) Type() string {
+	return "performanceProfileFiles"
+}
+
+// Set parse performance-profile-input-files flag and store it in ppf
+func (ppf *performanceProfileFiles) Set(value string) error {
+	if len(*ppf) > 0 {
+		return errors.New("performance-profile-input-files flag already set")
+	}
+
+	for _, s := range strings.Split(value, ",") {
+		*ppf = append(*ppf, s)
+	}
+	return nil
+}
+
+//NewRenderCommand creates a render command.
+func NewRenderCommand() *cobra.Command {
+	renderOpts := renderOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "render",
+		Short: "Render performance-addon-operator manifests",
+		Run: func(cmd *cobra.Command, args []string) {
+
+			if err := renderOpts.Validate(); err != nil {
+				klog.Fatal(err)
+			}
+
+			if err := renderOpts.Run(); err != nil {
+				klog.Fatal(err)
+			}
+		},
+	}
+
+	renderOpts.AddFlags(cmd.Flags())
+
+	return cmd
+}
+
+func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.Var(&r.performanceProfileInputFiles, "performance-profile-input-files", "A comma-separated list of performance-profile manifests.")
+	fs.StringVar(&r.assetsInDir, "asset-input-dir", components.AssetsDir, "Input path for the assets directory.")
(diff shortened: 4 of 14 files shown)


cmd/render/render.go
```diff
@@ -0,0 +1,128 @@
...
+
+type manifestable map[string]interface{}
+
+//NewRenderCommand creates a render command.
+func NewRenderCommand() *cobra.Command {
+	renderOpts := renderOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "render",
+		Short: "Render performance-addon-operator CR manifests",
```
yanirq, line 48: "Render performance-addon-operator CR manifests"   Maybe drop "CR"  its a bit confusing, manifests should be enough. If you want to elaborate you can write exactly what types of manifests are created

cmd/render/render.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	renderOpts.AddFlags(cmd.Flags())
+
+	return cmd
+}
+
+func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&r.ppInputFilePath, "performance-profile-file-path", r.ppInputFilePath, "Input path for the Performance Profile manifest.")
+	fs.StringVar(&r.assetsDir, "assets-dir-path", components.AssetsDir, "Input path for the assets directory.")
```
yanirq, line 68: assets  or generated assets ?

cmd/render/render.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/ghodss/yaml"
+	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/kubeletconfig"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/machineconfig"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/runtimeclass"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/tuned"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"k8s.io/klog"
```
yanirq, line 31: I'm starting to see a lot  log references in different files,  not use its too much of a difference but maybe we should stay consistent  (log "github.com/sirupsen/logrus" for example is some references)   
@cynepco3hahue  what's your opinion here ?

cmd/render/render.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&r.ppInputFilePath, "performance-profile-file-path", r.ppInputFilePath, "Input path for the Performance Profile manifest.")
+	fs.StringVar(&r.assetsDir, "assets-dir-path", components.AssetsDir, "Input path for the assets directory.")
+	fs.StringVar(&r.outputDir, "output-dir", r.outputDir, "Output path for the rendered manifests.")
+}
+
+func (r *renderOpts) Validate() error {
+	if len(r.ppInputFilePath) == 0 || len(r.outputDir) == 0 {
```
yanirq, line 73: why not split the validation here ? in case one of them is specified ?  maybe its more user friendly

cmd/render/render.go
```diff
@@ -0,0 +1,128 @@
...
+*/
+
+package render
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"github.com/ghodss/yaml"
+	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
```
yanirq, line 23: maybe not for now, but we need to think what happens when we bump the performance profile version (if we add new apis)

yanirq:
before jumping to e2e tests you can also provide unittests that checks generate examples  (compare using strings or regular expressions)

cmd/render/render.go
Tal-or: This [asset](https://github.com/openshift-kni/performance-addon-operators/tree/master/build/assets) directory.
I'm not sure myself, but it's necessary for the components creation.

fromanirh: this should be renamed as `asset-input-dir`

cmd/render/render.go
```diff
@@ -0,0 +1,128 @@
...
+func NewRenderCommand() *cobra.Command {
+	renderOpts := renderOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "render",
+		Short: "Render performance-addon-operator CR manifests",
+		Run: func(cmd *cobra.Command, args []string) {
+
+			if err := renderOpts.Validate(); err != nil {
+				klog.Error(err)
```
fromanirh, line 52: why not Fatal here?

cmd/render/render.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+	}
+
+	renderOpts.AddFlags(cmd.Flags())
+
+	return cmd
+}
+
+func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&r.ppInputFilePath, "performance-profile-file-path", r.ppInputFilePath, "Input path for the Performance Profile manifest.")
```
fromanirh, line 67: should this be part of `asset-input-dir` (e.g. read the file from there, fail if it is not inside)?

cmd/render/render.go
```diff
@@ -0,0 +1,128 @@
...
+	fs.StringVar(&r.assetsDir, "assets-dir-path", components.AssetsDir, "Input path for the assets directory.")
+	fs.StringVar(&r.outputDir, "output-dir", r.outputDir, "Output path for the rendered manifests.")
```
fromanirh, line 69: this should be renamed as `asset-output-dir`

cmd/render/render.go
fromanirh: +1 validate each field and return a separate error for each mandatory but missing field

cmd/render/render.go
```diff
@@ -0,0 +1,128 @@
...
+	performanceTuned, err := tuned.NewNodePerformance(r.assetsDir, profile)
+	if err != nil {
+		return err
+	}
+	components[performanceTuned.Kind] = performanceTuned
+
+	runtimeClass := runtimeclass.New(profile, machineconfig.HighPerformanceRuntime)
+	components[runtimeClass.Kind] = runtimeClass
+
+	for name, component := range components {
```
fromanirh, line 114: not sure the kind is a great fit for the name

cmd/render/render.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (r *renderOpts) Validate() error {
+	if len(r.ppInputFilePath) == 0 || len(r.outputDir) == 0 {
+		return fmt.Errorf("performance-profile-file-path and output-dir must be specified")
+	}
+	return nil
+}
+
+func (r *renderOpts) Run() error {
```
fromanirh, line 79: the problem here is we have a fully separate path wrt the online one. Say tomorrow we manage a new object, we will need to remember to add it here. How we can more likely we remember to do this?

main.go
```diff
@@ -83,11 +85,33 @@ func printVersion() {
 }
 
 func main() {
+	// Add klog flags
+	klog.InitFlags(nil)
+
+	command := newPAOCommand()
```
fromanirh, line 91: I'd call this `newRootCommand` for consistency with `cobra` usage patterns

cmd/render/render.go
Tal-or: The manifests also known as assets?

cmd/render/render.go
Tal-or: Changed that on purpose, but I misunderstood the way it works.
Will replace back to Fatal

cmd/render/render.go
Tal-or: Then we'll have to force the ```performance-profile.yaml``` to be on the same directory together with the other assets.
Is that something we want?

cmd/render/render.go
cynepco3hahue: In general, I prefer `"k8s.io/klog"` because it used by k8s, also I can find only a few places where `log "github.com/sirupsen/logrus"` used.

cmd/render/render.go
Tal-or: The names of the manifests are ```Kind.yaml``` so maybe instead of name call it ```fileName``` ?

cmd/render/render.go
fromanirh: manifests are the assets PAO produces, and are assets in general

cmd/render/render.go
fromanirh: let's keep like this for the time being and re-evaluate later.

cmd/render/render.go
Tal-or: As for now we have no record for the possible objects that we're creating. it is possible to create a central place like pool/list/file which contains all the objects we have and keep it always updated, and both online/offline paths will follow it (no sure about the implementation details and maybe it deserve a separate PR).
But the same question can also asked here, what if forget to update this central place?

cmd/render/render.go
Tal-or: Got it. Done

cmd/render/render.go
fromanirh: indeed the problem is we have two completely separate flows :)

fromanirh: for the future:  I think we should evaluate a refactoring/code extraction from  controllers/performanceprofile_controller.go:applyComponents

vendor/modules.txt
```diff
@@ -132,7 +132,8 @@ github.com/google/go-cmp/cmp/internal/function
 github.com/google/go-cmp/cmp/internal/value
```
fromanirh, line 132: Shouldn't this commit also include the updates to `go.mod` and `go.sum` ?

deploy/olm-catalog/performance-addon-operator/4.8.0/performance-addon-operator.v4.8.0.clusterserviceversion.yaml
```diff
@@ -119,7 +119,8 @@ metadata:
       ]
     capabilities: Basic Install
```
fromanirh, line 120: please ignore changes on clusterserviceversion file and merge changes to `go.mod` and `go.sum` in the previous commit.

main.go
```diff
@@ -37,6 +38,7 @@ import (
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	render "github.com/openshift-kni/performance-addon-operators/pkg/cmd/render"
```
fromanirh, line 41: the rename is unneeded, but no big deal. Is this done automatically by the tooling?

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,196 @@
...
+	"k8s.io/klog"
+)
+
+type renderOpts struct {
+	ppInputFileDir string
+	assetsInDir    string
+	assetsOutDir   string
+}
+
+type manifestable map[string]v1.Object
```
fromanirh, line 46: better use `manifestTable` (camelCase)

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,196 @@
...
+
+//NewRenderCommand creates a render command.
+func NewRenderCommand() *cobra.Command {
+	renderOpts := renderOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "render",
+		Short: "Render performance-addon-operator manifests",
+		Long: `Render performance-addon-operator manifests creates the following manifests:
```
fromanirh, line 55: This gets obsolete fast. Better to either autogenerate the list of output manifests or just be generic and tell "creates the output manifests" or so.

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,196 @@
...
+		},
+	}
+
+	renderOpts.AddFlags(cmd.Flags())
+
+	return cmd
+}
+
+func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&r.ppInputFileDir, "performance-profile-dir-path", r.ppInputFileDir, "Input path for the Performance Profile manifest's directory.")
```
fromanirh, line 78: do we need a separate directory? How other operators do?

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,196 @@
...
+	fs.StringVar(&r.assetsInDir, "asset-input-dir", components.AssetsDir, "Input path for the assets directory.")
+	fs.StringVar(&r.assetsOutDir, "asset-output-dir", r.assetsOutDir, "Output path for the rendered manifests.")
+	// environmet variables has precedence over standard input
+	r.readFlagsFromEnv()
+}
+
+func (r *renderOpts) readFlagsFromEnv() {
+	if ppDir := os.Getenv("PP_DIR"); len(ppDir) > 0 {
```
fromanirh, line 86: too terse: let's be consistent with the long options. So, if the long options is
```
--something-like-this
```
then the env variable should be
```
SOMETHING_LIKE_THIS
```

Besides, is this something the other operators do?

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,196 @@
...
+	return nil
+}
+
+func (r *renderOpts) Run() error {
+	PPFiles := getPPList(r.ppInputFileDir)
+	if len(PPFiles) == 0 {
+		return fmt.Errorf("no valid performance-profiles found under: %v", r.ppInputFileDir)
+	}
+
+	for idx, pp := range PPFiles {
```
fromanirh, line 117: this looks similar but different wrt to `applyComponents`. We should try harder NOT to have two similar but distinct flows. Can we factor out the two flows in a common path?

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,196 @@
...
+
+// getPPList return a list of performance-profile manifests under dirPath
+func getPPList(dirPath string) []string {
+	files, err := ioutil.ReadDir(dirPath)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// some regex to look for performance-profile manifests in directory's path
+	reg, err := regexp.Compile(`[P,p].*[-,_]*[P,p].*\.y[a]*ml`)
```
fromanirh, line 183: why is this needed?

pkg/cmd/render/render.go
Tal-or: During our engineering meeting we agreed to add the env variables option.
I'll change to the naming convention you just suggested

pkg/cmd/render/render.go
Tal-or: I thought we agreed to dedicate a separate PR for this, but I can work on that np.

pkg/cmd/render/render.go
Tal-or: I don't know the performance profile name and there might be other files in the directory. so i'm looking for yaml's file that matches this regex

pkg/cmd/render/render.go
fromanirh: Sure, did you check what other operators consume the env vars?

pkg/cmd/render/render.go
fromanirh: I don't recall, it could very much be :) but if so that PR should come right after this one, we should not carry this tech debt for too long.

main.go
Tal-or: Good question i have no idea. maybe it's a copy-paste mistake

pkg/cmd/render/render.go
fromanirh: How others operator detect the input file? Do they have a separate command line flag?
Another option could be just decode all the YAMLs in the input directory, and look for a performance profile. I actually like this more than looking for specific names.

pkg/cmd/render/render.go
fromanirh: I can think few things, but I believe is not worth (yet). I'll just go for the more generic phrasing.

pkg/cmd/render/render.go
Tal-or: they provided separate directories if needed. see [here](https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L151)
Besides that, I don't think we need to force the user to have the performance-profile manifests together with other input-assets. This is more flexible don't you think?

fromanirh: Looking at the example you provided, I see
```
		/usr/bin/cluster-config-operator render \
		--cluster-infrastructure-input-file=/assets/manifests/cluster-infrastructure-02-config.yml \
                ....
```
It seems to me a bit cleaner to just add something like
```
fs.StringVar(&r.performanceProfileInputFile, "performance-profile-input-file", r.performanceProfileInputFile, "Input path for the Performance Profile manifest.")
```
This way we get immediatley what we want, and we don't need anymore extra detection steps, e.g. with regexes.

Tal-or: This was the original implementation, but we agreed to have the ability to render multiple pp files. 
Maybe I can provide a list of files` names,  instead of single file name or directory name

fromanirh: Right. Having the option accept more than a single file (or having the operator accept the same option multiple times) seems the best option.

pkg/cmd/render/render.go
Tal-or: IIRC other operator don't. But we wanted this option so it will be more natural when running render inside a container

main.go
fromanirh: ok, no big deal anyway.

pkg/cmd/render/render.go
fromanirh: ok no big deal if the naming is consistent.

pkg/cmd/render/render.go
Tal-or: How we can implement the same for env variable?
Is something like that:
```PERFORMANCE_PROFILE_INPUT_FILES=<file1's path>,<file2's path>,<file3's path>```
or ```PERFORMANCE_PROFILE_INPUT_FILES=<file1's path>:<file2's path>:<file3's path>```
is okay?

Comment (fromanirh) [reply pkg/cmd/render/render.go:78]: yep, fine. Comma separated is a bit more common (while colon ":" is more uncommon) so I'd go for that.

File: controllers/performanceprofile_controller.go [hunk 21]
```diff
@@ -18,16 +18,17 @@ package controllers
 
 import (
 	"context"
+	e "errors"
```
fromanirh, line 21: well, this seems a bit too much

controllers/performanceprofile_controller.go
```diff
@@ -18,16 +18,17 @@ package controllers
...
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/kubeletconfig"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/runtimeclass"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/tuned"
```
fromanirh, line 24: please move back these imports below, because it breaks proper grouping:

1. stdlib
2. third party
3. our own modules

controllers/performanceprofile_controller.go
```diff
@@ -540,3 +542,34 @@ func hasMatchingLabels(performanceprofile *performancev2.PerformanceProfile, mcp
 	}
 	return true
 }
+
+// CompList a map that contains Kind string as key and component's instances as value
+type CompList map[string]interface{}
```
fromanirh, line 547: Well, I'd use a plain old struct like
```
type ManifestResultSet struct {
       MachineConfig *machineconfigv1.MachineConfig
       KubeletConfig *machineconfigv1.KubeletConfig
       Tuned         *tunedv1.Tuned
       RuntimeClass  *nodev1beta1.RuntimeClass
}
```
Not like we add new outputs very day (or even every major...)
This saves type assertions and looks a bit neater.

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,162 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package render
+
+import (
+	"errors"
+	"fmt"
+	"github.com/openshift-kni/performance-addon-operators/controllers"
```
fromanirh, line 21: please move this in the group below

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,162 @@
...
+	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+)
+
+type renderOpts struct {
+	ppInputFiles ppFiles
```
fromanirh, line 37: let's use `performanceProfileInputFiles` for consistency

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,162 @@
...
+	assetsInDir  string
+	assetsOutDir string
+}
+
+type ppFiles []string
```
fromanirh, line 42: let's use `performanceProfileFiles` for consistency

controllers/performanceprofile_controller.go
Tal-or: ```errors``` and ```err``` are taken

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,162 @@
...
+			return err
+		}
+		or := []v1.OwnerReference{
+			{
+				Kind: profile.Kind,
+				Name: profile.Name,
+			},
+		}
+
+		for name, component := range components {
```
fromanirh, line 146: if we want to keep the generic range, I'd add a method like
 ```
func (ms ManifestResultSet) ToObjects() []v1.Object
```

**Review by fromanirh — COMMENTED** [review]
thanks for the updates, getting there

File: controllers/performanceprofile_controller.go (hunk 21 above)
Comment (fromanirh) [reply controllers/performanceprofile_controller.go:21]: keep this "errors"...

File: controllers/performanceprofile_controller.go [hunk 28]
```diff
@@ -18,16 +18,17 @@ package controllers
...
 
 	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
-	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/kubeletconfig"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/machineconfig"
 	profileutil "github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/profile"
-	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/runtimeclass"
-	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/tuned"
 	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
 	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
```
fromanirh, line 34: ...and rename the one below like `apierrors` or `k8serrors`

controllers/performanceprofile_controller.go
Tal-or: But then we won't be able to iterate the struct, when most of the action are common and can be done using loops.

fromanirh: see below about `ToObjects`. Lacking generics, lacking better ideas (which admittledly I don't have now) we prefer quasi-duplicate, type-specific code to scattered type assertions.

Tal-or: yet you'll need a way to iterate the struct inside the ```ToObject``` in order to create this ```[]v1.Object``` but I see your point. type assertion might be confusing and less readable

fromanirh: In `ToObject` we should convert each field explicitely (no generic iteration). The idea is to keep at least all the conversion in a single place and let not them spread around in the codebase.

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,162 @@
...
+			component.(v1.Object).SetOwnerReferences(or)
+			b, err := yaml.Marshal(component)
```
Tal-or, line 148: @fromanirh  Any idea how to marshal here in a loop? converting to object isn't enough

fromanirh: I will have a closer look later

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/go-openapi/strfmt v0.19.5 // indirect
 	github.com/go-openapi/swag v0.19.9 // indirect
 	github.com/go-openapi/validate v0.19.8 // indirect
+	github.com/google/uuid v1.2.0 // indirect
```
fromanirh, line 21: these aren't really autogenerated changes, but rather deps bump/update

controllers/performanceprofile_controller.go
```diff
@@ -360,52 +358,36 @@ func (r *PerformanceProfileReconciler) applyComponents(profile *performancev2.Pe
...
 
-	// get mutated machine config
-	mc, err := machineconfig.New(r.AssetsDir, profile)
+	components, err := manifestset.GetNewComponents(profile, &r.AssetsDir)
 	if err != nil {
 		return nil, err
 	}
-	if err := controllerutil.SetControllerReference(profile, mc, r.Scheme); err != nil {
-		return nil, err
+	for _, componentObj := range components.ToObject() {
```
fromanirh, line 365: nit: should be `ToObjects` (plural)

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,165 @@
...
+//NewRenderCommand creates a render command.
+func NewRenderCommand() *cobra.Command {
+	renderOpts := renderOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "render",
+		Short: "Render performance-addon-operator manifests",
+		Run: func(cmd *cobra.Command, args []string) {
+
+			if err := renderOpts.Validate(); err != nil {
```
fromanirh, line 73: I don't think this (and Run) should not be a method of `renderOpts`, but we can surely live with it

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,165 @@
...
+			componentObj.SetOwnerReferences(or)
+		}
+
+		for name, manifest := range components.ToManifestTable() {
+			b, err := yaml.Marshal(manifest)
+			if err != nil {
+				return err
+			}
+
+			fileName := fmt.Sprintf("%s-%d.yaml", name, idx)
```
fromanirh, line 156: ok, this is the biggest and last point. I believe the output name should be like `$INPUT_NAME_tuned.yaml` `$INPUT_NAME_kubeletconfig.yaml` and so forth. IOW, use kind (`in_this_format`, `NotLikeThis`) and not a index. In case of ambiguity, we shoudl differentiate with object name if possible

fromanirh (commented):
let's fix the output names and we're good to go.

fromanirh (commented):
/lgtm
thanks for the updates

PR #623, opened 2021-04-29
[WIP] profile-creator: Add --info arg to PPC

marcel-apf:
When the flag is set, only the must-gather argument is expected, all the other arguments are optional.
Show the cluster information and exit in this case.


fromanirh:
we merged https://github.com/openshift-kni/performance-addon-operators/pull/628 - which includes commits from this PR.

PR #624, opened 2021-05-05
[release 4.7] Enabled stalld only from a minimum kernel version

yanirq:

The minimum kernel versions required for stalld if in oc 4.7 are
4.18.0-240.22.1.el8_3.x86_64
4.18.0-240.22.1.rt7.77.el8_3.x86_64

The enablement is determined by using tuned uname_regex option to check
if the kernel version prefix matches the minimum requirement

oc 4.7 kernel version starts from 4.18.0-240.10.1.el8_3.x86_64

### DIFF
--- a/build/assets/tuned/openshift-node-performance
+++ b/build/assets/tuned/openshift-node-performance
@@ -18,6 +18,9 @@ isolated_cores={{.IsolatedCpus}}
 {{end}}
 
 not_isolated_cores_expanded=${f:cpulist_invert:${isolated_cores_expanded}}
+# minimum kernel version applicable for enabling stalld is 4.18.0-240.22.1.el8_3.x86_64 or 4.18.0-240.22.1.rt7.77.el8_3.x86_64
+# starting kernel version for oc 4.7 is 4.18.0-240.10.1.el8_3.x86_64
+stalld_minimum_kernel_version_prefix=^Linux [^ ]+ (?!4\.18\.0-240\.([0-9]|1[0-9]|2[0-1])\.)
 
 [cpu]
 force_latency=cstate.id:1|3                   #  latency-performance  (override)
@@ -27,6 +30,7 @@ min_perf_pct=100                              #  latency-performance
 
 [service]
 service.stalld=start,enable
+uname_regex=${stalld_minimum_kernel_version_prefix}
 
 [vm]
 transparent_hugepages=never                   #  network-latency
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -204,8 +204,16 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				Expect(err).ToNot(HaveOccurred(), "failed to execute uname")
 				currentKernelPrefix := strings.Split(kernel, ".")
 				currentSubVersion, err := strconv.Atoi(currentKernelPrefix[3])
-				Expect(err).ToNot(HaveOccurred())
-				if currentSubVersion >= minSubVersion {
+				// check if current kernel is a new rt batch stream (higher than 0-240)
+				if err != nil && strings.Contains(currentKernelPrefix[3], "rt") {
+					batchVersionStr := strings.Split(currentKernelPrefix[2], "-")
+					batchVersion, err := strconv.Atoi(batchVersionStr[1])
+					Expect(err).ToNot(HaveOccurred())
+					Expect(batchVersion > 240).Should(BeTrue(), fmt.Sprintf("Current kernel version %s is incompatible to the current oc release", kernel))
+					tuned := tunedForNode(&node)
+					_, err = pods.ExecCommandOnPod(tuned, []string{"pidof", "stalld"})
+					Expect(err).ToNot(HaveOccurred())
+				} else if currentSubVersion >= minSubVersion {
 					tuned := tunedForNode(&node)
 					_, err := pods.ExecCommandOnPod(tuned, []string{"pidof", "stalld"})
 					Expect(err).ToNot(HaveOccurred())
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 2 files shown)


build/assets/tuned/openshift-node-performance
```diff
@@ -18,6 +18,9 @@ isolated_cores={{.IsolatedCpus}}
 {{end}}
 
 not_isolated_cores_expanded=${f:cpulist_invert:${isolated_cores_expanded}}
+# minimum kernel version applicable for enabling stalld is 4.18.0-240.22.1.el8_3.x86_64 or 4.18.0-240.22.1.rt7.77.el8_3.x86_64
+# starting kernel version for oc 4.7 is 4.18.0-240.10.1.el8_3.x86_64
+stalld_minimum_kernel_version_prefix=^Linux [^ ]+ (?!4\.18\.0-240\.([0-9]|1[0-9]|2[0-1])\.)
```
fromanirh, line 23: I wonder if there is a way to test this regex besides the minimal coverage the e2e tests give us

yanirq: Yes, but it should be under tuned in my opinion

fromanirh: Surely the feature itself needs to be tested in tuned, but the fact that this regex captures properly the range of versions we care about seem to belong here more than in tuned.

MarSik: We should definitely have a test in PAO that checks what we enable / not enable based on kernel version

MarSik: We already have one for starting stalld, maybe we can add another one for not-starting it.. I just wonder how to inject different uname

yanirq: we might be able to install an older kernel version , then reinstall the current one but that could really tamper the test environment and take a lot of time.  
another option might be adding a test suit for overriding tuned yamls as we do for hotfixes and add a different regex that will cause stalld to be disabled but that won't be a direct test to what we have under the default tuned base profile.

MarSik: Can we trick tuned to think the uname is different? By replacing the uname binary or something?

fromanirh: AFAIU tuned just calls `os.uname()` which is a cpython builtin. Maybe worth checking with tuned folks about the cleanest way to run this check.

fromanirh: *MAYBE* something like https://github.com/fromanirh/fromanirh/tree/main/tools/fake-uname ?

MarSik: Yeah, I was thinking about ld preload as well, but all our tests are part of the same binary, right? So we would need a way to change it for separate tests without restarting the binary.

fromanirh: Well a preload shlib can be made smarter - it's a testing tool after all, it will never see production - like reading a file each time, but we should get in touch with tuned folks and see if a code change on their side can make the testing easier. OTOH we'll have a tech debt until we can really test this flow.

fromanirh: Or maybe we can just (ab)use the tuned `uname_string` configuration option:
https://github.com/redhat-performance/tuned/blob/master/tuned-main.conf#L46

functests/1_performance/performance.go
```diff
@@ -211,6 +212,29 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+		It("[test_id:XXXX][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] stalld daemon is not running on the host", func() {
+			By("Running on a kernel version lower than 4.18.0-240.22.*")
+			invalidStalldKernelVersion := "4.18.0-240.21.1.el8_3.x86_64"
+			randomWorkerNode := workerRTNodes[rand.Intn(len(workerRTNodes))]
+			tuned := tunedForNode(&randomWorkerNode)
+			customUnameCmd := []string{"sed", "-i", fmt.Sprintf("'$ a uname_string = %s'", invalidStalldKernelVersion), "/etc/tuned/tuned-main.conf"}
+			_, err := pods.ExecCommandOnPod(tuned, customUnameCmd)
+			Expect(err).ToNot(HaveOccurred())
+			// kill tuned process will cause it to spin up again and reload the profile
+			killTunedCmd := []string{"pkill", "'^tuned'"}
```
fromanirh, line 225: please document with a code comment why not a clean restart/config reload

functests/1_performance/performance.go
```diff
@@ -211,6 +212,29 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				}
 			}
 		})
+
+		It("[test_id:XXXX][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] stalld daemon is not running on the host", func() {
+			By("Running on a kernel version lower than 4.18.0-240.22.*")
+			invalidStalldKernelVersion := "4.18.0-240.21.1.el8_3.x86_64"
+			randomWorkerNode := workerRTNodes[rand.Intn(len(workerRTNodes))]
+			tuned := tunedForNode(&randomWorkerNode)
+			customUnameCmd := []string{"sed", "-i", fmt.Sprintf("'$ a uname_string = %s'", invalidStalldKernelVersion), "/etc/tuned/tuned-main.conf"}
```
fromanirh, line 221: if memory serves well this just appends the `uname_string = ...` to the file, right? If so could a simpler `echo "uname_string = ..." >> /etc/tuned/tuned-main.conf` work? (or anything equivalent)

fromanirh (commented):
thanks for adding the e2e test. Looks good, minor comments inside

yanirq:
@jmencak  is it possible that uname_regex is ignored  by the mechanism invoked by NTO to enable stalld ?

jmencak:
> @jmencak is it possible that uname_regex is ignored by the mechanism invoked by NTO to enable stalld ?

NTO does not re-implement all the functionality `TuneD` provides, so what you're trying to do will not work until https://github.com/openshift/cluster-node-tuning-operator/pull/226 is merged for 4.8 hopefully.  I've checked last week and CoreOS still shipped old `1.6` stalld.

yanirq:
/hold  we cannot use this functionality until we switch to os shipped stalld .
once https://github.com/openshift/cluster-node-tuning-operator/pull/226 gets into 4.7 we can use this PR

jmencak:
> /hold we cannot use this functionality until we switch to os shipped stalld .
> once [openshift/cluster-node-tuning-operator#226](https://github.com/openshift/cluster-node-tuning-operator/pull/226) gets into 4.7 we can use this PR

This is not going to happen util we see RHCOS 8.4 content in OCP 4.7.  At the moment we still have RHCOS 8.3.

MarSik:
> > @jmencak is it possible that uname_regex is ignored by the mechanism invoked by NTO to enable stalld ?

Well NTO can ignore it when copying in the binary and the service file, but the service is still started by tuned, isn't it?

jmencak:
> > > @jmencak is it possible that uname_regex is ignored by the mechanism invoked by NTO to enable stalld ?
> 
> Well NTO can ignore it when copying in the binary and the service file, but the service is still started by tuned, isn't it?

Yes, the service is started by `TuneD`, but the problem that NTO "ignores"/doesn't_parse uname_regex means it will always be started via ignition, which is something you probably don't want when you have older kernel releases, right?

PR #628, opened 2021-05-14
profile-creator: add discovery mode

fromanirh:
continuation of https://github.com/openshift-kni/performance-addon-operators/pull/623


cmd/performance-profile-creator/cmd/root.go
```diff
@@ -55,45 +55,90 @@ type ProfileData struct {
...
+	}
+
+	var requiredFlags []string
+
+	root := &cobra.Command{
+		Use:   "performance-profile-creator",
+		Short: "A tool that automates creation of Performance Profiles",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !cmd.Flag("must-gather-dir-path").Changed {
+				return fmt.Errorf("missing cluster data")
```
swatisehgal, line 78: Lets make this more verbose and let the user know that the ` --must-gather-dir-path` must be specified.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -55,45 +55,90 @@ type ProfileData struct {
...
+	root.PersistentFlags().StringVar(&pcArgs.PowerConsumptionMode, "power-consumption-mode", profilecreator.ValidPowerConsumptionModes[0], fmt.Sprintf("The power consumption mode.  [Valid values: %s]", strings.Join(profilecreator.ValidPowerConsumptionModes, ", ")))
+	root.PersistentFlags().StringVar(&pcArgs.MustGatherDirPath, "must-gather-dir-path", "must-gather", "Must gather directory path")
+	root.PersistentFlags().StringVar(&pcArgs.ProfileName, "profile-name", "performance", "Name of the performance profile to be created")
+	root.PersistentFlags().StringVar(&pcArgs.TMPolicy, "topology-manager-policy", kubeletconfig.RestrictedTopologyManagerPolicy, fmt.Sprintf("Kubelet Topology Manager Policy of the performance profile to be created. [Valid values: %s, %s, %s]", kubeletconfig.SingleNumaNodeTopologyManagerPolicy, kubeletconfig.BestEffortTopologyManagerPolicy, kubeletconfig.RestrictedTopologyManagerPolicy))
+	root.PersistentFlags().BoolVar(&pcArgs.Info, "info", false, "Show cluster information, disregard the other arguments")
+
+	requiredFlags = append(requiredFlags,
```
swatisehgal, line 125: Why haven't we initialized this at line 71 itself (at the time of their declaration) rather than appending it here?
```

var requiredFlags: []string{"reserved-cpu-count",
		"mcp-name",
		"rt-kernel",
		"must-gather-dir-path",
}
```

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -55,45 +55,90 @@ type ProfileData struct {
...
+			if err != nil {
+				return fmt.Errorf("failed to parse the info flag: %v", err)
+			}
+			if info {
+				showClusterData(cluster)
+				return nil
+			}
+
+			if !haveRequiredFlags(cmd, requiredFlags...) {
+				return fmt.Errorf("missing required flags: %s", strings.Join(requiredFlags, ", "))
```
swatisehgal, line 97: This will return all the `requiredFlags` even if the user specifies a subset of them. We should either improve the error messge and say "please specify all the required flags:" or indicate the missing flag that the user did not specify.

swatisehgal (changes requested):
Thanks for a PR. Have a few suggestions/questions.

cmd/performance-profile-creator/cmd/root.go
fromanirh: Ack, done.

cmd/performance-profile-creator/cmd/root.go
fromanirh: I though about it and preferred the form I posted because this way the addition to the required set is closer to the flag definition. But the function is short enough that we can do this in one go like you suggested. Changed.

cmd/performance-profile-creator/cmd/root.go
fromanirh: good idea, done.

fromanirh:
/hold
I want to do some extra testing (and maybe add some e2e tests)

swatisehgal:
Just thinking out loud, maybe we should add some text to capture the `info` mode  in the [README](https://github.com/openshift-kni/performance-addon-operators/blob/master/cmd/performance-profile-creator/README.md) as well.

fromanirh:
> Just thinking out loud, maybe we should add some text to capture the `info` mode in the [README](https://github.com/openshift-kni/performance-addon-operators/blob/master/cmd/performance-profile-creator/README.md) as well.

yeah, why not? Will be adding in the next upload.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -311,6 +351,7 @@ func init() {
 	rootCmd.MarkPersistentFlagRequired("must-gather-dir-path")
 	rootCmd.PersistentFlags().StringVar(&args.ProfileName, "profile-name", "performance", "Name of the performance profile to be created")
 	rootCmd.PersistentFlags().StringVar(&args.TMPolicy, "topology-manager-policy", kubeletconfig.RestrictedTopologyManagerPolicy, fmt.Sprintf("Kubelet Topology Manager Policy of the performance profile to be created. [Valid values: %s, %s, %s]", kubeletconfig.SingleNumaNodeTopologyManagerPolicy, kubeletconfig.BestEffortTopologyManagerPolicy, kubeletconfig.RestrictedTopologyManagerPolicy))
+	rootCmd.PersistentFlags().BoolVar(&args.Info, "info", false, "Show cluster information, disregard the other arguments")
```
swatisehgal: Just one more comment that I missed in the review I submitted.
I think we should improve the message returned here. Instead of `disregard the other arguments` maybe we should say must specify  `--must-gather-dir-path`.

That makes me think of the case where the user provides arguments in addition to `--must-gather-dir-path`. Should we explicitly let the user know that other args are not needed or work as we do i.e. ignore the extra args and just show the cluster info. I am okay with the later but want to make sure we are consciously doing that and not because we missed this case.

fromanirh: the standard practice is to always run the sanity checks on args, but just ignore the value which aren't needed on a given flow. This is also what we do atm. I don't think this behaviour is surprising or unfriendly. Are you thinking about specific flows?

fromanirh:
/hold cancel
we have e2e tests :)

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -183,14 +190,95 @@ func getClusterData(mustGatherDirPath string) (ClusterData, error) {
...
+type NUMACellInfo struct {
+	ID       int   `json:"id"`
+	CoreList []int `json:"cores"`
+}
+
+// NodeInfo describe a Node in a MCP
+type NodeInfo struct {
+	Name           string         `json:"name"`
+	HTEnabled      bool           `json:"smt_enabled"`
+	NUMACellsCount int            `json:"numa_cells_count"`
```
swatisehgal, line 203: Is this really needed as we can compute this with by len(NUMACells) and don't see it being used for logging?

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -94,7 +96,12 @@ func NewRootCommand() *cobra.Command {
 				return fmt.Errorf("failed to parse the info flag: %v", err)
 			}
 			if info {
-				showClusterData(cluster)
+				// intentionally undocumented and unsupported output to make e2e testing easier.
+				if _, ok := os.LookupEnv("PERFORMANCE_PROFILE_CREATOR_OUTPUT_JSON"); ok {
```
swatisehgal, line 100: The flow of showClusterDataJSON and showClusterDataLog is exactly the same other than one  stores in a data structure and the other print to logs. Maybe we can streamline this by only having a single showClusterData with isOutputJsonEnabled as an arg and only store in the data structure if  isOutputJsonEnabled is true. WDYT?

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
```diff
@@ -75,6 +76,42 @@ var _ = Describe("[rfe_id:OCP-38968][ppc] Performance Profile Creator", func() {
...
+		// directory base name => full path
+		mustGatherDirs := getMustGatherDirs(mustGatherPath)
+
+		for name, path := range mustGatherDirs {
+			cmdArgs := []string{
+				"--info",
+				fmt.Sprintf("--must-gather-dir-path=%s", path),
+			}
+			cmdEnv := map[string]string{
+				"PERFORMANCE_PROFILE_CREATOR_OUTPUT_JSON": "1",
```
swatisehgal, line 92: Why is this equal to 1? I think it would be better to have `PERFORMANCE_PROFILE_CREATOR_OUTPUT_JSON= true`.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -94,7 +96,12 @@ func NewRootCommand() *cobra.Command {
 				return fmt.Errorf("failed to parse the info flag: %v", err)
 			}
 			if info {
-				showClusterData(cluster)
+				// intentionally undocumented and unsupported output to make e2e testing easier.
```
swatisehgal, line 99: I think we should document this as this could be useful if we want to parse the json data for some other purpose. I used it to pipe this into jq for better hierarchical view of the data! I think we should document this as another way to showing info and just mention in the README that if `PERFORMANCE_PROFILE_CREATOR_OUTPUT_JSON=true`  info mode would return an output in json format. By default when  `PERFORMANCE_PROFILE_CREATOR_OUTPUT_JSON` environment variable is not set or when`PERFORMANCE_PROFILE_CREATOR_OUTPUT_JSON=false`  info mode returns a human readable output.

swatisehgal (commented):
Nice work! Have a few questions/comments/recommendations :)

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -55,45 +55,90 @@ type ProfileData struct {
...
+		"mcp-name",
+		"rt-kernel",
+		"must-gather-dir-path",
+	}
+
+	root := &cobra.Command{
+		Use:   "performance-profile-creator",
+		Short: "A tool that automates creation of Performance Profiles",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !cmd.Flag("must-gather-dir-path").Changed {
```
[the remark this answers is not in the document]
fromanirh: While this makes sense from a code perspective (but looks like a sidegrade to me), from UX perspective I'm not convinced we gain much. Currently we get:
```
$ ./build/_output/bin/performance-profile-creator --info --reserved-cpu-count=2
Error: missing cluster data: please use the --must-gather-dir-path flag
Usage:
  performance-profile-creator [flags]

Flags:
[...]
```
which looks good to me because it clearly errors out and tells the user what's wrong and how to fix. Could you please elaborate a bit more the UX flow you would like to be improved?

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -146,23 +191,25 @@ func showClusterData(cluster ClusterData) {
...
-				log.Infof("Node: %s (NUMA cells: %d, HT: %v)", handle.Node.GetName(), len(topology.Nodes), htEnabled)
-				totalCPUs := 0
-				for id, node := range topology.Nodes {
-					var coreList []int
-					for _, core := range node.Cores {
-						coreList = append(coreList, core.LogicalProcessors...)
-					}
-					log.Infof("NUMA cell %d : %v", id, coreList)
-					totalCPUs += len(coreList)
+				continue
```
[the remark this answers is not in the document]
fromanirh: sure

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -183,14 +190,95 @@ func getClusterData(mustGatherDirPath string) (ClusterData, error) {
...
 
-func showClusterData(cluster ClusterData) {
+// NUMACellInfo describe a NUMA cell on a node
+type NUMACellInfo struct {
+	ID       int   `json:"id"`
+	CoreList []int `json:"cores"`
+}
+
+// NodeInfo describe a Node in a MCP
+type NodeInfo struct {
```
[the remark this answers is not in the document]
fromanirh: I'll start with a separate commit and squash here later if needed

cmd/performance-profile-creator/cmd/root.go
fromanirh: nope, not needed. Will remove.

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
fromanirh: shorter to type, and doesn't really make a difference since the value is ignored :)

cmd/performance-profile-creator/cmd/root.go
fromanirh: if we want to make this official we will need to streamline the UX a lot because suddenly requiring an environment variable to have a different output format is just bad design. We should rather turn `--info` in a value-requiring flag (e.g. `--info=json`, `--info=log`) or maybe even adding a `--format` option.
But there's more: if we support multiple output formats, then it becomes awkward to have an output mode which emits logs (on stderr) vs another which emits JSON (on stdout) which makes the output mode confusing.

cmd/performance-profile-creator/cmd/root.go
fromanirh: sure, let's see how it looks.

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
fromanirh: anyway, changing to "true"

cmd/performance-profile-creator/cmd/root.go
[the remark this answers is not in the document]
swatisehgal: Well, I was hoping that we would have consistent messaging across all the required args in case of `non-info` mode.
Specifically that we should have got:

`Error: missing required flags: must-gather-dir-path, mcp-name, rt-kernel`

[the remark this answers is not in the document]
fromanirh: ok, makes sense. FIxing.

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Yes, I was thinking about that as well. Better to have it in the UX itself! Lets go with --info=json, --info=log option and make log as default. That's what my instinct tells me :) 

Regarding the Json output to STDOUT, on one hand I think that json info to STDOUT should be okay because we either have performance profile or a json info sent to STDOUT on the other hand that I do understand it can be confusing. Maybe we can handle this with proper documentation?

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
swatisehgal: Let move this to UI instead!

cmd/performance-profile-creator/cmd/root.go
fromanirh: I think in info mode totally makes sense to have the output on stdout (that's what stdout is meant to be in the first place :) ).
However
1. it's weird to emit sometimes on stdout (proper output) and sometimes on stderr (because the output... is not error? is the expected output)
2. while this can be documented, I believe this leaves a quite strong "ok, but why?" feeling.

swatisehgal: Okay let me explain my point of view on this subject:
So if a user runs the following command
```
./build/_output/bin/performance-profile-creator --must-gather-dir-path=testdata/must-gather/must-gather.bare-metal --reserved-cpu-count=20 --rt-kernel=true --mcp-name=master > my-performanceprofile.yaml
```

If it works without error then great we have the performance profile stored in my-performanceprofile.yaml however if we were to send even the errors to STDOUT it would mean that  my-performanceprofile.yaml is storing the error message even without the user knowing about it!! This is something we don't want.

The case of info mode is different as the the user chooses to run the tool in info with json then that provides them the option to send it to a json file to view/use the cluster data . So fine to do that to STDOUT.
 
Hope this helps!

cmd/performance-profile-creator/cmd/root.go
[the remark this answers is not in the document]
swatisehgal: Ack

functests-performance-profile-creator/1_performance-profile_creator/ppc.go
fromanirh: sure, done!

cmd/performance-profile-creator/README.md
```diff
@@ -61,3 +61,11 @@ Depending on how must-gather directory was setup run the Performance profile Cre
    --rt-kernel=false --split-reserved-cpus-across-numa=true --topology-manager-policy=restricted \
    --power-consumption-mode=low-latency > performace-profile.yaml
    ```
+
+## Discovery mode
+
+To learn about the key details of the cluster you want to create a profile for, you may use the `discovery` (aka `info`) mode:
+```bash
+   ./hack/run-perf-profile-creator.sh -t must-gather.tar.gz -- --info
```
swatisehgal, line 69: This needs to be fixed now.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -94,7 +103,13 @@ func NewRootCommand() *cobra.Command {
 					return fmt.Errorf("failed to parse the cluster data: %v", err)
 				}
 
-				showClusterData(cluster)
+				clusterInfo := makeClusterInfoFromClusterData(cluster)
+				// intentionally undocumented and unsupported output to make e2e testing easier.
+				if infoMode == infoModeJSON {
+					showClusterInfoJSON(clusterInfo)
```
swatisehgal, line 109: This we miss the recommendation here: https://github.com/openshift-kni/performance-addon-operators/pull/628#discussion_r635967424? Not blocking in any way (can be done in another PR)?

fromanirh: my fault I missed it. Fixing.

cmd/performance-profile-creator/README.md
fromanirh: right, fixing

cmd/performance-profile-creator/cmd/root.go
swatisehgal: Ignore this comment: this is my bad

fromanirh:
cool!
```
curl: (56) OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 104
```

swatisehgal:
/lgtm

Thanks for your great work on this!

PR #633, opened 2021-05-19
e2e: hook in the render command e2e tests

fromanirh:
This PR makes sure we run the e2e render test in our CI flows.


fromanirh:
e2e-gcp lane failure is expected :)

fromanirh:
let's postpone the decision if we want to wire up the tests to an existing target or if we want a new target (and lane)

PR #634, opened 2021-05-20
[release 4.6] Enabled stalld only from a minimum kernel version

yanirq:

RHEL 8.2.z released with the fixed kernel: **kernel-4.18.0-193.49.1.el8_2** 

The enablement is determined by using tuned uname_regex option to check
if the kernel version prefix matches the minimum requirement


yanirq:
/hold  we cannot use this functionality until we switch to os shipped stalld .
**If** https://github.com/openshift/cluster-node-tuning-operator/pull/226 gets into 4.6 we can use this PR

yanirq:
this is an amended version of https://github.com/openshift-kni/performance-addon-operators/pull/638 to suit 4.6

yanirq:
closed in favor of https://github.com/openshift-kni/performance-addon-operators/pull/646

PR #637, opened 2021-05-24
e2e: add test for hk-cpus-inside-a-container - entrypoint

fromanirh:
We want the "housekeeping processes" inside a container,
so the entrypoint and anything we run with `oc exec` to
always execute in a well defined set of container cpus,
which we call "container HK cpus" (lacking better name).

This is not the cleanest behaviour, but it represents
an established usage pattern.

### DIFF
--- a/functests/1_performance/cpu_management.go
+++ b/functests/1_performance/cpu_management.go
@@ -33,6 +33,7 @@ import (
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
 )
 
+var workerRTNodes []corev1.Node
 var workerRTNode *corev1.Node
 var profile *performancev2.PerformanceProfile
 
@@ -53,7 +54,8 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 			Skip("Discovery mode enabled, performance profile not found")
 		}
 
-		workerRTNodes, err := nodes.GetByLabels(testutils.NodeSelectorLabels)
+		var err error
+		workerRTNodes, err = nodes.GetByLabels(testutils.NodeSelectorLabels)
 		Expect(err).ToNot(HaveOccurred())
 		workerRTNodes, err = nodes.MatchingOptionalSelector(workerRTNodes)
 		Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("error looking for the optional selector: %v", err))
@@ -215,6 +217,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 	When("pod runs with the CPU load balancing runtime class", func() {
 		var testpod *corev1.Pod
 		var defaultFlags map[int][]int
+		var annotations map[string]string
 
 		getCPUsSchedulingDomainFlags := func() (map[int][]int, error) {
 			cmd := []string{"/bin/bash", "-c", "more /proc/sys/kernel/sched_domain/cpu*/domain*/flags | cat"}
@@ -266,10 +269,9 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 			defaultFlags, err = getCPUsSchedulingDomainFlags()
 			Expect(err).ToNot(HaveOccurred())
 
-			annotations := map[string]string{
+			annotations = map[string]string{
 				"cpu-load-balancing.crio.io": "disable",
 			}
-			testpod = getTestPodWithAnnotations(annotations)
 		})
 
 		AfterEach(func() {
@@ -277,6 +279,8 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 		})
 
 		It("[test_id:32646] should disable CPU load balancing for CPU's used by the pod", func() {
+			testpod = getTestPodWithAnnotations(1, annotations)
+
 			By("Starting the pod")
 			err := testclient.Client.Create(context.TODO(), testpod)
 			Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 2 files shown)


functests/1_performance/cpu_management.go
```diff
@@ -324,6 +326,52 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+			By("Checking on which processor the container entry point is running")
+			// see: https://unix.stackexchange.com/questions/79520/determining-the-particular-processor-on-which-a-process-is-running
+			podCmd := []string{"/bin/bash", "-c", "/bin/cut -d' ' -f39 < /proc/1/stat"}
+			podOutput, err := pods.ExecCommandOnPod(testpod, podCmd)
+
+			psr, err := strconv.Atoi(strings.TrimSpace(string(podOutput)))
+			Expect(err).ToNot(HaveOccurred())
+
+			cpuIDs := cpus.ToSlice()
+			houseKeepingCPU := cpuIDs[0] // TODO this is not correct, also cpuIDs[0] thread sibling is actually OK
```
MarSik, line 368: Is the cpuIDs array always sorted?

fromanirh: Yep: https://github.com/kubernetes/kubernetes/blob/release-1.21/pkg/kubelet/cm/cpuset/cpuset.go#L191

functests/1_performance/cpu_management.go
```diff
@@ -324,6 +326,52 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
...
+			Expect(psr).To(Equal(houseKeepingCPU), fmt.Sprintf("pod entry point is running on cpu %d - housekeeping cpu = %d", psr, houseKeepingCPU))
+
+			By("Deleting the pod")
+			deleteTestPod(testpod)
```
Tal-or, line 372: If the test will fail, will the pod will be deleted anyway?

fromanirh: Yes, because we have the `AfterEach` above. Now I'm trying to think about the less confusing behaviour, though.

fromanirh:
/hold
- need to check failures
- need to make the tests themselves more robust (no pod leakage)

fromanirh:
/unhold
added all the changes I wanted to the test, ready for review

fromanirh:
/hold
we have a blocker issue here we need to discuss.
The bare minimum exclusive cores (as in virtual cpus) this test need is 4.
CI envs may NOT have 4 free exclusive cpus to run the tests. This will make the test fail.
So, either:
1. we check for availability of a worker in the cluster with 4+ allocatable cores, we skip the test (not fail) if not available
2. we accept the test will perma-fail and we move it into its own non-gating lane

Currently we implement option 1 but neither seems really satisfatory. Thoughts?

cynepco3hahue:
> we have a blocker issue here we need to discuss.
> The bare minimum exclusive cores (as in virtual cpus) this test need is 4.
> CI envs may NOT have 4 free exclusive cpus to run the tests. This will make the test fail.
> So, either:
> 
> 1. we check for availability of a worker in the cluster with 4+ allocatable cores, we skip the test (not fail) if not available
> 2. we accept the test will perma-fail and we move it into its own non-gating lane
> 
> Currently we implement option 1 but neither seems really satisfatory. Thoughts?

IMHO the first option is good enough because we have the d/s CI with bare metals and enough amount of CPUs, and I do not think any change under the PAO can break it.

fromanirh:
> > we have a blocker issue here we need to discuss.
> > The bare minimum exclusive cores (as in virtual cpus) this test need is 4.
> > CI envs may NOT have 4 free exclusive cpus to run the tests. This will make the test fail.
> > So, either:
> > 
> > 1. we check for availability of a worker in the cluster with 4+ allocatable cores, we skip the test (not fail) if not available
> > 2. we accept the test will perma-fail and we move it into its own non-gating lane
> > 
> > Currently we implement option 1 but neither seems really satisfatory. Thoughts?
> 
> IMHO the first option is good enough because we have the d/s CI with bare metals and enough amount of CPUs, and I do not think any change under the PAO can break it.

This is what I prefer as well. I can add a parameter/env var/whatever to make the test fail if the env is supposed to have the free cores, much like we do u/s. I'll push a followup commit in this PR about that.

functests/utils/cluster/cluster.go
```diff
@@ -0,0 +1,38 @@
...
+	"strconv"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// EnforceRequirements indicates if tests with specific cluster preconditions should fail rather than skip
+// if the cluster we are running against doesn't meet the requirements of the tests
+func EnforceRequirements() bool {
+	discoveryMode, _ := strconv.ParseBool(os.Getenv("ENFORCE_REQUIREMENTS"))
```
fromanirh, line 15: copy/paste artifact

cynepco3hahue (commented):
small nit, but except this one
/lgtm
/hold give people time to review it

functests/utils/cluster/cluster.go
```diff
@@ -0,0 +1,38 @@
...
+
+// EnforceRequirements indicates if tests with specific cluster preconditions should fail rather than skip
+// if the cluster we are running against doesn't meet the requirements of the tests
+func EnforceRequirements() bool {
+	enforceReqs, _ := strconv.ParseBool(os.Getenv("ENFORCE_REQUIREMENTS"))
+	return enforceReqs
+}
+
+func HaveEnoughCores(nodes []corev1.Node, cpus int) (bool, error) {
+	enoughCores := false
```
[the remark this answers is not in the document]
fromanirh: the intent of the code is to report no valid nodes unless proven otherwise (at least 1 suitable node found), if this is not clear enough I need to improve it, please feel free to suggest what confused you.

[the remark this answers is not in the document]
cynepco3hahue: mostly I meant to remove the variable and have something like 
```
func HaveEnoughCores(nodes []corev1.Node, cpus int) (bool, error) {
	requestCpu := resource.MustParse(fmt.Sprintf("%dm", cpus*1000))
	for _, node := range nodes {
		availCpu, ok := node.Status.Allocatable[corev1.ResourceCPU]
		if !ok {
			return false, fmt.Errorf("cpu resource not in allocatable on node %q", node.Name)
		}
		if availCpu.IsZero() {
			return false, fmt.Errorf("zero available cpu on node %q", node.Name)
		}

		if availCpu.Cmp(requestCpu) >= 1 {
			return true, nil
		}
	}

	return false, nil
}
```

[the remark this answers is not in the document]
fromanirh: done

fromanirh:
/hold
misread, sorry

functests/utils/cluster/cluster.go
```diff
@@ -29,3 +33,29 @@ func ComputeTestTimeout(baseTimeout time.Duration, isSno bool) time.Duration {
 
 	return testTimeout
 }
+
+// EnforceRequirements indicates if tests with specific cluster preconditions should fail rather than skip
+// if the cluster we are running against doesn't meet the requirements of the tests
+func EnforceRequirements() bool {
```
yanirq, line 39: maybe we can use this in other tests (different PR)

fromanirh: Yep, this is the plan

fromanirh:
/hold
we need a way to make this test skippable explicitely even if the cluster can run it.
Maybe just move to a different suite?

fromanirh:
actually, not needed anymore.

PR #638, opened 2021-05-25
[release 4.7] Enabled stalld only from a minimum openshift release

yanirq:

The minimum kernel versions required for stalld if in oc 4.7 are
4.18.0-240.22.1.el8_3.x86_64
4.18.0-240.22.1.rt7.77.el8_3.x86_64

These kernel version are availble only from openshift 4.7.7 and above

### DIFF
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/go-version"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gstruct"
@@ -203,16 +204,18 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 			err := testclient.Client.Get(context.TODO(), key, cv)
 			Expect(err).ToNot(HaveOccurred(), "cannot find the Cluster version object "+key.String())
 			clusterVersion = cv.Status.Desired.Version
-			splitClusterVersion := strings.Split(clusterVersion, ".")
-			Expect(len(splitClusterVersion) > 2).Should(BeTrue(), "Cluster version in X.Y.Z format cannot be found")
-			zStream, err := strconv.Atoi(splitClusterVersion[2])
-			// stalld should be enbaled for all version higher or equal to 4.7.6
-			if zStream >= 6 {
+			currentClusterVersion, err := version.NewVersion(clusterVersion)
+			Expect(err).ToNot(HaveOccurred())
+			requiredStalldClusterVersion, err := version.NewVersion("4.7.6")
+			Expect(err).ToNot(HaveOccurred())
+			if currentClusterVersion.GreaterThanOrEqual(requiredStalldClusterVersion) {
 				for _, node := range workerRTNodes {
 					tuned := tunedForNode(&node)
 					_, err = pods.ExecCommandOnPod(tuned, []string{"pidof", "stalld"})
 					Expect(err).ToNot(HaveOccurred())
 				}
+			} else {
+				klog.Warningf("Stalld is not enable, cluster version: %s", clusterVersion)
 			}
 
 			By("Checking minmal required kernel version")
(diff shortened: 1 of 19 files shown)


pkg/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -110,6 +112,14 @@ func NewNodePerformance(assetsDir string, profile *performancev2.PerformanceProf
 		templateArgs[templateGloballyDisableIrqLoadBalancing] = strconv.FormatBool(true)
 	}
 
+	splitClusterVersion := strings.Split(clusterVersion, ".")
```
cynepco3hahue, line 115: Can we use something nice to compare versions, for example, https://github.com/hashicorp/go-version will work for me.

functests/1_performance/performance.go
```diff
@@ -192,13 +194,55 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+			cv := &v1.ClusterVersion{}
+			key := types.NamespacedName{
+				Name: "version",
+			}
+			err := testclient.Client.Get(context.TODO(), key, cv)
+			Expect(err).ToNot(HaveOccurred(), "cannot find the Cluster version object "+key.String())
+			clusterVersion = cv.Status.Desired.Version
+			splitClusterVersion := strings.Split(clusterVersion, ".")
+			Expect(len(splitClusterVersion) > 2).Should(BeTrue(), "Cluster version in X.Y.Z format cannot be found")
+			zStream, err := strconv.Atoi(splitClusterVersion[2])
```
cynepco3hahue, line 208: I think CI release version can be different from the expected on because it built from master, so to debug e2e, you can print the version here.

yanirq: ack. notice that this is 4.7

functests/1_performance/performance.go
```diff
@@ -192,13 +195,55 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+			if currentClusterVersion.GreaterThanOrEqual(requiredStalldClusterVersion) {
+				for _, node := range workerRTNodes {
+					tuned := tunedForNode(&node)
+					_, err = pods.ExecCommandOnPod(tuned, []string{"pidof", "stalld"})
+					Expect(err).ToNot(HaveOccurred())
+				}
+			}
+
+			By("Checking minmal required kernel version")
+			stalldMinimumKernelVersionPrefix := "4.18.0-240.22"
```
cynepco3hahue, line 220: Does it still relevant to check the kernel version? I can not see that you check the kernel version under the controller.

yanirq: I am considering to leave that here since it already detected wrong versions in openshift releases. consider this as a double check . just need to see why its currently failing

yanirq:
currently setting stalld is not working against test cluster :
` Stalld is not enable, cluster version: 4.7.0-0.ci.test-2021-05-27-124943-ci-op-lzxyirkx-latest `
The current implementation is looking for  a version higher than 4.7.6

cynepco3hahue:
> currently setting stalld is not working against test cluster :
> `Stalld is not enable, cluster version: 4.7.0-0.ci.test-2021-05-27-124943-ci-op-lzxyirkx-latest`
> The current implementation is looking for a version higher than 4.7.6
> our CI clusters are always in a format of 4.7.0-ci-* vs real releases.
> cc @cynepco3hahue @MarSik

We can ask OpenShift CI folks if it is possible to get the expected z-stream version under the CI.

yanirq:
@MarSik @cynepco3hahue  
A possible solution would be : if the version does not conform with official releases then enable stalld.
meaning for CI and test releases that do not have a X.Y.Z (semver version) just enable by default.
The reasoning for that would be that we always test the latest lanes of ci and nightly builds in e2e.

functests/1_performance/performance.go
MarSik: I would like to see a comment explaining the purpose then, like you just did here in the review.

yanirq: done

MarSik (approved):

PR #639, opened 2021-05-26
E2E: irqbalance: improve the test output

fromanirh:
Make the test more robust and with better logging to help understanding why we're getting failures lately.

### DIFF
--- a/functests/utils/nodes/nodes.go
+++ b/functests/utils/nodes/nodes.go
@@ -195,6 +195,11 @@ func BannedCPUs(node corev1.Node) (banned cpuset.CPUSet, err error) {
 		return cpuset.NewCPUSet(), fmt.Errorf("failed to execute %v: %v", cmd, err)
 	}
 
+	if bannedCPUs == "" {
+		testlog.Infof("Banned CPUs on node %q returned empty set", node.Name)
+		return cpuset.NewCPUSet(), nil // TODO: should this be a error?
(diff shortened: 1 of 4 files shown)


functests/utils/nodes/nodes.go
```diff
@@ -195,6 +195,11 @@ func BannedCPUs(node corev1.Node) (banned cpuset.CPUSet, err error) {
 		return cpuset.NewCPUSet(), fmt.Errorf("failed to execute %v: %v", cmd, err)
 	}
 
+	if bannedCPUs == "" {
+		testlog.Infof("Banned CPUs on node %q returned empty set", node.Name)
+		return cpuset.NewCPUSet(), nil // TODO: should this be a error?
```
cynepco3hahue, line 200: I think we should return an error, otherwise, it will fail during the test check.

fromanirh: The more I think about it, the more I agree. Fixing.

fromanirh: I thought more, and I think that we should not fail here. We should fail one level up. From the perspective of this function having empty banned cpu string is legal and means "no cpus are banned".

cynepco3hahue: I just afraid that it will harder to debug the failing test. But in general, both approaches work for me.

PR #643, opened 2021-05-31
Bug 1957291: add resource request and toleration's to the operator

cynepco3hahue:
Add resource request to be sure that our operator pod will have buarstable QoS.
Add tolerations to be sure that our operator will not evict because of CPU or
memory consumptions peaks.

### DIFF
--- a/deploy/olm-catalog/performance-addon-operator/4.8.0/performance-addon-operator.v4.8.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/performance-addon-operator/4.8.0/performance-addon-operator.v4.8.0.clusterserviceversion.yaml
@@ -239,11 +239,23 @@ spec:
                 image: REPLACE_IMAGE
                 imagePullPolicy: Always
                 name: performance-operator
-                resources: {}
+                resources:
+                  requests:
+                    cpu: 10m
+                    memory: 25Mi
               serviceAccountName: performance-operator
               tolerations:
               - effect: NoSchedule
                 key: node-role.kubernetes.io/master
+                operator: Exists
+              - effect: NoExecute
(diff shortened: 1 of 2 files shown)


deploy/olm-catalog/performance-addon-operator/4.8.0/performance-addon-operator.v4.8.0.clusterserviceversion.yaml
```diff
@@ -239,11 +239,23 @@ spec:
                 image: REPLACE_IMAGE
                 imagePullPolicy: Always
                 name: performance-operator
-                resources: {}
+                resources:
+                  requests:
+                    cpu: 10m
```
fromanirh, line 244: is this realistic or a placeholder value we set to be able to have `requests`?
Same for memory below

fromanirh:
conceptually it's fine, question about the actual values.

deploy/olm-catalog/performance-addon-operator/4.8.0/performance-addon-operator.v4.8.0.clusterserviceversion.yaml
cynepco3hahue: in general, I just took some values from https://github.com/openshift/machine-config-operator/blob/master/manifests/machineconfigcontroller/deployment.yaml and halved it, because we do much less work than MC controller.

But mostly I added it to be sure that our pod under the buarstable QoS and not under the best-effort one.

fromanirh: ok, makes sense. And since these are requests (not limits), it seems like a good start.

PR #646, opened 2021-06-02
[release 4.6] Enabled stalld only from a minimum openshift release

yanirq:

The minimum kernel versions required for stalld if in oc 4.6 are
RHEL 8.2.z released with the fixed kernel: kernel-4.18.0-193.49.1.el8_2
rt kernel  kernel-4.18.0-193.49.1.rt.el8_2

These kernel version are availble only from openshift 4.6.27 and above
(4.18.0-193.51.1.el8_2.x86_64)


yanirq:
this is an amended version of #638 to suit 4.6

PR #647, opened 2021-06-03
Provide the runtime class documentation

cynepco3hahue:
The documentation will explain what annotations you can use together with the performance profile runtime class.


cynepco3hahue:
@MarSik FYI

docs/runtime-class.md
```diff
@@ -0,0 +1,27 @@
+# Performance profile runtime class
+
+The performance-addon-operators for each profile creates the runtime class that you can use under the pod specification. The runtime class allows you to configure a set of annotations:
```
Tal-or, line 3: Nit: I would rephrase that: ```The performance-addon-operators creates for each profile a runtime class that...```

PR #649, opened 2021-06-06
[release-4.7] Fix code to test number of Hugepages per numa node

mrniranjan:
Backporting to 4.7 
 
1. functests/nodes: Add function to return numa nodes and associated cpus of the node  Add shared function GetNumaNodes which returns the number of numa nodes  and the associated cpus

2. Fix functests/2_performance/updating_profile: fix test to get  number of hugepages. Modify the test code to get the numa  nodes and then fetch fetch 1G and 2M hugepages per numa node..

### DIFF
--- a/functests/2_performance_update/updating_profile.go
+++ b/functests/2_performance_update/updating_profile.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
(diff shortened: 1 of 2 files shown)


yanirq:
@mrniranjan lets keep in the PR title the convention of `[release 4.7] ` as a prefix instead of the suffix you added

PR #651, opened 2021-06-07
Bug 1963951: Improve the logic behind setting PAO components labels and selectors

cynepco3hahue:
The idea is to fetch the MCP related to the performance profile by the node selector and to take all relevant labels and selectors from it. When it not possible like in the case of the `machineConfigSelector` we will validate our generated machine config label with the machine config pool selector.

Once the performance profile does not have the related MCP, it will have degraded status and will not create any components.


controllers/performanceprofile_controller.go
```diff
@@ -522,3 +534,31 @@ func hasMatchingLabels(performanceprofile *performancev2.PerformanceProfile, mcp
 	}
 	return true
 }
+
+func (r *PerformanceProfileReconciler) getMachineConfigPoolByProfile(profile *performancev2.PerformanceProfile) (*mcov1.MachineConfigPool, error) {
+	nodeSelector := labels.Set(profile.Spec.NodeSelector)
+
+	mcpList := &mcov1.MachineConfigPoolList{}
```
MarSik, line 541: It seems you are matching MCPs that use the same node selector. But that might not be accurate. What about going over all matching nodes and using the machineconfiguration.openshift.io/currentConfig annotation on those nodes to collect the MCPs that are in charge of the matched nodes?

Also, in case multiple MCPs match, we must report a degraded condition as well!

MarSik: See also https://github.com/openshift-kni/performance-addon-operators/blob/77d9d8640391f3b54c86cb31c7b1c71a21558778/pkg/profilecreator/mcp.go#L120

cynepco3hahue: In general, it should not happen(two MCPs with the same node selector), and should be prevented by MCO validation, but IIRC the MCO does not have such validation, so you are right and I should handle it

yanirq:
Try to also consider what will happen in other layers that are using discovery mode. just double check.

cynepco3hahue:
> Try to also consider what will happen in other layers that are using discovery mode. just double check.

In general discovery mode assumes that performance profile already created:
If it targets the same nodes as some MCP we will not have any problems, otherwise, it will have degraded state(but also no configuration will applied to nodes), so IMHO it should work as it is

controllers/performanceprofile_controller.go
```diff
@@ -140,14 +141,14 @@ func (r *PerformanceProfileReconciler) mcpToPerformanceProfile(mcpObj client.Obj
 
 	var requests []reconcile.Request
 	for i, profile := range profiles.Items {
-		machineConfigPoolSelector := labels.Set(profileutil.GetMachineConfigPoolSelector(&profile))
-		selector, err := metav1.LabelSelectorAsSelector(mcp.Spec.MachineConfigSelector)
+		profileNodeSelector := labels.Set(profile.Spec.NodeSelector)
```
fromanirh, line 144: I'm not completely clear why we need this logic (this is a part of the code I don't really grok). Could you please elaborate to help me understand?

controllers/performanceprofile_controller.go
fromanirh: +1 about Martin's request

fromanirh (commented):
I'm less familiar than I'd like with the label matching logic, that I struggle a bit to grasp (in general). After some attempts, looks reasonnable with few minor comments/questions inside.

controllers/performanceprofile_controller.go
cynepco3hahue: I will start with the method description, we are using this method to initiate reconcile loop for MCP objects. But the reconcile loop should work only on one type of object, in our case, it is the performance profile. So during MCP create, update or delete our watch logic get the MCP object and trying to fetch from it the relevant performance profile, to fetch it the code checks the MCP node selector and finds all performance profiles with the same node selector. I hope it makes things clearer, give me know if you need a more descriptive explanation.

functests/utils/mcps/mcps.go
```diff
@@ -216,3 +218,19 @@ func WaitForProfilePickedUp(mcpName string, profile *performancev2.PerformancePr
 		return false
 	}, cluster.ComputeTestTimeout(10*time.Minute, runningOnSingleNode), 30*time.Second).Should(BeTrue(), "PerformanceProfile's %q MC was not picked up by MCP %q in time", profile.Name, mcpName)
 }
+
+func Delete(name string) error {
+	mcp := &machineconfigv1.MachineConfigPool{}
+	if err := testclient.Client.Get(context.TODO(), types.NamespacedName{Name: name}, mcp); err != nil {
```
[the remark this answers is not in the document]
cynepco3hahue: In general, I can do the empty object, but I think it is just the preferred way, to fetch an object before deletion, and it should not take any time, because we are fetching from the cache.

functests/utils/profiles/profiles.go
```diff
@@ -115,3 +115,20 @@ func WaitForCondition(nodeLabels map[string]string, conditionType v1.ConditionTy
 		return (GetCondition(nodeLabels, conditionType)).Status
 	}, 15*time.Minute, 30*time.Second).Should(Equal(conditionStatus), "Failed to met performance profile condition %v", conditionType)
 }
+
+// Delete delete the existing profile by name
+func Delete(name string) error {
+	profile := &performancev2.PerformanceProfile{}
+	if err := testclient.Client.Get(context.TODO(), types.NamespacedName{Name: name}, profile); err != nil {
```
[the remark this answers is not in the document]
cynepco3hahue: see my answer above

pkg/controller/performanceprofile/components/manifestset/manifestset.go
```diff
@@ -47,13 +49,21 @@ func (ms *ManifestResultSet) ToManifestTable() ManifestTable {
 }
 
 // GetNewComponents return a list of all component's instances that should be created according to profile
-func GetNewComponents(profile *performancev2.PerformanceProfile, assetDir *string) (*ManifestResultSet, error) {
+func GetNewComponents(profile *performancev2.PerformanceProfile, profileMCP *mcov1.MachineConfigPool, assetDir *string) (*ManifestResultSet, error) {
+	// we still need to construct the machineConfigPoolSelector when the command called from the render command
+	machineConfigPoolSelector := profilecomponent.GetMachineConfigPoolSelector(profile)
+	// we do not really need profile.spec.machineConfigPoolSelector anymore, but we should use it for backward
+	// compatability
+	if profile.Spec.MachineConfigPoolSelector == nil && profileMCP != nil {
```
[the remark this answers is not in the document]
cynepco3hahue: yes, it makes sense

controllers/performanceprofile_controller.go
cynepco3hahue: done, see - https://github.com/openshift-kni/performance-addon-operators/pull/651/files#diff-892a3fda0a96b281dc4e189cc080eeedfa1d9416d5967458ac7651adf7624af3R525

functests/utils/mcps/mcps.go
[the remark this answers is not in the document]
fromanirh: works for me

fromanirh (commented):
/lgtm
/hold
the core logic seems good to me. I'd prefer another ack before merge.

yanirq:
/lgtm
we should keep an eye on cnf-tests in case something was missed here and we did'nt notice

cynepco3hahue:
@yanirq I will create the following PR under the cnf-features-deploy to update PAO deps.

PR #652, opened 2021-06-08
e2e test  Automation for netqueues

mrniranjan:

### DIFF
--- a/functests/1_performance/netqueues.go
+++ b/functests/1_performance/netqueues.go
@@ -1,0 +1,350 @@
+package __performance
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
+	testutils "github.com/openshift-kni/performance-addon-operators/functests/utils"
+	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/cluster"
+	testlog "github.com/openshift-kni/performance-addon-operators/functests/utils/log"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/nodes"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/pods"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/profiles"
+)
+
+var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", func() {
+	var workerRTNodes []corev1.Node
+	var err error
+	var net bool = true
+	var profile, initialProfile *performancev2.PerformanceProfile
+	workerRTNodes, err = nodes.GetByLabels(testutils.NodeSelectorLabels)
+	Expect(err).ToNot(HaveOccurred())
+	workerRTNodes, err = nodes.MatchingOptionalSelector(workerRTNodes)
+	Expect(err).ToNot(HaveOccurred())
+	nodeLabel := testutils.NodeSelectorLabels
+	profile, err = profiles.GetByNodeLabels(nodeLabel)
+	Expect(err).ToNot(HaveOccurred())
+	testutils.BeforeAll(func() {
+		isSNO, err := cluster.IsSingleNode()
+		Expect(err).ToNot(HaveOccurred())
+		RunningOnSingleNode = isSNO
+		By("Backing up the profile")
+		initialProfile = profile.DeepCopy()
+	})
+
+	BeforeEach(func() {
+		profile, err = profiles.GetByNodeLabels(nodeLabel)
+		Expect(err).ToNot(HaveOccurred())
+		if profile.Spec.Net == nil {
+			By("Enable UserLevelNetworking in Profile")
+			profile.Spec.Net = &performancev2.Net{
+				UserLevelNetworking: &net,
+			}
+			By("Updating the performance profile")
+			profiles.UpdateWithRetry(profile)
+		}
+	})
+
+	AfterEach(func() {
+		By("Reverting the Profile")
+		spec, err := json.Marshal(initialProfile.Spec)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(testclient.Client.Patch(context.TODO(), profile,
+			client.RawPatch(
+				types.JSONPatchType,
+				[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
+			),
+		)).ToNot(HaveOccurred())
+	})
+
+	Context("Updating performance profile for netqueues", func() {
+		It("[test_id:40308][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Network device queues Should be set to the profile's reserved CPUs count ", func() {
+			devices := make(map[string][]int)
+			count := 0
+			if profile.Spec.Net != nil {
+				if profile.Spec.Net.UserLevelNetworking != nil && *profile.Spec.Net.UserLevelNetworking && len(profile.Spec.Net.Devices) == 0 {
+					By("To all non virtual network devices when no devices are specified under profile.Spec.Net.Devices")
+					err := checkDeviceSupport(workerRTNodes, devices)
+					Expect(err).ToNot(HaveOccurred())
+					for _, sizes := range devices {
+						for _, size := range sizes {
+							if size == getReservedCPUSize(profile.Spec.CPU) {
+								count++
+							}
+						}
+					}
+					if count == 0 {
+						Skip("Skipping Test: Unable to set Network queue size to reserved cpu count")
+					}
+				}
+			}
+		})
+
+		It("[test_id:40542] Verify the number of network queues of all supported network interfaces are equal to reserved cpus count", func() {
+
+			devices := make(map[string][]int)
+			count := 0
+			// Populate the device map with queue sizes
+			Eventually(func() bool {
+				err := checkDeviceSupport(workerRTNodes, devices)
+				Expect(err).ToNot(HaveOccurred())
+				return true
+			}, cluster.ComputeTestTimeout(200*time.Second, RunningOnSingleNode), testPollInterval*time.Second).Should(BeTrue())
+			//Verify the tuned profile is created on the worker-cnf nodes:
+			tunedCmd := []string{"tuned-adm", "profile_info", "openshift-node-performance-performance"}
+			for _, node := range workerRTNodes {
+				tunedPod := nodes.TunedForNode(&node, RunningOnSingleNode)
+				_, err := pods.ExecCommandOnPod(testclient.K8sClient, tunedPod, tunedCmd)
+				Expect(err).ToNot(HaveOccurred())
+			}
+			for _, sizes := range devices {
+				for _, size := range sizes {
+					if size == getReservedCPUSize(profile.Spec.CPU) {
+						count++
+					}
+				}
+			}
+			if count == 0 {
+				Skip("Skipping Test: Unable to set Network queue size to reserved cpu count")
+			}
+		})
+
+		It("[test_id:40543] Add interfaceName and verify the interface netqueues are equal to reserved cpus count.", func() {
+			devices := make(map[string][]int)
+			device := ""
+			count := 0
+			err := checkDeviceSupport(workerRTNodes, devices)
+			Expect(err).ToNot(HaveOccurred())
+			for d := range devices {
+				device = d
+			}
+			if profile.Spec.Net.UserLevelNetworking != nil && *profile.Spec.Net.UserLevelNetworking && len(profile.Spec.Net.Devices) == 0 {
+				By("Enable UserLevelNetworking and add Devices in Profile")
+				profile.Spec.Net = &performancev2.Net{
+					UserLevelNetworking: &net,
+					Devices: []performancev2.Device{
+						{
+							InterfaceName: &device,
+						},
+					},
+				}
+				By("Updating the performance profile")
+				profiles.UpdateWithRetry(profile)
+
+				Eventually(func() bool {
+					err := checkDeviceSupport(workerRTNodes, devices)
+					Expect(err).ToNot(HaveOccurred())
+					return true
+				}, cluster.ComputeTestTimeout(200*time.Second, RunningOnSingleNode), testPollInterval*time.Second).Should(BeTrue())
+			}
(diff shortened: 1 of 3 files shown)


mrniranjan:
Marking the PR wip to debug why network queues are not getting changed

yanirq:
@mrniranjan  please check if there is a duplication in one of the tests with https://github.com/openshift-kni/performance-addon-operators/blob/d9ca9e8e48a9dabad12ee294970b910318fb5c45/functests/1_performance/performance.go#L365 
If not then I suggest to move the test in here as well

mrniranjan:
> @mrniranjan please check if there is a duplication in one of the tests with
> 
> https://github.com/openshift-kni/performance-addon-operators/blob/d9ca9e8e48a9dabad12ee294970b910318fb5c45/functests/1_performance/performance.go#L365
> 
> If not then I suggest to move the test in here as well

@yanirq In terms of test case i don't think it's a duplicate,  But most of my test code is derived from this test case. The difference in the test case is that  test case in performance.go assumes that profile already has userLevelNetworking enabled.  Where as test case that i have enables userLevelNetworking if not present in the profile. I can move your test case to netqueues.go

cynepco3hahue:
Great looks like it were some flakiness with the upgrading job.

functests/1_performance/netqueues.go
```diff
@@ -0,0 +1,359 @@
...
+				out, err := pods.ExecCommandOnPod(tunedPod, cmdCombinedChannelsCurrent)
+				if strings.Contains(string(out), "n/a") {
+					fmt.Printf("Device %s doesn't support multiple queues\n", d)
+				} else {
+					channelCurrentCombined, err = strconv.Atoi(strings.TrimSpace(string(out)))
+					if err != nil {
+						testlog.Warningf(fmt.Sprintf("unable to retrieve current multi-purpose channels hardware settings for device %s on %s",
+							d, node.Name))
+					}
+					if channelCurrentCombined == 1 {
```
yanirq, line 290: channelCurrentCombined == 1 necessarily means the the device does not support multiple queues ?

functests/1_performance/netqueues.go
```diff
@@ -0,0 +1,359 @@
...
+			if err == nil {
+				cmdCombinedChannelsCurrent := []string{"bash", "-c",
+					fmt.Sprintf("ethtool -l %s | sed -n '/Current hardware settings:/,/Combined:/{s/^Combined:\\s*//p}'", d)}
+				out, err := pods.ExecCommandOnPod(tunedPod, cmdCombinedChannelsCurrent)
+				if strings.Contains(string(out), "n/a") {
+					fmt.Printf("Device %s doesn't support multiple queues\n", d)
+				} else {
+					channelCurrentCombined, err = strconv.Atoi(strings.TrimSpace(string(out)))
+					if err != nil {
+						testlog.Warningf(fmt.Sprintf("unable to retrieve current multi-purpose channels hardware settings for device %s on %s",
```
yanirq, line 287: I wonder if the warning is enough here,  this will cause the test **not** to fail on flakiness of getting the channels but would exclude an actual check for a specific device in some test.

functests/1_performance/netqueues.go
```diff
@@ -0,0 +1,359 @@
...
+							d, node.Name))
+					}
+					if channelCurrentCombined == 1 {
+						fmt.Printf("Device %s doesn't support multiple queues\n", d)
+					} else {
+						fmt.Printf("Device %s supports multiple queues\n", d)
+						devices[d] = channelCurrentCombined
```
yanirq, line 294: if setting the map  devices with a device name  and its corresponding value for each node we might end up overriding values here.
example:  node A has device eth0 with channelCurrentCombined =1 and node B has device eth0 with channelCurrentCombined=4 ,  what will happen ?

functests/1_performance/netqueues.go
```diff
@@ -0,0 +1,359 @@
...
+	// check the length of the map, if the map is empty , then
+	// there are no supported devices. Hence we skip
+	if len(devices) == 0 {
+		Skip("Skipping Test: There are no supported Network Devices")
+	}
+	return err
+}
+
+//Get Tuned Pods
+func getTunedPod(node *corev1.Node) *corev1.Pod {
```
yanirq, line 309: maybe we should extract this function and place it under the util lib since this is duplicated in https://github.com/openshift-kni/performance-addon-operators/blob/c3a829509aa75edfd7f3ef52081b6c27f96d5a3b/functests/1_performance/performance.go#L1154

functests/1_performance/netqueues.go
```diff
@@ -0,0 +1,359 @@
...
+		}
+		return true
+	}, cluster.ComputeTestTimeout(testTimeout, RunningOnSingleNode), testPollInterval*time.Second).Should(BeTrue(),
+		"there should be one tuned daemon per node")
+	return &tunedList.Items[0]
+}
+
+//Get Reserved CPU Count
+func getReservedCPUSize(CPU *performancev2.CPU) int {
+
```
yanirq, line 336: remove line

functests/1_performance/netqueues.go
```diff
@@ -0,0 +1,359 @@
...
+				return false
+			}
+		}
+		return true
+	}, cluster.ComputeTestTimeout(testTimeout, RunningOnSingleNode), testPollInterval*time.Second).Should(BeTrue(),
+		"there should be one tuned daemon per node")
+	return &tunedList.Items[0]
+}
+
+//Get Reserved CPU Count
```
yanirq, line 334: when the functions are not public (e.g start with capitals) I dont see the need for the comment above if the name is clear

functests/1_performance/netqueues.go
```diff
@@ -0,0 +1,359 @@
...
+		It("[test_id:40308][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Network device queues Should be set to the profile's reserved CPUs count ", func() {
+			devices := make(map[string]int)
+			count := 0
+			if profile.Spec.Net != nil {
+				if profile.Spec.Net.UserLevelNetworking != nil && *profile.Spec.Net.UserLevelNetworking && len(profile.Spec.Net.Devices) == 0 {
+					By("To all non virtual network devices when no devices are specified under profile.Spec.Net.Devices")
+					err := checkDeviceSupport(workerRTNodes, devices)
+					Expect(err).ToNot(HaveOccurred())
+					for _, size := range devices {
+						if size == getReservedCPUSize(profile.Spec.CPU) {
```
yanirq, line 60: why do we need this counter and check ?  I would fail the test in the first mismatch we get since the devices we get are suppose to be supported devices

functests/1_performance/netqueues.go
```diff
@@ -0,0 +1,359 @@
...
+
+		It("[test_id:40542] Verify the number of network queues of all supported network interfaces are equal to reserved cpus count", func() {
+			devices := make(map[string]int)
+			count := 0
+			if profile.Spec.Net == nil {
+				By("Enable UserLevelNetworking in Profile")
+				profile.Spec.Net = &performancev2.Net{
+					UserLevelNetworking: &net,
+				}
+				profiles.UpdateWithRetry(profile)
```
yanirq, line 79: notice that from that point the profile will have this updated field also for other tests.  think about reverting this at the end of the test

functests/1_performance/netqueues.go
```diff
@@ -0,0 +1,359 @@
...
+			}
+			// Populate the device map with queue sizes
+			Eventually(func() bool {
+				err := checkDeviceSupport(workerRTNodes, devices)
+				Expect(err).ToNot(HaveOccurred())
+				return true
+			}, cluster.ComputeTestTimeout(200*time.Second, RunningOnSingleNode), testPollInterval*time.Second).Should(BeTrue())
```
yanirq, line 86: what does RunningOnSingleNode mean here ?

functests/1_performance/netqueues.go
```diff
@@ -0,0 +1,359 @@
...
+			for _, size := range devices {
```
yanirq, line 87: same comment as above on why are we using count and not throwing an error in case of a mismatch

functests/1_performance/netqueues.go
```diff
@@ -0,0 +1,359 @@
...
+				if size == getReservedCPUSize(profile.Spec.CPU) {
+					count++
+				}
+			}
+			if count == 0 {
+				Skip("Skipping Test: Unable to set Network queue size to reserved cpu count")
+			}
+			//Verify the tuned profile is created on the worker-cnf nodes:
+			tunedCmd := []string{"tuned-adm", "profile_info", "openshift-node-performance-performance"}
```
yanirq, line 96: maybe this check should happen before we compare queue sizes to reserved CPU size ?

functests/1_performance/netqueues.go
mrniranjan: @yanirq i am not sure how to address this because since these tests are added in both upstream and down stream. And in upstream the support of supported netqueue devices is something not guaranteed. The tests always fails if i fail it.

functests/1_performance/netqueues.go
mrniranjan: @yanirq the reason is that on the upstream GCP RT lane, the hardware doesn't seem to support resize and so currently we are skipping in case failure .

functests/1_performance/netqueues.go
mrniranjan: @yanirq  The need for this occurred because in the GCP Lane the nic ens3 showed that it has support for queueresize (i came to this conclusion based on Combined queue size which it shows as 4 )  but it doesn't get resized , i could not debug it more. But the same code works on Downstream BM.

functests/1_performance/netqueues.go
yanirq: ack . so lets leave the warning

functests/1_performance/netqueues.go
yanirq: ack

functests/1_performance/netqueues.go
mrniranjan: In the latest patch, i have addressed this by using a integer slice . So the map looks like : map[string][]int.

functests/1_performance/netqueues.go
mrniranjan: I have not yet addressed this in my latest patch . I didn't had access to cluster to test it. I will address this as soon as i have access to BM cluster.

functests/1_performance/netqueues.go
mrniranjan: okay addressed it in the latest patch.

functests/1_performance/netqueues.go
mrniranjan: ah okay will address this by taking backup of the profile and updating it.

functests/1_performance/netqueues.go
mrniranjan: i didn't intend to use it, but based on examples of using Eventually, i found this usage. I am not sure if this usage makes it to run on SNO also (i have not tested it in SNO), but if you think this is un-necessary, i will remove it.

functests/1_performance/netqueues.go
mrniranjan: Yes in some BM, they don't support:
```
        Channel parameters for ens3:
        Pre-set maximums:
        RX:             n/a
        TX:             n/a
        Other:          n/a
        Combined:       1
        Current hardware settings:
        RX:             n/a
        TX:             n/a
        Other:          n/a
        Combined:       1
```

functests/1_performance/netqueues.go
mrniranjan: I tried putting in utils, but in the performance.go this function uses constants RunningOnSingleNode, testPollInterval and testPollInterval , i was not sure how to address this constants and yet keep it generic to be put in utils. Can you let me know how to address this.

functests/1_performance/netqueues.go
mrniranjan: updated in the latest patch.

functests/1_performance/netqueues.go
mrniranjan: @yanirq addressed this in the latest patch.

functests/1_performance/netqueues.go
mrniranjan: addressed this in latest patch.

functests/1_performance/netqueues.go
yanirq: ok, 1 is minimum anyway so its fine in our case.

functests/1_performance/netqueues.go
yanirq: we can expand tunedForNode to get these arguments or make them available in every suit

mrniranjan: i have addressed this in the latest patch

functests/1_performance/netqueues.go
yanirq: @mrniranjan can you point me to the change ?

mrniranjan: @yanirq https://github.com/mrniranjan/performance-addon-operators/blob/netqueues/functests/1_performance/netqueues.go#L97 the code now first checks the tuned profile and then the size. The same is for all the test cases:

functests/1_performance/netqueues.go
```diff
@@ -0,0 +1,350 @@
...
+			By("Updating the performance profile")
+			profiles.UpdateWithRetry(profile)
+		}
+	})
+
+	AfterEach(func() {
+		By("Reverting the Profile")
+		spec, err := json.Marshal(initialProfile.Spec)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(testclient.Client.Patch(context.TODO(), profile,
```
yanirq, line 66: is this still needed ?  you are updating the profile before each,  if its still needed i wonder if we need to get the profile first before patching it  (e.g. do profile, err = profiles.GetByNodeLabels(nodeLabel) before patching)

mrniranjan: Yes, i am updating the profile in BeforeEach if the spec.userLevelNetworking is nil . I will update the code to get Profile

mrniranjan: @yanirq Actually i just checked by printing the profile spec , we don't need to  add the profile, err = profiles.GetByNodeLabels(nodeLabel) in the Aftereach section as the profile object has the latest changes from the last run test. Actually adding the line profile, err = profiles.GetByNodeLabels(nodeLabel) is throwing the error:
```
    Timed out after 60.000s.
    Expected
        <*errors.StatusError | 0xc0003a3900>: {
            ErrStatus: {
                TypeMeta: {Kind: "", APIVersion: ""},
                ListMeta: {
                    SelfLink: "",
                    ResourceVersion: "",
                    Continue: "",
                    RemainingItemCount: nil,
                },
                Status: "Failure",
                Message: "Operation cannot be fulfilled on performanceprofiles.performance.openshift.io \"performance\": the object has been modified; please apply your changes to the latest version and try again",
                Reason: "Conflict",
                Details: {
                    Name: "performance",
                    Group: "performance.openshift.io",
                    Kind: "performanceprofiles",
                    UID: "",
                    Causes: nil,
                    RetryAfterSeconds: 0,
                },
                Code: 409,
            },
        }
    to be nil
```

yanirq: ack. lets keep this then and revisit if we get flaky results or unexpected outcomes in BM

PR #656, opened 2021-06-08
[release-4.8] Advance 4.7 bundle z-stream version

cynepco3hahue:


deploy/olm-catalog/performance-addon-operator/4.7.4/performance-addon-operator.v4.7.4.clusterserviceversion.yaml
```diff
@@ -161,8 +172,6 @@ spec:
           strategy: {}
           template:
             metadata:
-              annotations:
-                target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
```
fromanirh, line 165: we probably don't want to lose this

fromanirh (commented):
besides the fact it seems we're losing an important annotation (please check inside), looks good

deploy/olm-catalog/performance-addon-operator/4.7.4/performance-addon-operator.v4.7.4.clusterserviceversion.yaml
cynepco3hahue: it 4.7, and the feature is available only for 4.8, so it should be ok. I think it was added in first place by mistake :)

fromanirh: fair enough

yanirq:
/hold
is this the right release ?  this fix was made only for 4.6 and 4.7  , I see you are merging it to 4.8

cynepco3hahue:
> is this the right release ? this fix was made only for 4.6 and 4.7 , I see you are merging it to 4.8

The problem that 4.8 uses 4.7 bundle for the upgrade test.

yanirq:
> > is this the right release ? this fix was made only for 4.6 and 4.7 , I see you are merging it to 4.8
> 
> The problem that 4.8 uses 4.7 bundle for the upgrade test.

Ack
/hold cancel
/lgtm

PR #659, opened 2021-06-15
Bug 1967317: Do not create systemd update-rps service for veth devices

cynepco3hahue:
The start and shutdown of big amount of pods will initiate the creation of the systemd
service that should update the new interfaces `rps_cpus` mask and can create an additional
CPU load under the cluster.

The PR introduces to changes that should prevent it:
1. The OCI hook will update the pod virtual interfaces RPS mask under the node.
2. Exclude veth devices from the udev rule.


build/assets/scripts/low-latency-hooks.sh
```diff
@@ -9,6 +9,14 @@ pid=$(jq '.pid' /dev/stdin 2>&1)
 ns=$(ip netns identify "${pid}" 2>&1)
 [[ $? -eq 0 && -n "${ns}" ]] || { logger "${0} Failed to identify the namespace: ${ns}"; exit 0; }
 
+# Updates the container veth RPS mask on the node
+netns_link_indexes=$(ip netns exec "${ns}" ip -j link | jq ".[] | select(.link_index != null) | .link_index")
+for link_index in ${netns_link_indexes}; do
+  container_veth=$(ip -j link | jq ".[] | select(.ifindex == ${link_index}) | .ifname" | tr -d '"')
+  echo ${mask} > /sys/devices/virtual/net/${container_veth}/queues/rx-0/rps_cpus
```
yanirq, line 16: is that the only path to be added ?  rx-0 ?

cynepco3hahue: yep

cynepco3hahue:
@browsell @MarSik I think we want this fix for 4.8?

MarSik:
Yes, I think we want to backport it. Has it been tested yet?

cynepco3hahue:
> Yes, I think we want to backport it. Has it been tested yet?

I tested it on my cluster. @browsell Did you have a chance to run it on top of the SNO cluster?

cynepco3hahue:
We can not really on `ID_NET_DRIVER` because another udev rule set it by using `ethtool` and `sed` so probably it can be some races, but I can rely on `ENV{DEVPATH}` because it not affected by the renaming.

PR #664, opened 2021-06-21
Add test for cyclictest and hwlatetect binaries under cnf-tests image

Tal-or:
Those tests will imitate the same behavior as oslat has.
It will create a pod, run the binaries test and check the latency results.

### DIFF
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -130,7 +130,7 @@ var _ = Describe("[performance] Latency Test", func() {
 		})
 
 		It("should succeed", func() {
-			oslatArgs := []string {
+			oslatArgs := []string{
 				fmt.Sprintf("-runtime=%s", latencyTestRuntime),
 			}
 			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, oslatArgs)
@@ -167,6 +167,78 @@ var _ = Describe("[performance] Latency Test", func() {
 			}
 		})
 	})
+
+	Context("with the cyclictest image", func() {
+		BeforeEach(func() {
+			testName = "cyclictest"
+			initMaximumLatencyValue(testName)
+
+			if profile.Spec.CPU.Isolated == nil {
+				Skip(fmt.Sprintf("Skip the cyclictest test, the profile %q does not have isolated CPUs", profile.Name))
+			}
+		})
+
+		It("should succeed", func() {
+			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, []string{})
+			createLatencyTestPod(latencyTestPod)
+
+			// verify the maximum latency only when it requested, because this value can be very different
+			// on different systems
+			if maximumLatency == -1 {
+				return
+			}
+
+			latencies := extractLatencyValues("cyclictest", `# Max Latencies:\t*\s*(.*)\s*\t*`, workerRTNode)
+			for _, lat := range strings.Split(latencies, " ") {
+				if lat == "" {
+					continue
+				}
+
+				curr, err := strconv.Atoi(lat)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d", curr, maximumLatency)
+			}
+		})
+	})
+
+	Context("with the hwlatdetect image", func() {
+		BeforeEach(func() {
+			testName = "hwlatdetect"
+			initMaximumLatencyValue(testName)
+		})
+
+		It("should succeed", func() {
+			// This value should be > than max latency,
+			// in order to prevent the hwlatdetect return with error 1 in case latency value is bigger than expected.
+			// in case latency value is bigger than expected, it will be handled on different flow.
+			hardlimit := 1000
+			hwlatdetectArgs := []string{
+				fmt.Sprintf("-hardlimit=%d", hardlimit),
+			}
+
+			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, hwlatdetectArgs)
+			createLatencyTestPod(latencyTestPod)
+
+			// verify the maximum latency only when it requested, because this value can be very different
+			// on different systems
+			if maximumLatency == -1 {
+				return
+			}
+
+			latencies := extractLatencyValues(testName, `Max Latency:\t*\s*(.*)\s*\t*us`, workerRTNode)
+			for _, lat := range strings.Split(latencies, " ") {
+				if lat == "" {
+					continue
+				}
+
+				curr, err := strconv.Atoi(lat)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d", curr, maximumLatency)
+			}
+		})
+	})
 })
 
 func getLatencyTestPod(profile *performancev2.PerformanceProfile, node *corev1.Node, testName string, testSpecificArgs []string) *corev1.Pod {
@@ -182,7 +254,7 @@ func getLatencyTestPod(profile *performancev2.PerformanceProfile, node *corev1.N
 		fmt.Sprintf("-log_file=/host/%s.log", testName),
 	}
 
-	latencyTestRunnerArgs = append (latencyTestRunnerArgs, testSpecificArgs...)
+	latencyTestRunnerArgs = append(latencyTestRunnerArgs, testSpecificArgs...)
 
 	if latencyTestDelay > 0 {
 		latencyTestRunnerArgs = append(latencyTestRunnerArgs, fmt.Sprintf("-%s-start-delay=%d", testName, latencyTestDelay))
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -39,6 +39,12 @@ var (
 	maximumLatency     = -1
 )
 
+const (
+	oslatTestName       = "oslat"
+	cyclictestTestName  = "cyclictest"
+	hwlatdetectTestName = "hwlatdetect"
+)
+
 // LATENCY_TEST_DELAY delay the run of the oslat binary, can be useful to give time to the CPU manager reconcile loop
 // to update the default CPU pool
 // LATENCY_TEST_RUN: indicates if the latency test should run
@@ -69,7 +75,6 @@ var _ = Describe("[performance] Latency Test", func() {
 	var workerRTNode *corev1.Node
 	var profile *performancev2.PerformanceProfile
 	var latencyTestPod *corev1.Pod
-	var testName string
 
 	BeforeEach(func() {
 		if !latencyTestRun {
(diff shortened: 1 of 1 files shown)


fromanirh:
/retest
it seems infra issue

functests/4_latency/latency.go
```diff
@@ -63,25 +63,19 @@ func init() {
...
-		}
-	}
 }
 
 var _ = Describe("[performance] Latency Test", func() {
 	var workerRTNode *corev1.Node
 	var profile *performancev2.PerformanceProfile
-	var oslatPod *corev1.Pod
+	var latencyTestPod *corev1.Pod
+	var testName string
```
fromanirh, line 72: Can `testName` be moved on each specific `Context`? IOW, does it need to be a top-level variable?

functests/4_latency/latency.go
```diff
@@ -253,3 +245,49 @@ func getOslatPod(profile *performancev2.PerformanceProfile, node *corev1.Node) *
...
+	Expect(maximumLatency).ToNot(BeNil())
+
+	return latencies
+}
+
+// initMaximumLatencyValue should look for one of the following environment variables:
+// OSLAT_MAXIMUM_LATENCY: the expected maximum latency for all buckets in us
+// CYCLICTEST_MAXIMUM_LATENCY: the expected maximum latency for all buckets in us
+// HWLATDETECT_MAXIMUM_LATENCY: the expected maximum latency for all buckets in us
+func initMaximumLatencyValue(testName string) {
```
fromanirh, line 356: I'd call it `checkMaximumEnvironLatencyValue`

functests/4_latency/latency.go
```diff
@@ -253,3 +245,49 @@ func getOslatPod(profile *performancev2.PerformanceProfile, node *corev1.Node) *
...
+	envVariableName := fmt.Sprintf("%s_MAXIMUM_LATENCY", strings.ToUpper(testName))
```
fromanirh, line 357: we should validate `testName` is one of the allowed values, return a proper error if not

functests/4_latency/latency.go
```diff
@@ -253,3 +245,49 @@ func getOslatPod(profile *performancev2.PerformanceProfile, node *corev1.Node) *
...
+	maximumLatencyEnv := os.Getenv(envVariableName)
+	if maximumLatencyEnv != "" {
+		var err error
+		if maximumLatency, err = strconv.Atoi(maximumLatencyEnv); err != nil {
+			klog.Fatalf("the environment variable %q has incorrect value %q", envVariableName, maximumLatencyEnv)
```
fromanirh, line 362: Not a fan of failing by side effect, especially on helper functions. It would be better to return a proper error and fail in the calling `Context`

functests/4_latency/latency.go
```diff
@@ -167,6 +167,78 @@ var _ = Describe("[performance] Latency Test", func() {
 			}
 		})
 	})
+
+	Context("with the cyclictest image", func() {
+		BeforeEach(func() {
+			testName = "cyclictest"
+			initMaximumLatencyValue(testName)
+
+			if profile.Spec.CPU.Isolated == nil {
```
fromanirh, line 176: should this be made a common check? maybe merged with the check we currently do in `initMaximumLatencyValue`?

functests/4_latency/latency.go
```diff
@@ -167,6 +167,78 @@ var _ = Describe("[performance] Latency Test", func() {
...
+			}
+		})
+
+		It("should succeed", func() {
+			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, []string{})
+			createLatencyTestPod(latencyTestPod)
+
+			// verify the maximum latency only when it requested, because this value can be very different
+			// on different systems
+			if maximumLatency == -1 {
```
fromanirh, line 187: better log a message (or Skip) here

functests/4_latency/latency.go
```diff
@@ -167,6 +167,78 @@ var _ = Describe("[performance] Latency Test", func() {
...
+	Context("with the hwlatdetect image", func() {
+		BeforeEach(func() {
+			testName = "hwlatdetect"
+			initMaximumLatencyValue(testName)
+		})
+
+		It("should succeed", func() {
+			// This value should be > than max latency,
+			// in order to prevent the hwlatdetect return with error 1 in case latency value is bigger than expected.
+			// in case latency value is bigger than expected, it will be handled on different flow.
```
fromanirh, line 214: so should it possibly overridable by users, but with a sane defaults?

functests/4_latency/latency.go
```diff
@@ -167,6 +167,78 @@ var _ = Describe("[performance] Latency Test", func() {
...
+			hwlatdetectArgs := []string{
+				fmt.Sprintf("-hardlimit=%d", hardlimit),
+			}
+
+			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, hwlatdetectArgs)
+			createLatencyTestPod(latencyTestPod)
+
+			// verify the maximum latency only when it requested, because this value can be very different
+			// on different systems
+			if maximumLatency == -1 {
```
fromanirh, line 225: we need a log or a `Skip` here

fromanirh (commented):
mostly ok, but quite some questions inside

functests/4_latency/latency.go
Tal-or: Not for all tools. 
For example, ```hwlatdetect``` doesn't care about this check.

functests/4_latency/latency.go
Tal-or: Will add a log message for that, but what do you mean by skip? isn't ```return``` act as a skip in this context?

functests/4_latency/latency.go
Tal-or: This is just a sanity test for the tools per se.
The user case is different, he will run the ```cnf-tests``` image directly and provides the parameters as it sees fit.

functests/4_latency/latency.go
fromanirh: surely with return the test won't go any further (thus won't fail) but will this be notified or logged in any way? at the end of the standard ginkgo report we have the count of test succeeded, failed and skipped.
We should make sure is trivial to understand what happened, adding an explicit skip seems the best approach, logging and returning is the second best, doing nothing silently is the less convenient option.

Tal-or: I see. so I'm not sure that skip is correct, because even if we are not evaluating the latency values, we still made sure the binaries and mechanism are functions properly - which is the main purpose of the test though.

cynepco3hahue: In general, my ideas was:
1. Running test under the openshift CI with -1 to check if binaries work
2. Running test under the baremetal CI with 20us to see if have expected value for the latency

functests/4_latency/latency.go
cynepco3hahue: What exactly hard limit should check?

Tal-or: TL;DR - A hack to avoid ```hwlatdetect``` from return with an error.
As mentioned in the comment, in case that the actual latency is bigger than the expected one,  ```hwlatdetect``` will return with an error.
This is undesired behavior because we aren't always want to fail the test on anomaly latency values (in case we do, we have a separate check for that [here](https://github.com/openshift-kni/performance-addon-operators/pull/664/files#diff-a11540b25a190b62a4789801be3fa3b70b91847dc903680998aa8a5837550646R252))

So by configuring ```hardlimit``` we are telling ```hwlatdetect``` to ignore the configured latency value, and return with an error only if the latency value is bigger than the ```hardlimit``` value instead (which is incredibly high and not expected).

cynepco3hahue: so, you can rewrite the test
```
hardLimit := maximumLatency
if hardLimit == -1 {
    hardlimit := 1000
}

hwlatdetectArgs := []string{
    fmt.Sprintf("-hardlimit=%d", hardlimit),
}

latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, hwlatdetectArgs)
createLatencyTestPod(latencyTestPod)
```
and you can drop all part related to latency parsing

Tal-or: I thought about that, but then we might not be able to distinguish between an error code that returns from ```hwlatdetect``` because of high latency value or some unrelated internal error, right?

cynepco3hahue: you can just print the log output at the end of the test to make it more clear, you anyway do not return any viable error in case if the pod failed to start.

functests/4_latency/latency.go
fromanirh: ack

functests/4_latency/latency.go
```diff
@@ -185,6 +195,7 @@ var _ = Describe("[performance] Latency Test", func() {
 			// verify the maximum latency only when it requested, because this value can be very different
 			// on different systems
 			if maximumLatency == -1 {
+				klog.Info("no maximum latency value provided, skips the latency value check")
```
fromanirh, line 198: what the benefits of this approach over `Skip` are, again?

fromanirh (commented):
a question inside and we're good to go

functests/4_latency/latency.go
```diff
@@ -349,17 +350,20 @@ func extractLatencyValues(testName string, exp string, node *corev1.Node) string
...
 }
 
-// initMaximumLatencyValue should look for one of the following environment variables:
+// checkMaximumEnvironLatencyValue should look for one of the following environment variables:
 // OSLAT_MAXIMUM_LATENCY: the expected maximum latency for all buckets in us
 // CYCLICTEST_MAXIMUM_LATENCY: the expected maximum latency for all buckets in us
 // HWLATDETECT_MAXIMUM_LATENCY: the expected maximum latency for all buckets in us
-func initMaximumLatencyValue(testName string) {
+func checkMaximumEnvironLatencyValue(testName string) error {
+	if testName != strings.ToUpper(oslatTestName) && strings.ToUpper(cyclictestTestName) != strings.ToUpper(hwlatdetectTestName) {
```
fromanirh, line 358: copy/paste artifact?
this part of the `if` seems to be always true:
`strings.ToUpper(cyclictestTestName) != strings.ToUpper(hwlatdetectTestName)`

Tal-or: yup. nice catch

functests/4_latency/latency.go
Tal-or: Because even if we are not evaluating the latency values, we still need to make sure the binaries and mechanism are functions properly - which is the main purpose of the test though.
If we'll skip the whole test we won't have any indication, not even if the binary run is good.

functests/4_latency/latency.go
Tal-or: Because even if we are not evaluating the latency values, we still need to make sure the binaries and mechanism are functions properly - which is the main purpose of the test though.
If we'll skip the whole test we won't have any indication, not even if the binary run is good.

fromanirh: I think I'm missing something.
With the current approach we go this far, then we return. And the test is accounted as succeeded. We will later need to dig in the logs to learn that it was actually skipped.
If we skip here (instead of return), we can clearly see the test was skipped, and the message is presented to us.

cynepco3hahue: Probably @fromanirh is right.
We drop some specific skip message, something like `skip buckets latency check`, and if the test will fail before you still will see the failure.

fromanirh (approved):
/lgtm

fromanirh:
@cynepco3hahue if you are happy (changes after you reviewed), we can remove the hold

PR #673, opened 2021-06-28
Enable workload partitioning work

Tal-or:
This PR contains 2 main parts:
1. PerformanceProfileInfo struct which is purpose is to allow the operator to handle extra data which it cares about, but it's not part of the public API.
2. Allow render command to add the necessary bits and bytes to ```MachineConfig``` object in order to enable the workload partitioning feature.

The API for the ```WorkloadPartition```  is still missing, hence only the logic for the render command was implemented.

Once the ```WorkloadPartition``` will be available through the API, I'll post additional PR with the logic for watching the object inside the reconciliation loop and update the ```MachineConfig``` accordingly.

### DIFF
--- a/build/assets/configs/01-workload-partitioning.conf
+++ b/build/assets/configs/01-workload-partitioning.conf
@@ -1,0 +1,4 @@
+[crio.runtime.workloads.management]
+activation_annotation = "target.workload.openshift.io/management"
+annotation_prefix = "resources.workload.openshift.io"
+resources = { "cpuset" = "{{.ReservedCpus}}" }
--- a/build/assets/configs/openshift-workload-pinning.json
+++ b/build/assets/configs/openshift-workload-pinning.json
@@ -1,0 +1,5 @@
+{
+  "management": {
+    "cpuset": "{{.ReservedCpus}}"
+  }
+}
--- a/controllers/performanceprofile_controller.go
+++ b/controllers/performanceprofile_controller.go
@@ -27,6 +27,7 @@ import (
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/machineconfig"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/manifestset"
 	profileutil "github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/profile"
+	pinfo "github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/profileinfo"
 	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
 	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
@@ -342,36 +343,41 @@ func (r *PerformanceProfileReconciler) applyComponents(profile *performancev2.Pe
 		return nil, nil
 	}
 
-	components, err := manifestset.GetNewComponents(profile, profileMCP, &r.AssetsDir)
+	profileInfo := &pinfo.PerformanceProfileInfo{
+		PerformanceProfile:       *profile,
+		WorkloadPartitionEnabled: false,
+	}
+
+	newComp, err := manifestset.GetNewComponents(profileInfo, profileMCP, &r.AssetsDir)
 	if err != nil {
 		return nil, err
 	}
-	for _, componentObj := range components.ToObjects() {
+	for _, componentObj := range newComp.ToObjects() {
 		if err := controllerutil.SetControllerReference(profile, componentObj, r.Scheme); err != nil {
 			return nil, err
 		}
 	}
 
 	// get mutated machine config
-	mcMutated, err := r.getMutatedMachineConfig(components.MachineConfig)
+	mcMutated, err := r.getMutatedMachineConfig(newComp.MachineConfig)
 	if err != nil {
 		return nil, err
 	}
 
 	// get mutated kubelet config
-	kcMutated, err := r.getMutatedKubeletConfig(components.KubeletConfig)
+	kcMutated, err := r.getMutatedKubeletConfig(newComp.KubeletConfig)
 	if err != nil {
 		return nil, err
 	}
 
 	// get mutated performance tuned
-	performanceTunedMutated, err := r.getMutatedTuned(components.Tuned)
+	performanceTunedMutated, err := r.getMutatedTuned(newComp.Tuned)
 	if err != nil {
 		return nil, err
 	}
 
 	// get mutated RuntimeClass
-	runtimeClassMutated, err := r.getMutatedRuntimeClass(components.RuntimeClass)
+	runtimeClassMutated, err := r.getMutatedRuntimeClass(newComp.RuntimeClass)
 	if err != nil {
 		return nil, err
 	}
--- a/controllers/performanceprofile_controller_test.go
+++ b/controllers/performanceprofile_controller_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/kubeletconfig"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/machineconfig"
+	pinfo "github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/profileinfo"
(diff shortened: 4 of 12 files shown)


fromanirh:
would it make sense to create an *internal-only* type which extends a `PerformanceProfile` and which have extra fields, like `isWorkloadPartitionEnabled`?

Tal-or:
> would it make sense to create an _internal-only_ type which extends a `PerformanceProfile` and which have extra fields, like `isWorkloadPartitionEnabled`?

You mean to add an additional field under the ```PerformanceProfile``` struct?

fromanirh:
> > would it make sense to create an _internal-only_ type which extends a `PerformanceProfile` and which have extra fields, like `isWorkloadPartitionEnabled`?
> 
> You mean to add an additional field under the `PerformanceProfile` struct?

Something like

```
type PerformanceProfileInfo struct {
    pp *v2.PerformanceProfile
    // extra data the operator cares about but which is not part of the public API
    WorkloadPartitionEnabled bool
}
```

Embedding the type is another option but I think we need to carefully evaluate the implications on backward compatibility.
Adding a pointer field should be fine, we don't have hot loops in PAO.

build/assets/configs/01-workload-partitioning.conf
```diff
@@ -0,0 +1,4 @@
+[crio.runtime.workloads.management]
+activation_annotation = "target.workload.openshift.io/management"
+annotation_prefix = "resources.workload.openshift.io"
+resources = { "cpushares" = 0, "cpuset" = "{{.ReservedCpus}}" }
```
cynepco3hahue, line 4: remove `cpushares` and add new line

build/assets/configs/openshift-workload-pinning.json
```diff
@@ -0,0 +1,5 @@
+{
+  "management": {
+    "cpuset": "{{.ReservedCpus}}"
+  }
+}
```
cynepco3hahue, line 5: new line

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -80,7 +83,7 @@ const (
 )
 
 // New returns new machine configuration object for performance sensitive workloads
-func New(assetsDir string, profile *performancev2.PerformanceProfile) (*machineconfigv1.MachineConfig, error) {
+func New(assetsDir string, profile *performancev2.PerformanceProfile, mngPartition bool) (*machineconfigv1.MachineConfig, error) {
```
cynepco3hahue, line 86: mngPartition -> workloadPartitioningEnabled

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -159,6 +162,39 @@ func getIgnitionConfig(assetsDir string, profile *performancev2.PerformanceProfi
 		return nil, err
 	}
 
+	// add crio config snippet under the node /etc/crio/crio.conf.d/ directory
```
cynepco3hahue, line 165: and workload partition file under `/etc/kubernetes`

cynepco3hahue (commented):
Please add unittest for new functionallity

controllers/performanceprofile_controller.go
```diff
@@ -342,36 +343,41 @@ func (r *PerformanceProfileReconciler) applyComponents(profile *performancev2.Pe
 		return nil, nil
 	}
 
-	components, err := manifestset.GetNewComponents(profile, profileMCP, &r.AssetsDir)
+	profileInfo := &pinfo.PerformanceProfileInfo{
+		PerformanceProfile:       *profile,
+		WorkloadPartitionEnabled: false,
```
Tal-or, line 348: When the API for the ```WorkloadPartition``` object will be ready, we will have to watch for the object's existence and change this field accordingly.

pkg/cmd/render/render.go
```diff
@@ -105,6 +111,14 @@ func (r *renderOpts) readFlagsFromEnv() {
 	if assetsOutDir := os.Getenv("ASSET_OUTPUT_DIR"); len(assetsOutDir) > 0 {
 		r.assetsOutDir = assetsOutDir
 	}
+
+	if workloadPartitioningEnabled := os.Getenv("ENABLE_WORKLOAD_PARTITIONING"); len(workloadPartitioningEnabled) > 0 {
```
fromanirh, line 115: (very) minor: or use `os.LookupEnv`

controllers/performanceprofile_controller_test.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/kubeletconfig"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/machineconfig"
+	pinfo "github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/profileinfo"
```
yanirq, line 19: do we have 2 types of profile info ?  why do we need a differentiation here?

functests-render-command/1_render_command/render_test.go
```diff
@@ -16,55 +16,109 @@ var (
...
+			cmdLineArgs = append(cmdLineArgs,
 				"--performance-profile-input-files", ppInFiles,
 				"--asset-input-dir", assetsInDir,
 				"--asset-output-dir", assetsOutDir,
-			}
-			fmt.Fprintf(GinkgoWriter, "running: %v\n", cmdline)
+			)
+		})
+
+		It("should produces the expected components to output directory", func() {
```
yanirq, line 56: s/produces/produce

functests-render-command/1_render_command/render_test.go
```diff
@@ -16,55 +16,109 @@ var (
...
+			file, diff, err := cmpRenderToRef()
```
yanirq, line 57: s/cmpRenderToRef/compareRenderToreference 

In general I would rather have longer name with clear meaning

functests-render-command/1_render_command/render_test.go
```diff
@@ -78,27 +132,48 @@ func cleanArtifacts() {
 	os.RemoveAll(assetsOutDir)
 }
 
-func runAndCompare(cmd *exec.Cmd) {
-	_, err := cmd.Output()
-	Expect(err).ToNot(HaveOccurred())
-
+func cmpRenderToRef() (string, string, error) {
```
yanirq, line 135: just a general question,  will we always have the same file structure after rendering ?  can we have for example a right outcome but in different layout on the file? (whitespaces,different lines,etc)

pkg/controller/performanceprofile/components/profileinfo/profileinfo.go
```diff
@@ -0,0 +1,10 @@
+package profileinfo
+
+import performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
+
+// PerformanceProfileInfo is a wrapper for PerformanceProfile that can hold extra data and configuration
+type PerformanceProfileInfo struct {
+	performancev2.PerformanceProfile
```
cynepco3hahue, line 7: please use a pointer

pkg/controller/performanceprofile/components/profileinfo/profileinfo.go
```diff
@@ -0,0 +1,10 @@
+package profileinfo
+
+import performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
+
+// PerformanceProfileInfo is a wrapper for PerformanceProfile that can hold extra data and configuration
+type PerformanceProfileInfo struct {
```
cynepco3hahue, line 6: I am still very skeptical about why do we need this new struct, I do not see a reason to merge this PR until the workload partition API is available and once it will be available you can update PR and we can merge it.
Once you will have API you can fetch the relevant resource under the controller and pass worload partition boolean to the MachineConfig generator.

controllers/performanceprofile_controller_test.go
Tal-or: This is just a shorthand instead of using the full ```profileinfo``` name

pkg/controller/performanceprofile/components/profileinfo/profileinfo.go
fromanirh: For the record this was done after my suggestion. Since we are embedding the `PerformanceProfile` struct, and we can pass the new type by pointer like we do for `PerformanceProfile`, this seems ok. I don't really mind moving to the pointer - or removing the new type entirely, though.

Tal-or:
Obsolete

Tal-or:
Obsolete and probably won't be relevant. Closing.

PR #674, opened 2021-06-28
automated 1972701 bug verification

shajmakh:
automated 1972701 bug verification :check that stalld is running as FIFO with priority 10

### DIFF
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -200,6 +200,18 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				Expect(err).ToNot(HaveOccurred())
 			}
 		})
+		It("[test_id:42400][crit:medium][vendor:cnf-qe@redhat.com][level:acceptance] stalld daemon is running as sched_fifo", func() {
+			for _, node := range workerRTNodes {
+				pid, err := nodes.ExecCommandOnNode([]string{"pgrep", "-f", "stalld"}, &node)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(pid).ToNot(BeEmpty())
+				sched_tasks, err := nodes.ExecCommandOnNode([]string{"chrt", "-ap", pid}, &node)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(sched_tasks).To(ContainSubstring("scheduling policy: SCHED_FIFO"))
+				Expect(sched_tasks).To(ContainSubstring("scheduling priority: 10"))
+			}
+		})
+
 	})
 
 	Context("Additional kernel arguments added from perfomance profile", func() {


functests/1_performance/performance.go
```diff
@@ -200,6 +200,15 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				Expect(err).ToNot(HaveOccurred())
 			}
 		})
+		It("[test_id:42400][crit:medium][vendor:cnf-qe@redhat.com][level:acceptance] stalld daemon is running as sched_fifo", func() {
+                         for _, node := range workerRTNodes {
+                                 pid, err :=nodes.ExecCommandOnNode([]string{"pgrep", "-f","stalld"},&node)
+                                 Expect(err).ToNot(HaveOccurred())
+                                 sched_tasks,_:=nodes.ExecCommandOnNode([]string{"chrt","-ap",pid},&node)
+                                 Expect(strings.Contains(sched_tasks,"scheduling policy: SCHED_FIFO") && strings.Contains(sched_tasks,"scheduling priority: 10")).To(Equal(true))
```
fromanirh, line 208: This is a bit convoluted. Better:
```
Expect(sched_tasks).To(ContainSubstring("scheduling policy: SCHED_FIFO"))
Expect(sched_tasks).To(ContainSubstring("scheduling priority: 10"))
```

functests/1_performance/performance.go
```diff
@@ -200,6 +200,15 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				Expect(err).ToNot(HaveOccurred())
 			}
 		})
+		It("[test_id:42400][crit:medium][vendor:cnf-qe@redhat.com][level:acceptance] stalld daemon is running as sched_fifo", func() {
+                         for _, node := range workerRTNodes {
+                                 pid, err :=nodes.ExecCommandOnNode([]string{"pgrep", "-f","stalld"},&node)
```
fromanirh, line 205: weird spacing here

fromanirh (commented):
could you please add a bit more context in the commit message (preferred) and/or in the PR description (copypasting the same commit message is fine).
Also, you may want to check your git configuration, because it looks like the commit is from "root" (?)

functests/1_performance/performance.go
shajmakh: thanks , I modified it .

functests/1_performance/performance.go
shajmakh: Fixed , thanks !

functests/1_performance/performance.go
```diff
@@ -200,13 +200,14 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				Expect(err).ToNot(HaveOccurred())
 			}
 		})
-		It("[test_id:42400][crit:medium][vendor:cnf-qe@redhat.com][level:acceptance] stalld daemon is running as sched_fifo", func() {
+		FIt("[test_id:42400][crit:medium][vendor:cnf-qe@redhat.com][level:acceptance] stalld daemon is running as sched_fifo", func() {
```
fromanirh, line 203: Do we (still?) need `FIt`?

functests/1_performance/performance.go
```diff
@@ -200,13 +200,14 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
                          for _, node := range workerRTNodes {
-                                 pid, err :=nodes.ExecCommandOnNode([]string{"pgrep", "-f","stalld"},&node)
+                                 pid, err := nodes.ExecCommandOnNode([]string{"pgrep", "-f", "stalld"}, &node)
                                  Expect(err).ToNot(HaveOccurred())
-                                 sched_tasks,_:=nodes.ExecCommandOnNode([]string{"chrt","-ap",pid},&node)
-                                 Expect(strings.Contains(sched_tasks,"scheduling policy: SCHED_FIFO") && strings.Contains(sched_tasks,"scheduling priority: 10")).To(Equal(true))
-                         }
+                                 sched_tasks,_:=nodes.ExecCommandOnNode([]string{"chrt", "-ap", pid}, &node)
+				 Expect(sched_tasks).To(ContainSubstring("scheduling policy: SCHED_FIFO"))
+				 Expect(sched_tasks).To(ContainSubstring("scheduling priority: 10"))
```
fromanirh, line 209: still weird spacing, and the CI lane is probably not going to like it

fromanirh (commented):
thanks, looks better. Could you squash this commit in the previous one please?

functests/1_performance/performance.go
shajmakh: my mistake ..fixed it , thanks !

functests/1_performance/performance.go
shajmakh: Thanks, reformatted the code , please let me know if I missed anything.

functests/1_performance/performance.go
```diff
@@ -201,14 +201,14 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
-                         for _, node := range workerRTNodes {
-                                 pid, err := nodes.ExecCommandOnNode([]string{"pgrep", "-f", "stalld"}, &node)
-                                 Expect(err).ToNot(HaveOccurred())
-                                 sched_tasks, _ := nodes.ExecCommandOnNode([]string{"chrt", "-ap", pid}, &node)
-				 Expect(sched_tasks).To(ContainSubstring("scheduling policy: SCHED_FIFO"))
-				 Expect(sched_tasks).To(ContainSubstring("scheduling priority: 10"))
-			 }
-                 })
+			for _, node := range workerRTNodes {
+				pid, err := nodes.ExecCommandOnNode([]string{"pgrep", "-f", "stalld"}, &node)
```
fromanirh, line 205: can ever "pid" be empty string? IOW, are we guaranteed that if `err == nil` THEN the "pid" is valid?

fromanirh (commented):
looks good now. One last comment about the logic and we should be done.

fromanirh:
thanks for the updates! Now let's just squash all the commits in one. This is not something we do all the time, but in this case is warranted.

functests/1_performance/performance.go
shajmakh: after checking this , if I search for non existed service pid :
 automatically : it will exit with status 1 error & the test will fail 
 manually : I redirected the stderr to another file but found nothing 
 So I added a check for the pid to not to be empty after checking the error.

PR #678, opened 2021-07-08
Set maximumLatency variable

Tal-or:
In the last update, setting the ```maximumLatency``` via the environment variables value was skipped (bug).
This update will set the values correctly.
Moreover, this update introduces a new unified environment variable for the maximum latency that will allow the user to use the same environment variable for all tests.
The specific environment variables will have precedence over this unified one.

### DIFF
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -118,7 +118,7 @@ var _ = Describe("[performance] Latency Test", func() {
 		testName := oslatTestName
 
 		BeforeEach(func() {
-			err := checkMaximumEnvironLatencyValue(testName)
+			err := setMaximumLatencyValue(testName)
 			Expect(err).ToNot(HaveOccurred())
 
 			if profile.Spec.CPU.Isolated == nil {
@@ -179,7 +179,7 @@ var _ = Describe("[performance] Latency Test", func() {
 		testName := cyclictestTestName
 
 		BeforeEach(func() {
-			err := checkMaximumEnvironLatencyValue(testName)
+			err := setMaximumLatencyValue(testName)
 			Expect(err).ToNot(HaveOccurred())
 
 			if profile.Spec.CPU.Isolated == nil {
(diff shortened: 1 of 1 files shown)


functests/4_latency/latency.go
```diff
@@ -348,21 +353,30 @@ func extractLatencyValues(testName string, exp string, node *corev1.Node) string
...
 		testName != strings.ToUpper(cyclictestTestName) &&
 		testName != strings.ToUpper(hwlatdetectTestName) {
 		return fmt.Errorf("testName variable has incorrect value %q", testName)
 	}
 
+	var err error
+	unifiedMaxLatencyEnv := os.Getenv("MAXIMUM_LATENCY")
+	if unifiedMaxLatencyEnv != "" {
+		maximumLatency, err = strconv.Atoi(unifiedMaxLatencyEnv)
+		return fmt.Errorf("err: %v the environment variable MAXIMUM_LATENCY has incorrect value %q", err, unifiedMaxLatencyEnv)
```
fromanirh, line 372: not sure this is correct. If we have the var set we parse as int... and then we always fail?

Tal-or: Copied it with the original mistake.
Thanks!

functests/4_latency/latency.go
```diff
@@ -348,22 +353,34 @@ func extractLatencyValues(testName string, exp string, node *corev1.Node) string
...
 	if testName != strings.ToUpper(oslatTestName) &&
 		testName != strings.ToUpper(cyclictestTestName) &&
 		testName != strings.ToUpper(hwlatdetectTestName) {
 		return fmt.Errorf("testName variable has incorrect value %q", testName)
 	}
 
+	var err error
+	unifiedMaxLatencyEnv := os.Getenv("MAXIMUM_LATENCY")
+	if unifiedMaxLatencyEnv != "" {
+		if maximumLatency, err = strconv.Atoi(unifiedMaxLatencyEnv); err != nil {
```
fromanirh, line 371: in general (let's not change this just yet) we should not set non-local variables by side effect. A better way:
```
maximumLatency, err = strconv.Atoi(unifiedMaxLatencyEnv)
if err != nil {
    ...
}
```

fromanirh (commented):
/lgtm
/hold
looks good, but how we can test this before the merge? is it covered by existing tests (which run automatically on CI)?

Tal-or:
> /lgtm
> /hold
> looks good, but how we can test this before the merge? is it covered by existing tests (which run automatically on CI)?

This test should be consumed by the cnf-feature-deploy. I'm not aware to a test framework which test our tests

PR #680, opened 2021-07-09
Add test case id to kernel argument managed_irq flag test case

mrniranjan:
Add the Missing test case id for kernel argument managed_irq flag test case


fromanirh (commented):
/lgtm

PR #686, opened 2021-07-13
Enable the memory manager

cynepco3hahue:
Set the memory manager policy to `Static` and reserve
the memory on the NUMA node 0. The amount of reserved memory
equals `kube-reserved + system-reserved + hard eviction threshold`.

**Note to reviewers:**
I will provide e2e tests under the separate PR once the rebase on top of k8s 1.22 done.

### DIFF
--- a/pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
+++ b/pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
@@ -4,6 +4,9 @@ import (
 	"encoding/json"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+
 	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
 	profile2 "github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/profile"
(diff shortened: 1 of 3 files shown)


pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
```diff
@@ -15,11 +18,13 @@ import (
...
-	defaultKubeReservedMemory   = "500Mi"
-	defaultSystemReservedCPU    = "1000m"
-	defaultSystemReservedMemory = "500Mi"
+	cpuManagerPolicyStatic       = "static"
+	memoryManagerPolicyStatic    = "Static"
+	defaultKubeReservedCPU       = "1000m"
+	defaultKubeReservedMemory    = "500Mi"
+	defaultSystemReservedCPU     = "1000m"
+	defaultSystemReservedMemory  = "500Mi"
+	defaultHardEvictionThreshold = "100Mi"
```
fromanirh, line 27: it would be nice to document (a comment is fine) how we determined this value.
Future PRs should retroactively add explanations for the other existing defaults.

fromanirh (commented):
/lgtm
/hold
to give time to other reviewers

pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
cynepco3hahue: Some interesting insights regarding reserved memory:
- https://docs.openshift.com/container-platform/4.7/nodes/nodes/nodes-nodes-resources-configuring.html
```
kube-reserved | This setting is not used with OpenShift Container Platform. Add the CPU and memory resources that you planned to reserve to the system-reserved setting.
-- | --
system-reserved | This setting identifies the resources to reserve for the node components and system components. The default settings depend on the OpenShift Container Platform and Machine Config Operator versions. Confirm the default systemReserved parameter on the machine-config-operator repository.
```
- setting the default under the machine-config-operator, https://github.com/openshift/machine-config-operator/blob/fbc4d8e46a7746442f4de3651113d2181d458b12/templates/common/_base/files/kubelet-auto-sizing.yaml

@fromanirh Probably we need a separate PR to reflect these findings. WDT?

MarSik:
@cynepco3hahue @fromanirh Should we unblock this if nobody else has comments?

PR #687, opened 2021-07-14
Latency tests: remove a condition which is always false

Tal-or:
**Why this fix is needed:**
An issue was discovered during a local build of the cnf-test image that @shajmakh did.
She built the image from https://github.com/Tal-or/cnf-features-deploy/tree/ship_hwlatdetect_cyclictest and ran the latency test which failed with ```"testName variable has incorrect value..."```
a careful examination of the code shows that this check will always fail, IOW the condition will always be false and it will throw this message and exit the test. 

**How do we fix it**
Removing the faulty check. In the first place, this check was intended to verify some flow which eventually was not written at all.

### DIFF
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -359,12 +359,6 @@ func extractLatencyValues(testName string, exp string, node *corev1.Node) string
 // HWLATDETECT_MAXIMUM_LATENCY: the expected maximum latency for all buckets in us
 // MAXIMUM_LATENCY: unified expected maximum latency for all tests
 func setMaximumLatencyValue(testName string) error {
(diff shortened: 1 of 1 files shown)


cynepco3hahue:
@Tal-or Why does it WIP PR?

Tal-or:
> @Tal-or Why does it WIP PR?

I wanted to test the fix first

fromanirh:
please explain a bit more in the commit message (and then in the PR description)
1. how you find this
2. how this can help
TL;DR: we need the "why"

Tal-or:
> please explain a bit more in the commit message (and then in the PR description)
> 
> 1. how you find this
> 2. how this can help
>    TL;DR: we need the "why"

Done

PR #688, opened 2021-07-14
kubelet: enable podresources feature gate

fromanirh:
We will need this feature to enable TAS.
It's pretty safe to enable - noone is actually consuming
the API at the moment, so we do it early.


pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
```diff
@@ -41,6 +42,9 @@ func New(profile *performancev2.PerformanceProfile) (*machineconfigv1.KubeletCon
 			"cpu":    defaultSystemReservedCPU,
 			"memory": defaultSystemReservedMemory,
 		},
+		FeatureGates: map[string]bool{
```
cynepco3hahue, line 45: It will not work, you will need a FeatureGate resource

fromanirh: "It seems too easy" I was thinking. Thanks for pointing this out.

fromanirh:
we will wait for kube 1.23 and we will re-evaluate

PR #691, opened 2021-07-19
add the OCP label index under metadata annotations

yanirq:
Currently, pipeline will only look for the Label in the image (label under the dockerfile of the bundle)
to know what version should be added in each OCP OLM index image,
however, note that the official olm sdk tool ensure that it is in the annotation
which means that this information might be used for any other purpose in the future and then,
the operator bundle might not be processed accordingly if the dockerfile and annotations are not keep in sync.


fromanirh (approved):
/lgtm

yanirq:
@fromanirh  almost forgot the generator change. now its added as well.

fromanirh (approved):
/lgtm

PR #693, opened 2021-07-22
e2e: fix infra containers reserved CPUs test

cynepco3hahue:
From now, by default the CRI-O will drop infra containers and will leave
them only for some special use cases as pod with `ShareProcessNamespace`
parameter under it.


fromanirh:
/lgtm
/approve
I'm not convinced that using this flag is the best approach, but let's unblock CI and let's re-evaluate later.

cynepco3hahue:
> I'm not convinced that using this flag is the best approach, but let's unblock CI and let's re-evaluate later.

I am not sure as well, but in general dropping, the infra containers will be the default behavior unless a user will override crio configuration.

PR #694, opened 2021-07-22
e2e: replace the oc exec with oc rsh

cynepco3hahue:
Output from the `oc exec` sometimes returns the empty output, one of
tries to fix it, is to replace the exec command with the rsh one.


fromanirh:
/approve
/lgtm
let's try!

PR #696, opened 2021-07-26
e2e: use REST client instead of oc to exec commands

cynepco3hahue:
We have a strange behavior during the call to the `oc rsh` or `oc exec`, sometimes it returns empty stdout without any errors.
As an additional attempt to fix the issue, the PR changes the `oc` exec by REST client exec, it means we get rid of the `oc` tool during the exec call.


cynepco3hahue:
@SchSeba FYI

fromanirh:
I think this PR is an improvement in general anyway, but please write few lines to provide context about why we believe this change can fix the weird behaviour we observed.

cynepco3hahue:
@fromanirh Done

fromanirh (approved):
/lgtm

PR #697, opened 2021-07-26
Bug 1986006: e2e: fix oslat test

cynepco3hahue:
Verify all maximum latency values and not only the first one.
Improve the reg expression logic to return correct result.

### DIFF
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -149,14 +149,14 @@ var _ = Describe("[performance] Latency Test", func() {
 				Skip("no maximum latency value provided, skip buckets latency check")
 			}
 
-			latencies := extractLatencyValues("oslat", `Maximum:\t*\s*(.*)\s*\(us\)`, workerRTNode)
+			latencies := extractLatencyValues("oslat", `Maximum:\t*([\s\d]*)\(us\)`, workerRTNode)
 
 			// under the output of the oslat very often we have one anomaly high value, for example
 			// Maximum:    16543 15 15 14 13 12 12 13 12 12 12 12 12 12 12 12 12 (us)
 			// it still unclear if it oslat bug or the kernel one, but we definitely do not want to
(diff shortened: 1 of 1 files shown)


functests/4_latency/latency.go
```diff
@@ -347,10 +347,10 @@ func extractLatencyValues(testName string, exp string, node *corev1.Node) string
 	maximumRegex, err := regexp.Compile(exp)
 	Expect(err).ToNot(HaveOccurred())
 
-	latencies := maximumRegex.FindStringSubmatch(out)[1]
-	Expect(maximumLatency).ToNot(BeNil())
+	latencies := maximumRegex.FindStringSubmatch(out)
+	Expect(len(latencies)).To(Equal(2))
```
fromanirh, line 351: why 2? should not be like 16 (17?) in the example above?

MarSik: Yeah, this should probably use the same value the test pod used, to be future proof

MarSik: Oh, I think I know why. The regex has two groups. The second one contains the list of the results.

cynepco3hahue: in general, it returns two outputs, the first one is the matched string and the second is sub-group, it looks like
```
["Maximum:       23 32 23 42 19 20 29 19 23 33 11 (us)",  "23 32 23 42 19 20 29 19 23 33 11 "]
```

fromanirh: right, my fault

Tal-or: Will this change apply to hwlat and cyclic as well?

cynepco3hahue: I do not change the logic under the method, I just want to guarantee that we will not fail with the index error once you try to get the right string under `maximumRegex.FindStringSubmatch(out)[1]`

PR #704, opened 2021-07-29
e2e: Automate test plan for latency measurment tools

shajmakh:
Automate the test plan of running test of latency tools for 4.9 epic

### DIFF
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 
 	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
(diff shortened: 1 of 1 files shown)


functests/4_latency/latency.go
```diff
@@ -114,66 +115,66 @@ var _ = Describe("[performance] Latency Test", func() {
...
+				continue
 			}
-		})
-	})
+
+			Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d", curr, maximumLatency)
+		}
+	},
+		Entry("Valid test case #1", 0, true, 10, 10, true),
+		Entry("Valid test case #2", 0, false, 10, 10, true),
```
shajmakh, line 176: Regarding the call here , even if I modified the inner function of describeTable , it will still not update the values of the variable

functests/4_latency/latency.go
```diff
@@ -245,6 +246,18 @@ var _ = Describe("[performance] Latency Test", func() {
 	})
 })
 
+/*
+DescribeTable("Test different scenarios",func(lat_test_delay int, lat_test_run bool, runtime string, max_latency int){
+		os.Setenv("LATENCY_TEST_DELAY",lat_test_delay)
+		os.Setenv("LATENCY_TEST_RUN",lat_test_run)
+		os.Setenv("LATENCY_TEST_RUNTIME",runtime)
+		os.Setenv("MAXIMUM_LATENCY",max_latency)
+		init()
```
fromanirh, line 255: this may be legal (= compiler doesn't complain) but it's almost always a bad idea. I'd just extract the code which set the globals in a function you can call in the latency tests proper - so just call it in `init` and in the test' tests

shajmakh: Yeah I agree.This was just one of my retries, but I'll consider using your idea

PR #705, opened 2021-07-30
e2e: Modify regex for searching rcu{c,b} processes

shajmakh:
The test should compare priorities of processes stalld, ksoftirq and rcuc or rcub

### DIFF
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -259,16 +259,18 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 					testlog.Warning("Skip checking rcu since RT kernel is disabled")
 					return
 				}
-				rcuc_pid, err := nodes.ExecCommandOnNode([]string{"pgrep", "-f", "rcuc", "-n"}, &node)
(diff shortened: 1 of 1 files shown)


functests/1_performance/performance.go
```diff
@@ -253,17 +253,17 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				ksoftirq_prio, err := strconv.Atoi(match[1])
 				Expect(err).ToNot(HaveOccurred())
 
-				rcuc_pid, err := nodes.ExecCommandOnNode([]string{"pgrep", "-f", "rcuc", "-n"}, &node)
+				rcu_pid, err := nodes.ExecCommandOnNode([]string{"pgrep", "-f", "rcu[c,b]", "-n"}, &node)
```
fromanirh, line 256: it would be nice to document when we should expect kernel threads named `rcuc` and when `rcub`. What's the difference?

shajmakh: I added two notes according to these articles :
https://www.kernel.org/doc/html/latest/admin-guide/kernel-per-CPU-kthreads.html 
https://lwn.net/Articles/220677/

MarSik (approved):

PR #706, opened 2021-07-30
Latency test: Replace init() with getters

shajmakh:
In a goal to not cache the values of the environment variables, we replaced the removed init() and replaced its functionality with getters that are being called inside the main Describe

### DIFF
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -27,23 +27,27 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/klog"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 	"k8s.io/utils/pointer"
-)
-
-var (
-	latencyTestDelay   = 0
-	latencyTestRun     = false
-	latencyTestRuntime = "300"
-	maximumLatency     = -1
-	latencyTestCpus    = -1
 )
 
 const (
 	oslatTestName       = "oslat"
 	cyclictestTestName  = "cyclictest"
 	hwlatdetectTestName = "hwlatdetect"
+	defaultTestDelay    = 0
+	defaultTestRun      = false
+	defaultTestRuntime  = "300"
+	defaultMaxLatency   = -1
+	defaultTestCpus     = -1
+)
+
+var (
+	latencyTestDelay   = defaultTestDelay
+	latencyTestRun     = defaultTestRun
+	latencyTestRuntime = defaultTestRuntime
+	maximumLatency     = defaultMaxLatency
+	latencyTestCpus    = defaultTestCpus
 )
 
 // LATENCY_TEST_DELAY delay the run of the binary, can be useful to give time to the CPU manager reconcile loop
@@ -51,42 +55,25 @@ const (
 // LATENCY_TEST_RUN: indicates if the latency test should run
 // LATENCY_TEST_RUNTIME: the amount of time in seconds that the latency test should run
 // LATENCY_TEST_CPUS: the amount of CPUs the pod which run the latency test should request
-func init() {
-	latencyTestRunEnv := os.Getenv("LATENCY_TEST_RUN")
-	if latencyTestRunEnv != "" {
-		if latencyTestRunEnv == "true" {
-			latencyTestRun = true
-		}
-	}
-
-	latencyTestRuntimeEnv := os.Getenv("LATENCY_TEST_RUNTIME")
-	if latencyTestRuntimeEnv != "" {
-		latencyTestRuntime = latencyTestRuntimeEnv
-	}
-
-	latencyTestDelayEnv := os.Getenv("LATENCY_TEST_DELAY")
-	if latencyTestDelayEnv != "" {
-		var err error
-		if latencyTestDelay, err = strconv.Atoi(latencyTestDelayEnv); err != nil {
-			klog.Fatalf("the environment variable LATENCY_TEST_DELAY has incorrect value %q", latencyTestDelayEnv)
-		}
-	}
-
-	latencyTestCpusEnv := os.Getenv("LATENCY_TEST_CPUS")
-	if latencyTestCpusEnv != "" {
-		var err error
-		if latencyTestCpus, err = strconv.Atoi(latencyTestCpusEnv); err != nil {
-			klog.Fatalf("the environment variable LATENCY_TEST_CPUS has incorrect value %q", latencyTestCpusEnv)
-		}
-	}
-}
 
 var _ = Describe("[performance] Latency Test", func() {
 	var workerRTNode *corev1.Node
 	var profile *performancev2.PerformanceProfile
 	var latencyTestPod *corev1.Pod
+	var err error
 
 	BeforeEach(func() {
+		latencyTestRun, err = getLatencyTestRun()
+		Expect(err).ToNot(HaveOccurred())
+
+		latencyTestDelay, err = getLatencyTestDelay()
+		Expect(err).ToNot(HaveOccurred())
+
+		latencyTestCpus, err = getLatencyTestCpus()
+		Expect(err).ToNot(HaveOccurred())
+
+		latencyTestRuntime = getLatencyTestRuntime()
+
 		if !latencyTestRun {
 			Skip("Skip the latency test, the LATENCY_TEST_RUN set to false")
 		}
@@ -95,7 +82,6 @@ var _ = Describe("[performance] Latency Test", func() {
 			Skip("Discovery mode enabled, performance profile not found")
 		}
 
-		var err error
 		profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
 		Expect(err).ToNot(HaveOccurred())
 
@@ -110,7 +96,6 @@ var _ = Describe("[performance] Latency Test", func() {
 	})
 
 	AfterEach(func() {
-		var err error
 		err = testclient.Client.Delete(context.TODO(), latencyTestPod)
 		if err != nil {
 			testlog.Error(err)
@@ -128,7 +113,7 @@ var _ = Describe("[performance] Latency Test", func() {
 		testName := oslatTestName
 
 		BeforeEach(func() {
-			err := setMaximumLatencyValue(testName)
+			maximumLatency, err = getMaximumLatency(testName)
 			Expect(err).ToNot(HaveOccurred())
 
 			if profile.Spec.CPU.Isolated == nil {
@@ -189,7 +174,7 @@ var _ = Describe("[performance] Latency Test", func() {
 		testName := cyclictestTestName
 
 		BeforeEach(func() {
-			err := setMaximumLatencyValue(testName)
+			maximumLatency, err = getMaximumLatency(testName)
 			Expect(err).ToNot(HaveOccurred())
 
 			if profile.Spec.CPU.Isolated == nil {
@@ -225,7 +210,7 @@ var _ = Describe("[performance] Latency Test", func() {
 		testName := hwlatdetectTestName
 
 		BeforeEach(func() {
-			err := setMaximumLatencyValue(testName)
+			maximumLatency, err = getMaximumLatency(testName)
 			Expect(err).ToNot(HaveOccurred())
 		})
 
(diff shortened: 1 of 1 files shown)


shajmakh:
@fromanirh @MarSik

functests/4_latency/latency.go
```diff
@@ -255,6 +228,65 @@ var _ = Describe("[performance] Latency Test", func() {
 	})
 })
 
+//read LATENCY_TEST_RUN environment var and parse it if set
+func getLatencyTestRun() {
+	latencyTestRunEnv := os.Getenv("LATENCY_TEST_RUN")
+	if latencyTestRunEnv != "" {
+		if latencyTestRunEnv == "true" {
+			latencyTestRun = true
```
MarSik, line 236: Thsi is still caching. The getter should just parse the value and return true/false. The `latencyTestRun` variable should be removed

functests/4_latency/latency.go
```diff
@@ -255,6 +228,65 @@ var _ = Describe("[performance] Latency Test", func() {
...
+		}
+	}
+}
+
+//read LATENCY_TEST_RUNTIME environment var and parse it if set
+func getLatencyTestRuntime() {
+	latencyTestRuntimeEnv := os.Getenv("LATENCY_TEST_RUNTIME")
+	if latencyTestRuntimeEnv != "" {
+		latencyTestRuntime = latencyTestRuntimeEnv
```
MarSik, line 245: This is still caching. The getter should just parse the value and return the string. The `latencyTestRuntime` variable should be removed

functests/4_latency/latency.go
```diff
@@ -255,6 +228,65 @@ var _ = Describe("[performance] Latency Test", func() {
...
+}
+
+//read LATENCY_TEST_DELAY environment var and parse it if set
+func getLatencyTestDelay() {
+	latencyTestDelayEnv := os.Getenv("LATENCY_TEST_DELAY")
+	if latencyTestDelayEnv != "" {
+		var err error
+		if latencyTestDelay, err = strconv.Atoi(latencyTestDelayEnv); err != nil {
+			klog.Fatalf("the environment variable LATENCY_TEST_DELAY has incorrect value %q", latencyTestDelayEnv)
+		}
```
MarSik, line 256: Same here, instead of saving the parsed value, return it.

functests/4_latency/latency.go
```diff
@@ -255,6 +228,65 @@ var _ = Describe("[performance] Latency Test", func() {
...
+
+//read LATENCY_TEST_CPUS environment var and parse it if set
+func getLatencyTestCpus() {
+	latencyTestCpusEnv := os.Getenv("LATENCY_TEST_CPUS")
+	if latencyTestCpusEnv != "" {
+		var err error
+		if latencyTestCpus, err = strconv.Atoi(latencyTestCpusEnv); err != nil {
+			klog.Fatalf("the environment variable LATENCY_TEST_CPUS has incorrect value %q", latencyTestCpusEnv)
+		}
+	}
```
MarSik, line 268: Same here, instead of saving the parsed value, return it.

functests/4_latency/latency.go
```diff
@@ -255,6 +228,65 @@ var _ = Describe("[performance] Latency Test", func() {
...
+
+//read MAXIMUM_LATENCY environment var and parse it if set
+func getMaxLatency() {
+	unifiedMaxLatencyEnv := os.Getenv("MAXIMUM_LATENCY")
+	if unifiedMaxLatencyEnv != "" {
+		var err error
+		if maximumLatency, err = strconv.Atoi(unifiedMaxLatencyEnv); err != nil {
+			klog.Fatalf("the environment variable MAXIMUM_LATENCY has incorrect value %q", unifiedMaxLatencyEnv)
+		}
+	}
```
MarSik, line 279: Same here, instead of saving the parsed value, return it.

functests/4_latency/latency.go
```diff
@@ -255,6 +228,65 @@ var _ = Describe("[performance] Latency Test", func() {
...
+}
+
+func readAndParseEnvVar() {
```
MarSik, line 282: This is not needed at all, each test should just call the getter directly

functests/4_latency/latency.go
```diff
@@ -375,13 +407,6 @@ func extractLatencyValues(testName string, exp string, node *corev1.Node) string
 // MAXIMUM_LATENCY: unified expected maximum latency for all tests
 func setMaximumLatencyValue(testName string) error {
```
MarSik, line 408: I think this should be converted into a getter as well to get rid of the `maximumLatency` variable

MarSik (changes requested):

functests/4_latency/latency.go
shajmakh: @MarSik should we still check the condition if the value is not empty?

shajmakh: because the intention in the initial code was to set new value into the vars if the new value is not empty , otherwise to keep it default

fromanirh: ti seems a good opportunity to use `LookupEnv` + `ParseBool`:
```
func getLatencyTestRun() {
        // TODO: log if/where needed
        if envValue, ok := os.LookupEnv("LATENCY_TEST_RUN"); ok {
            if value, err := strconv.ParseBool(envValue); err == nil {
                 return value // this could be true or false, but it explicitely signal the intent of the user
            }    
        }
        // something was wrong or missing, stick with the builtin default
      return false
}
```

functests/4_latency/latency.go
shajmakh: fixed, thanks.

functests/4_latency/latency.go
shajmakh: fixed, thanks.

functests/4_latency/latency.go
shajmakh: fixed, thanks.

functests/4_latency/latency.go
shajmakh: fixed, thanks.

functests/4_latency/latency.go
shajmakh: fixed , thanks .

functests/4_latency/latency.go
shajmakh: fixed , thanks

functests/4_latency/latency.go
```diff
@@ -229,62 +235,74 @@ var _ = Describe("[performance] Latency Test", func() {
...
+func getMaximumLatency(testName string) (parsedValue int, errMsg error) {
+	var err error
+	var val int
 	unifiedMaxLatencyEnv := os.Getenv("MAXIMUM_LATENCY")
 	if unifiedMaxLatencyEnv != "" {
 		var err error
-		if maximumLatency, err = strconv.Atoi(unifiedMaxLatencyEnv); err != nil {
-			klog.Fatalf("the environment variable MAXIMUM_LATENCY has incorrect value %q", unifiedMaxLatencyEnv)
+		if val, err = strconv.Atoi(unifiedMaxLatencyEnv); err != nil {
+			err = fmt.Errorf("the environment variable MAXIMUM_LATENCY has incorrect value %q", unifiedMaxLatencyEnv)
```
Tal-or, line 294: You should return the error here, otherwise, you might override it later, when you're trying to parse the other env variables.

functests/4_latency/latency.go
```diff
@@ -56,8 +55,15 @@ var _ = Describe("[performance] Latency Test", func() {
 	var workerRTNode *corev1.Node
 	var profile *performancev2.PerformanceProfile
 	var latencyTestPod *corev1.Pod
+	var err error
+	latencyTestDelay, err = getLatencyTestDelay()
```
Tal-or, line 59: If I understood correctly, this PR intended to check the environment variables each time for each tool.
As @MarSik mentioned, storing the values here, won't achieve that, because the ```Describe()``` is common to all three test cases.
You can call theses ```get*()``` functions inside each ```Context()``` or set the global variables (```latencyTestDelay```, ```latencyTestRuntime```, etc.)  in the ```BeforeEach()```

functests/4_latency/latency.go
fromanirh: correct. In addition you should wrap errors here: https://blog.golang.org/go1.13-errors
TL;DR: use something like
```
err = fmt.Errorf("the environment variable MAXIMUM_LATENCY has incorrect value %q: %w", unifiedMaxLatencyEnv, err)
```

shajmakh: fixed, thanks.

functests/4_latency/latency.go
shajmakh: fixed, thanks.

functests/4_latency/latency.go
shajmakh: yes good one, I used it. thanks .

functests/4_latency/latency.go
Tal-or: @fromanirh please keep me honest here, but yet we should have a ```return``` statement here. Storing the error in ```err``` won't keep it from being override later on in the function.

fromanirh: it won't indeed

shajmakh: After discussing the matter with Talor, I added a return statement on the first error check.Thanks.

functests/4_latency/latency.go
```diff
@@ -235,45 +253,42 @@ var _ = Describe("[performance] Latency Test", func() {
...
-	latencyTestRunEnv := os.Getenv("LATENCY_TEST_RUN")
-	if latencyTestRunEnv != "" {
-		if latencyTestRunEnv == "true" {
-			val = true
-		}
+	if latencyTestRunEnv, ok := os.LookupEnv("LATENCY_TEST_RUN"); ok {
+		val, err = strconv.ParseBool(latencyTestRunEnv)
 	}
-	return val
+	return val, err
```
fromanirh, line 250: this is good. Even better:
```
func getLatencyTestRun() (bool, error) {
	if latencyTestRunEnv, ok := os.LookupEnv("LATENCY_TEST_RUN"); ok {
		return strconv.ParseBool(latencyTestRunEnv)
	}
	return defaultTestRun, nil
}
```

functests/4_latency/latency.go
```diff
@@ -235,45 +253,42 @@ var _ = Describe("[performance] Latency Test", func() {
...
 func getLatencyTestDelay() (parsedValue int, errMsg error) {
 	var err error
-	var val int
-	latencyTestDelayEnv := os.Getenv("LATENCY_TEST_DELAY")
-	if latencyTestDelayEnv != "" {
+	val := defaultTestDelay
+	if latencyTestDelayEnv, ok := os.LookupEnv("LATENCY_TEST_DELAY"); ok {
 		if val, err = strconv.Atoi(latencyTestDelayEnv); err != nil {
-			err = fmt.Errorf("the environment variable LATENCY_TEST_DELAY has incorrect value %q", latencyTestDelayEnv)
+			err = fmt.Errorf("the environment variable LATENCY_TEST_DELAY has incorrect value %q: %w", latencyTestDelayEnv, err)
```
fromanirh, line 267: this is good. Even better: make it look like `getLatencyTestRuntime`. The skeleton code is the same, it should be doable.

fromanirh (commented):
looks good but please squash all the commits into the first one, so we can actually check and do a final pass.

functests/4_latency/latency.go
shajmakh: @fromanirh  Should we preserve the modified error message below ?
fmt.Errorf("the environment variable LATENCY_TEST_DELAY has incorrect value %q: %w", latencyTestDelayEnv, err)

fromanirh (commented):
codewise looks nice, please check the rebase is correct.
I'll have a final pass later today.

shajmakh:
> codewise looks nice, please check the rebase is correct.
> I'll have a final pass later today.
Thank you, the rebase is correct now.

fromanirh:
the CI failure is not a flake:
```
hack/verify-generated.sh
 M functests/4_latency/latency.go
make: *** [Makefile:297: verify] Error 1
{"component":"entrypoint","error":"wrapped process failed: exit status 2","file":"prow/entrypoint/run.go:80","func":"k8s.io/test-infra/prow/entrypoint.Options.Run","level":"error","msg":"Error executing test process","severity":"error","time":"2021-08-05T07:53:47Z"} 
```

functests/4_latency/latency.go
```diff
@@ -255,6 +240,71 @@ var _ = Describe("[performance] Latency Test", func() {
 	})
 })
 
+//read LATENCY_TEST_RUN environment var and parse it if set
```
fromanirh, line 243: minor: this is not a good godoc comment - but it's not relevant because the function is not exported. A good godoc comment should be in the form
```
// Foobar does...
func Foobar() {}
```
note the comment starts with the same name as the function

fromanirh (commented):
/lgtm

functests/4_latency/latency.go
shajmakh: thanks for the info will consider it next time. I will remove it.

fromanirh: no need, let's consider next time.

shajmakh: got that !

PR #708, opened 2021-08-02
Bug 1967911: provide a way to configure additional kubelet arguments

cynepco3hahue:


api/v2/performanceprofile_types.go
```diff
@@ -70,6 +71,18 @@ type PerformanceProfileSpec struct {
...
+	// To find the specific argument see https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/.
+	// By default, the performance-addon-operator will override:
+	// 1. CPU manager policy
+	// 2. CPU manager reconcile period
+	// 3. Topology manager policy
+	// 4. Reserved CPUs
+	// 5. Memory manager policy
+	// 6. Reserved Memory
+	// Please avoid specifying them and use the relevant API to configure these parameters.
+	AdditionalKubeletArguments *v1beta1.KubeletConfiguration `json:"additionalKubeletArguments,omitempty"`
```
MarSik, line 85: Just a design question. Is it a good idea to expose the arguments like this? What about just allowing the user to provide extra command line args? The disadvantage would be that there will be no validation.. but our code would be simpler.

One additional option might be a bit of a hack, but what about letting the user create a full KubeletConfig-like CR object? That would not pollute the PerfProfile API.

I do not insist on any of those two, I would just like to see those options being discussed.

cynepco3hahue: Regarding the first option, the validations are not a single problem. In general, you can not just simply convert from the `map[string]string` to some format acceptable by `KubeletConfiguration`, so you will need to explicitly check every parameter under the map and convert it to the expected format during OpenShift KubeletConfig generation.

Regarding the second option, I think it is the exact reason why do we have PAO, to make it easier for a user to configure nodes from a single place. Configuration of the additional resource is an option, but it will make it harder for a user to configure it:
1. Like we will need to somehow map it between PP and KubeletConfig-like CR(we already knows that configuring labels and similar stuff can be error-prone)
2. Will it be possible to create a number of KubeletConfig-like CR for a single PP?
I think it will raise more questions during the implementation and together with it some corner cases that I prefer to avoid.

pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
```diff
@@ -6,48 +6,63 @@ import (
...
-		},
-		KubeReserved: map[string]string{
-			"cpu":    defaultKubeReservedCPU,
-			"memory": defaultKubeReservedMemory,
-		},
-		SystemReserved: map[string]string{
-			"cpu":    defaultSystemReservedCPU,
-			"memory": defaultSystemReservedMemory,
-		},
+	kubeletConfig := &kubeletconfigv1beta1.KubeletConfiguration{}
```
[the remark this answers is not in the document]
cynepco3hahue: I can reject the creation of the PP under the validation webhook once the feature gates parameter exists under additional kubelet arguments, will it work for you?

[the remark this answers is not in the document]
fromanirh: yes, but I'd just pass it through for now. Let's save this option for a later date.

api/v2/performanceprofile_types.go
MarSik: To the second point. PAO makes it easier, but we also avoid adding extra fields that would be just a direct proxy to a different kind of object. Think about Tuned overrides. The API has to provide a value to the most of the use cases.

Now, I understand that with KubeletConfig the situation is what it is with no inheritance available. But assuming that gets fixed we will still have to maintain our API for a year at least.

cynepco3hahue: Like tuned override is a hack and it requires a lot of actions from a user side, needed the knowledge article points to the fact that it is not easy to configure. I want to avoid it for kubelet configurations especially when the inheritance or priority is not supported by KubeletConfig. From my perspective, we can leave it as it is for one year.

MarSik: > it requires a lot of actions from a user side, needed the knowledge article points to the fact that it is not easy to configure

That is the point :) It should be an exception.

> From my perspective, we can leave it as it is for one year.

Ok.

fromanirh (approved):
/lgtm
/hold
my comments and concerns have been addressed. Once @MarSik is happy, we can merge.

pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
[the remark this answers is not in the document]
williamcaban: Please, make sure this also supports params like the garbage collection configs on custom KubeletConfig

- https://docs.openshift.com/container-platform/4.8/nodes/nodes/nodes-nodes-garbage-collection.html#nodes-nodes-garbage-collection-configuring_nodes-nodes-configuring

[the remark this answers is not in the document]
cynepco3hahue: In general, it will create any kubeletconfig that you will pass, it will just override parameters that are relevant for the PAO.

docs/kubelet-config-snippets.md
```diff
@@ -0,0 +1,37 @@
+# Kubelet Config Snippet
+
+To avoid KubeletConfig custom resource limitation, we introduced the experimental
+kubelet config snippet annotation `experimental.kubelet.config.snippet`.
```
MarSik, line 4: Hmm I think I would use a simpler name. Something like: kubeletconfig.experimental

docs/kubelet-config-snippets.md
```diff
@@ -0,0 +1,37 @@
...
+### Setting unsafe sysctl
+
+```yaml
+apiVersion: performance.openshift.io/v2
+kind: PerformanceProfile
+metadata:
+  name: unsafe-sysctl-performanceprofile
+  annotations:
+    experimental.kubelet.config.snippet: |
+      {"allowedUnsafeSysctls":["net.core.somaxconn"]}
```
MarSik, line 34: This is text, right? Not a structured dictionary

cynepco3hahue: yep, it the json format for the kubelet config

MarSik (approved):
/lgtm

docs/kubelet-config-snippets.md
```diff
@@ -0,0 +1,37 @@
...
+for the kubelet config snippet that you will provide under the annotation,
+so please make sure to set only parameters specified under the [v1beta1 specification](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/).
+Also, by default, the performance-addon-operator will override:
+
+1. CPU manager policy
+2. CPU manager reconcile period
+3. Topology manager policy
+4. Reserved CPUs
+5. Memory manager policy
+6. Reserved Memory
```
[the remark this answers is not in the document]
MarSik: @cynepco3hahue ^^

PR #710, opened 2021-08-03
[release-4.8] Bug 1989531: Control CPUs amount under latency-test pod

Tal-or:
By having the new LATENCY_TEST_CPUS env variable, the user will be able to specify the amount of CPUs
that will be used by the pod which runs the latency-tests.

The LATENCY_TEST_CPUS env variable is optional, the default value is the total amount of isolated CPUs minus one (this behavior is identical to the prior one).


fromanirh (commented):
/approve
/lgtm
we will need to also bump the deps in cnf-tests later

PR #714, opened 2021-08-05
Bug 1990054: remove log file after test finishes

shajmakh:
Store the latency test log in unique file name with timestamp and when the test finishes print it to the console then remove it.

### DIFF
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -61,8 +61,11 @@ var _ = Describe("[performance] Latency Test", func() {
 	var profile *performancev2.PerformanceProfile
 	var latencyTestPod *corev1.Pod
 	var err error
+	var logName string
 
 	BeforeEach(func() {
+		logName = time.Now().Format("20060102150405")
+
 		latencyTestRun, err = getLatencyTestRun()
 		Expect(err).ToNot(HaveOccurred())
 
@@ -96,6 +99,7 @@ var _ = Describe("[performance] Latency Test", func() {
 	})
 
 	AfterEach(func() {
+		removeLogfile(workerRTNode, logName)
 		err = testclient.Client.Delete(context.TODO(), latencyTestPod)
 		if err != nil {
 			testlog.Error(err)
@@ -135,38 +139,30 @@ var _ = Describe("[performance] Latency Test", func() {
 			oslatArgs := []string{
 				fmt.Sprintf("-runtime=%s", latencyTestRuntime),
 			}
-			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, oslatArgs)
-			createLatencyTestPod(latencyTestPod)
+			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, oslatArgs, logName)
+			createLatencyTestPod(latencyTestPod, workerRTNode, logName)
+			logFileContent := getLogFile(workerRTNode, logName)
 
 			// verify the maximum latency only when it requested, because this value can be very different
 			// on different systems
 			if maximumLatency == -1 {
+				testlog.Info(logFileContent)
 				Skip("no maximum latency value provided, skip buckets latency check")
 			}
 
-			latencies := extractLatencyValues("oslat", `Maximum:\t*([\s\d]*)\(us\)`, workerRTNode)
-
-			// under the output of the oslat very often we have one anomaly high value, for example
-			// Maximum:    16543 15 15 14 13 12 12 13 12 12 12 12 12 12 12 12 12 (us)
-			// it still unclear if it oslat bug or the kernel one, but we definitely do not want to
-			// fail our test on it
-			var anomaly bool
-			for _, lat := range strings.Split(latencies, " ") {
+			latencies := extractLatencyValues(logName, `Maximum:\t*([\s\d]*)\(us\)`, workerRTNode)
+			latenciesList := strings.Split(latencies, " ")
+			for _, lat := range latenciesList {
 				if lat == "" {
 					continue
 				}
-
 				curr, err := strconv.Atoi(lat)
 				Expect(err).ToNot(HaveOccurred())
 
-				// skip the anomaly value
-				if curr > maximumLatency && !anomaly {
-					anomaly = true
-					continue
-				}
+				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d : \n %s", curr, maximumLatency, logFileContent)
 
-				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d", curr, maximumLatency)
 			}
+			testlog.Info(logFileContent)
 		})
 	})
 
@@ -186,16 +182,17 @@ var _ = Describe("[performance] Latency Test", func() {
 			cyclictestArgs := []string{
 				fmt.Sprintf("-duration=%s", latencyTestRuntime),
 			}
-			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, cyclictestArgs)
-			createLatencyTestPod(latencyTestPod)
+			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, cyclictestArgs, logName)
+			createLatencyTestPod(latencyTestPod, workerRTNode, logName)
+			logFileContent := getLogFile(workerRTNode, logName)
 
 			// verify the maximum latency only when it requested, because this value can be very different
 			// on different systems
 			if maximumLatency == -1 {
+				testlog.Info(logFileContent)
 				Skip("no maximum latency value provided, skip buckets latency check")
 			}
-
-			latencies := extractLatencyValues("cyclictest", `# Max Latencies:\t*\s*(.*)\s*\t*`, workerRTNode)
+			latencies := extractLatencyValues(logName, `# Max Latencies:\t*\s*(.*)\s*\t*`, workerRTNode)
 			for _, lat := range strings.Split(latencies, " ") {
 				if lat == "" {
 					continue
@@ -204,8 +201,10 @@ var _ = Describe("[performance] Latency Test", func() {
 				curr, err := strconv.Atoi(lat)
 				Expect(err).ToNot(HaveOccurred())
 
-				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d", curr, maximumLatency)
+				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d : \n %s", curr, maximumLatency, logFileContent)
+
 			}
+			testlog.Info(logFileContent)
 		})
 	})
 
@@ -236,10 +235,14 @@ var _ = Describe("[performance] Latency Test", func() {
 				hwlatdetectArgs = append(hwlatdetectArgs, fmt.Sprintf("-threshold=%d", maximumLatency))
 			}
 
-			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, hwlatdetectArgs)
-			createLatencyTestPod(latencyTestPod)
+			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, hwlatdetectArgs, logName)
+			createLatencyTestPod(latencyTestPod, workerRTNode, logName)
+			logFileContent := getLogFile(workerRTNode, logName)
+
 			// here we don't need to parse the latency values.
 			// hwlatdetect will do that for us and exit with error if needed.
+			testlog.Info(logFileContent)
+
 		})
 	})
 })
@@ -305,7 +308,7 @@ func getMaximumLatency(testName string) (int, error) {
 	return val, err
 }
 
-func getLatencyTestPod(profile *performancev2.PerformanceProfile, node *corev1.Node, testName string, testSpecificArgs []string) *corev1.Pod {
+func getLatencyTestPod(profile *performancev2.PerformanceProfile, node *corev1.Node, testName string, testSpecificArgs []string, logName string) *corev1.Pod {
 	runtimeClass := components.GetComponentName(profile.Name, components.ComponentNamePrefix)
 	testNamePrefix := fmt.Sprintf("%s-", testName)
 	runnerName := fmt.Sprintf("%srunner", testNamePrefix)
@@ -322,7 +325,7 @@ func getLatencyTestPod(profile *performancev2.PerformanceProfile, node *corev1.N
 	latencyTestRunnerArgs := []string{
 		"-logtostderr=false",
 		"-alsologtostderr=true",
-		fmt.Sprintf("-log_file=/host/%s.log", testName),
+		fmt.Sprintf("-log_file=/host/%s.log", logName),
 	}
 
 	latencyTestRunnerArgs = append(latencyTestRunnerArgs, testSpecificArgs...)
@@ -387,7 +390,7 @@ func getLatencyTestPod(profile *performancev2.PerformanceProfile, node *corev1.N
 	}
 }
 
-func createLatencyTestPod(testPod *corev1.Pod) {
+func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string) {
 	err := testclient.Client.Create(context.TODO(), testPod)
 	Expect(err).ToNot(HaveOccurred())
 
@@ -401,13 +404,11 @@ func createLatencyTestPod(testPod *corev1.Pod) {
 	By("Waiting another two minutes to give enough time for the cluster to move the pod to Succeeded phase")
 	podTimeout := time.Duration(timeout + 120)
 	err = pods.WaitForPhase(testPod, corev1.PodSucceeded, podTimeout*time.Second)
-	Expect(err).ToNot(HaveOccurred())
+	Expect(err).ToNot(HaveOccurred(), getLogFile(node, logName))
 }
 
-func extractLatencyValues(testName string, exp string, node *corev1.Node) string {
-	cmd := []string{"cat", fmt.Sprintf("/rootfs/var/log/%s.log", testName)}
-	out, err := nodes.ExecCommandOnNode(cmd, node)
-	Expect(err).ToNot(HaveOccurred())
+func extractLatencyValues(logName string, exp string, node *corev1.Node) string {
+	out := getLogFile(node, logName)
 
 	maximumRegex, err := regexp.Compile(exp)
 	Expect(err).ToNot(HaveOccurred())
@@ -416,4 +417,22 @@ func extractLatencyValues(testName string, exp string, node *corev1.Node) string
 	Expect(len(latencies)).To(Equal(2))
 
 	return latencies[1]
+}
+
+func getLogFile(node *corev1.Node, logName string) string {
+	cmd := []string{"cat", fmt.Sprintf("/rootfs/var/log/%s.log", logName)}
+	out, err := nodes.ExecCommandOnNode(cmd, node)
+	if err != nil {
+		testlog.Error(err)
+	}
+	return out
+}
+
+func removeLogfile(node *corev1.Node, logName string) {
+	cmd := []string{"rm", "-f", fmt.Sprintf("/rootfs/var/log/%s.log", logName)}
+	_, err := nodes.ExecCommandOnNode(cmd, node)
+	if err != nil {
+		testlog.Error(err)
+	}
+
 }


functests/4_latency/latency.go
```diff
@@ -314,11 +320,12 @@ func getLatencyTestPod(profile *performancev2.PerformanceProfile, node *corev1.N
 		cpus := cpuset.MustParse(string(*profile.Spec.CPU.Isolated))
 		latencyTestCpus = cpus.Size() - 1
 	}
-
+	currentTime := time.Now().Format("20060102150405")
```
fromanirh, line 323: do you need to know this file name? would a temporary file (see: https://pkg.go.dev/io/ioutil#TempFile) be good?

functests/4_latency/latency.go
```diff
@@ -413,3 +420,15 @@ func extractLatencyValues(testName string, exp string, node *corev1.Node) string
...
 	return latencies[1]
 }
+
+func removeLogfile(node *corev1.Node) {
+	cmd := []string{"cat", fmt.Sprintf("/rootfs/var/log/%s.log", logName)}
+	out, err := nodes.ExecCommandOnNode(cmd, node)
+	Expect(err).ToNot(HaveOccurred())
+	fmt.Printf("the test's output log is: \n %s \n", out)
+
+	cmd = []string{"rm", "-fr", fmt.Sprintf("/rootfs/var/log/%s.log", logName)}
```
fromanirh, line 430: let's not use `-r` here please unless it is really needed (why?)

functests/4_latency/latency.go
```diff
@@ -144,28 +146,32 @@ var _ = Describe("[performance] Latency Test", func() {
...
+				if err == nil {
+					latenciesInt = append(latenciesInt, curr)
+				}
+			}
+			for _, lat := range latenciesInt {
 
-				// skip the anomaly value
-				if curr > maximumLatency && !anomaly {
+				// skip the anomaly value if not the only value
+				if lat > maximumLatency && !anomaly && len(latenciesInt) > 1 {
```
fromanirh, line 169: shouldn't the `len` check be moved outside the `for` loop?

functests/4_latency/latency.go
```diff
@@ -144,28 +146,32 @@ var _ = Describe("[performance] Latency Test", func() {
...
 					continue
 				}
-
 				curr, err := strconv.Atoi(lat)
-				Expect(err).ToNot(HaveOccurred())
+				if err == nil {
+					latenciesInt = append(latenciesInt, curr)
+				}
+			}
+			for _, lat := range latenciesInt {
```
fromanirh, line 166: why do we need two for loops now?

functests/4_latency/latency.go
```diff
@@ -48,6 +48,7 @@ var (
 	latencyTestRuntime = defaultTestRuntime
 	maximumLatency     = defaultMaxLatency
 	latencyTestCpus    = defaultTestCpus
+	logName            = ""
```
cynepco3hahue, line 51: you can use it as local variable and pass it to all relevant methods, it less error-prone than re-assign the global variable each time

functests/4_latency/latency.go
```diff
@@ -144,28 +146,32 @@ var _ = Describe("[performance] Latency Test", func() {
...
+			latencies := extractLatencyValues(`Maximum:\t*([\s\d]*)\(us\)`, workerRTNode)
 			// under the output of the oslat very often we have one anomaly high value, for example
 			// Maximum:    16543 15 15 14 13 12 12 13 12 12 12 12 12 12 12 12 12 (us)
 			// it still unclear if it oslat bug or the kernel one, but we definitely do not want to
 			// fail our test on it
 			var anomaly bool
-			for _, lat := range strings.Split(latencies, " ") {
+			var latenciesInt []int
+			latenciesList := strings.Split(latencies, " ")
+			for _, lat := range latenciesList {
```
cynepco3hahue, line 157: Can you please explain why do you need it? IIUC you want to be sure that you pass only over ints, but if we have some unexpected output we should improve or regex

functests/4_latency/latency.go
cynepco3hahue: So I am expecting you will have
```
Describe(..., func() {
...
var logName string
...
BeforeEach(func() {
logName = time.Now().Format("20060102150405")
pod := getPod(..., logName)
})

AfterEach(func() {
removeLog(node, logName)
})
})
```

functests/4_latency/latency.go
cynepco3hahue: hm we pass the name to the latency tool, so I think we do not really use temp files(because our wrapper is responsible to create the file)

functests/4_latency/latency.go
```diff
@@ -413,3 +420,15 @@ func extractLatencyValues(testName string, exp string, node *corev1.Node) string
 
 	return latencies[1]
 }
+
+func removeLogfile(node *corev1.Node) {
+	cmd := []string{"cat", fmt.Sprintf("/rootfs/var/log/%s.log", logName)}
+	out, err := nodes.ExecCommandOnNode(cmd, node)
+	Expect(err).ToNot(HaveOccurred())
+	fmt.Printf("the test's output log is: \n %s \n", out)
```
cynepco3hahue, line 428: use testlog

functests/4_latency/latency.go
shajmakh: Done , thanks !

functests/4_latency/latency.go
shajmakh: yes, you are right, I improved this code.

functests/4_latency/latency.go
shajmakh: Done , thanks !

functests/4_latency/latency.go
shajmakh: you are right , we don't. I improved the code .Thanks !

functests/4_latency/latency.go
shajmakh: yes , I improved the code , Thanks!

functests/4_latency/latency.go
shajmakh: it is for recursive deleting, & we actually don't need it indeed, I removed it . Thanks!

functests/4_latency/latency.go
```diff
@@ -61,8 +61,11 @@ var _ = Describe("[performance] Latency Test", func() {
 	var profile *performancev2.PerformanceProfile
 	var latencyTestPod *corev1.Pod
 	var err error
+	var logName string
 
 	BeforeEach(func() {
+		logName = time.Now().Format("20060102150405")
```
Tal-or, line 67: You will have the same ```logName``` for all the tests.
So in case that more than one tool is running, the log file of the first tool will be overridden by the log file of the second tool.

functests/4_latency/latency.go
```diff
@@ -413,3 +425,20 @@ func extractLatencyValues(testName string, exp string, node *corev1.Node) string
...
 }
+
+func removeLogfile(node *corev1.Node, logName string) {
+
+	cmd := []string{"cat", fmt.Sprintf("/rootfs/var/log/%s.log", logName)}
+	out, err := nodes.ExecCommandOnNode(cmd, node)
+	if err != nil {
+		testlog.Error(err)
+	}
+	fmt.Printf("the test's output log is: \n %s \n", out)
```
Tal-or, line 436: In case of test failure, the log will be printed anyway, so now it will be printed twice?
Also, nothing was printed in case of success, now the output will be printed anyway. is that the desired behavior?

shajmakh: "In case of test failure, the log will be printed anyway, so now it will be printed twice?"
will fix that 
  "Also, nothing was printed in case of success, now the output will be printed anyway. is that the desired behavior?"
Following the discussion with @cynepco3hahue, no we don't, we need it outputted to the console just in case of failure

shajmakh: @Tal-or after checking the first part , the log is printed only once

functests/4_latency/latency.go
shajmakh: @Tal-or  logName will be different for each test , since it is initialized in "BeforeEach" & will be deleted at the end of each test in "afterEach"

MarSik:
I only have one suggestion. We might want to debug the test eventually, is there a way to capture the log if needed / asked for?

shajmakh:
> I only have one suggestion. We might want to debug the test eventually, is there a way to capture the log if needed / asked for?

the log will be printed to the console upon failure

functests/4_latency/latency.go
```diff
@@ -144,29 +148,38 @@ var _ = Describe("[performance] Latency Test", func() {
...
 				if lat == "" {
 					continue
 				}
-
 				curr, err := strconv.Atoi(lat)
 				Expect(err).ToNot(HaveOccurred())
 
-				// skip the anomaly value
-				if curr > maximumLatency && !anomaly {
+				if !anomaly {
```
cynepco3hahue, line 168: it does not correct, you want to enter the anomaly `if condition` only when `curr > expected_latency` or we can remove this anomaly check at all in the light of https://issues.redhat.com/browse/CNF-1048

fromanirh:
> > > I only have one suggestion. We might want to debug the test eventually, is there a way to capture the log if needed / asked for?
> > 
> > 
> > the log will be printed to the console upon failure
> 
> Think about a bug in our code that would prevent that. I wonder if we can somehow make the log on the node temporary without deleting it right away.

like... with a tempfile? :) Which we can clean up only on the happy path, exactly to cover this use case?

functests/4_latency/latency.go
```diff
@@ -413,3 +425,21 @@ func extractLatencyValues(testName string, exp string, node *corev1.Node) string
 
 	return latencies[1]
 }
+
+func printLogFile(node *corev1.Node, logName string) string {
```
fromanirh, line 429: nit: this function doesn't `print`. It `get`s.

functests/4_latency/latency.go
```diff
@@ -400,8 +412,8 @@ func createLatencyTestPod(testPod *corev1.Pod) {
 	Expect(err).ToNot(HaveOccurred())
 }
 
-func extractLatencyValues(testName string, exp string, node *corev1.Node) string {
-	cmd := []string{"cat", fmt.Sprintf("/rootfs/var/log/%s.log", testName)}
+func extractLatencyValues(logName string, exp string, node *corev1.Node) string {
+	cmd := []string{"cat", fmt.Sprintf("/rootfs/var/log/%s.log", logName)}
```
fromanirh, line 416: minor: why aren't we using (a renamed) `printLogFile` here?

fromanirh (commented):
minor/nits inline. My only remaining concern is about the filename proper, echoing @MarSik

functests/4_latency/latency.go
Tal-or: @fromanirh @MarSik @cynepco3hahue  
TL;DR - we need to call this function in case of any error.
The current implementation is getting the log only in case the expected latency > actual.
But the case might fail for various other reasons.
For example, in hwlatdetect it would compare the value (inside the tool) and exit with an error if the measured value is bigger than the expected, so the log will never get printed.

cynepco3hahue: works for me

functests/4_latency/latency.go
shajmakh: at this stage we don't apply the check yet, I just save the first value of the latencies list for later, so in case there was only one value we can apply the check on it & not pass the test. however, if the bug is already in absolute, I agree to remove this check.

functests/4_latency/latency.go
shajmakh: I agree with Talor, I'll handle that.

functests/4_latency/latency.go
cynepco3hahue: I think you meant obsolete, lets remove these check at all, it will make test clearer

shajmakh: yes :)

shajmakh:
> > > > I only have one suggestion. We might want to debug the test eventually, is there a way to capture the log if needed / asked for?
> > > 
> > > 
> > > the log will be printed to the console upon failure
> > 
> > 
> > Think about a bug in our code that would prevent that. I wonder if we can somehow make the log on the node temporary without deleting it right away.
> 
> like... with a tempfile? :) Which we can clean up only on the happy path, exactly to cover this use case?

1)this point makes sense, OTOH it is not possible to do that using TempFile because we don't explicitly create the file on the node, but instead, it is done using the pod's YAML which we set the log file name as an argument.
2)If I am not mistaken ,  I don't think that we would have a problem if we had a bug, because it is only in one step that the logfile output would be helpful for debugging, & that will be upon failing to get the "succeeded" status on the pod . only in this case, we will have the log with some info, regardless the result is positive or negative, the log will be printed anyway to the console.

functests/4_latency/latency.go
```diff
@@ -135,38 +139,28 @@ var _ = Describe("[performance] Latency Test", func() {
...
-
 				curr, err := strconv.Atoi(lat)
 				Expect(err).ToNot(HaveOccurred())
 
-				// skip the anomaly value
-				if curr > maximumLatency && !anomaly {
-					anomaly = true
-					continue
-				}
+				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d : \n %s", curr, maximumLatency, getLogFile(workerRTNode, logName))
```
fromanirh, line 160: `getLogFile` should not be called inline like this, in the error message explaining a failed assertion, because it can fail, like we observed recently. Should always be called explicitely, and we need to handle any errors from it explicitely.
Can be fixed in a later PR.

functests/4_latency/latency.go
fromanirh: I'm still not fond of this way of creating log filenames, but we can improve in a later PR.

fromanirh (commented):
/lgtm
/hold
if @cynepco3hahue is happy enough, we can merge.

functests/4_latency/latency.go
shajmakh: I see , ok will note that & fix it later.Thanks !

functests/4_latency/latency.go
```diff
@@ -135,38 +139,28 @@ var _ = Describe("[performance] Latency Test", func() {
...
 
-				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d", curr, maximumLatency)
 			}
+			fmt.Println(getLogFile(workerRTNode, logName))
```
cynepco3hahue, line 163: please use `testlog`

functests/4_latency/latency.go
```diff
@@ -204,8 +198,10 @@ var _ = Describe("[performance] Latency Test", func() {
 				curr, err := strconv.Atoi(lat)
 				Expect(err).ToNot(HaveOccurred())
 
-				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d", curr, maximumLatency)
+				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d : \n %s", curr, maximumLatency, getLogFile(workerRTNode, logName))
+
 			}
+			fmt.Println(getLogFile(workerRTNode, logName))
```
cynepco3hahue, line 204: the same

functests/4_latency/latency.go
```diff
@@ -236,10 +232,12 @@ var _ = Describe("[performance] Latency Test", func() {
 				hwlatdetectArgs = append(hwlatdetectArgs, fmt.Sprintf("-threshold=%d", maximumLatency))
 			}
 
-			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, hwlatdetectArgs)
-			createLatencyTestPod(latencyTestPod)
+			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, hwlatdetectArgs, logName)
+			createLatencyTestPod(latencyTestPod, workerRTNode, logName)
 			// here we don't need to parse the latency values.
 			// hwlatdetect will do that for us and exit with error if needed.
+			fmt.Println(getLogFile(workerRTNode, logName))
```
cynepco3hahue, line 239: and the same

functests/4_latency/latency.go
shajmakh: updated , Thanks !

functests/4_latency/latency.go
shajmakh: updated , thanks !

functests/4_latency/latency.go
shajmakh: updated , thanks !

functests/4_latency/latency.go
shajmakh: since I had to make additional change , I have modified this part of code as well .

functests/4_latency/latency.go
```diff
@@ -61,8 +61,12 @@ var _ = Describe("[performance] Latency Test", func() {
 	var profile *performancev2.PerformanceProfile
 	var latencyTestPod *corev1.Pod
 	var err error
+	var logName string
+	var logFileContent string
```
cynepco3hahue, line 65: Why exactly do we need this one?

functests/4_latency/latency.go
```diff
@@ -135,38 +140,30 @@ var _ = Describe("[performance] Latency Test", func() {
 			oslatArgs := []string{
 				fmt.Sprintf("-runtime=%s", latencyTestRuntime),
 			}
-			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, oslatArgs)
-			createLatencyTestPod(latencyTestPod)
+			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, oslatArgs, logName)
+			createLatencyTestPod(latencyTestPod, workerRTNode, logName)
+			logFileContent = getLogFile(workerRTNode, logName)
```
cynepco3hahue, line 145: you can just `logFileContent := getLogFile(workerRTNode, logName)`

cynepco3hahue: The same true for all other places

functests/4_latency/latency.go
shajmakh: to avoid using calling getLogFile(..) inside the gomega Expect, according to @fromanirh we need to call it & handle it explicitly & not nest it like it was .

functests/4_latency/latency.go
shajmakh: okay i will change it , I thought to make use of it the same as logName.

fromanirh (commented):
/lgtm

Tal-or:
> /retest

AFAIK there is a problem with our CI, so it won't help here.
We should w8 for https://github.com/openshift/machine-config-operator/pull/2718 to be merge

fromanirh:
/retest
if it's really a flake we will override

Tal-or:
I see the following error:
```
• Failure [3.448 seconds]
[ref_id: 40307][pao]Resizing Network Queues
/go/src/github.com/openshift-kni/performance-addon-operators/functests/1_performance/netqueues.go:30
  Updating performance profile for netqueues
  /go/src/github.com/openshift-kni/performance-addon-operators/functests/1_performance/netqueues.go:75
    [test_id:40543] Add interfaceName and verify the interface netqueues are equal to reserved cpus count. [It]
    /go/src/github.com/openshift-kni/performance-addon-operators/functests/1_performance/netqueues.go:127
    Expected
      
        <bool>: false
    to be true
    /go/src/github.com/openshift-kni/performance-addon-operators/functests/1_performance/netqueues.go:163
 ```
 Empty string issue again?

Tal-or:
> I see the following error:
> 
> ```
> • Failure [3.448 seconds]
> [ref_id: 40307][pao]Resizing Network Queues
> /go/src/github.com/openshift-kni/performance-addon-operators/functests/1_performance/netqueues.go:30
>   Updating performance profile for netqueues
>   /go/src/github.com/openshift-kni/performance-addon-operators/functests/1_performance/netqueues.go:75
>     [test_id:40543] Add interfaceName and verify the interface netqueues are equal to reserved cpus count. [It]
>     /go/src/github.com/openshift-kni/performance-addon-operators/functests/1_performance/netqueues.go:127
>     Expected
>       
>         <bool>: false
>     to be true
>     /go/src/github.com/openshift-kni/performance-addon-operators/functests/1_performance/netqueues.go:163
> ```
> 
> Empty string issue again?

The CI keeps falling at this point.
I see there was some changes done in this area https://github.com/openshift-kni/performance-addon-operators/commit/206d982d705ca1c38119d915e0a589c5ea370c5c 
And you don't have it under your branch.
Maybe try rebase and see if it helps.

Tal-or:
@shajmakh try to rebase again now that we have https://github.com/openshift-kni/performance-addon-operators/pull/723 fix and retest.

PR #715, opened 2021-08-09
e2e: wait for the pod output

cynepco3hahue:
Some of the tests under our CI are failing because that pod exec returns an empty output.
We do not know the reason, but we can mitigate a problem by re-running the command on the pod again as a short-term solution.


fromanirh:
/lgtm
The fact we need this fix is unfortunate, but we can't have too many flakes because of this.
I'm starting to think this is an issue in the kube infrastructure (= libraries)

PR #716, opened 2021-08-10
e2e: fix profile UpdateWithRetry method

cynepco3hahue:
We should get the updated version of the performance profile.
Otherwise, we will always try to update the profile with the old resource version
and will always get the conflict error.


fromanirh:
/lgtm
/approve
This seems a good idea in general

PR #717, opened 2021-08-12
Bug:1992710 Latency tests: Pass latencyTestRuntime as an argument to the test

Tal-or:
We need to pass latencyTestRuntime (which is originally the LATENCY_TEST_RUNTIME env variable) as an argument before running the latency test pod.
Without it the user won't be able to control for how long the test will run.

### DIFF
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -183,7 +183,10 @@ var _ = Describe("[performance] Latency Test", func() {
 		})
 
 		It("should succeed", func() {
-			latencyTestPod = getLatencyTestPod(profile, workerRTNode, testName, []string{})
+			cyclictestArgs := []string{
+				fmt.Sprintf("-duration=%s", latencyTestRuntime),
+			}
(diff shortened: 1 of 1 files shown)


functests/4_latency/latency.go
```diff
@@ -183,7 +183,10 @@ var _ = Describe("[performance] Latency Test", func() {
 		})
 
 		It("should succeed", func() {
```
yanirq, line 185: maybe we can elaborate more here : should succeed with values.. should succeed running latency tests with maximum X

fromanirh:
```
 ERRO[2021-08-12T13:16:05Z] Some steps failed:                           
ERRO[2021-08-12T13:16:05Z] 
  * could not run steps: step e2e-gcp failed: "e2e-gcp" post steps failed: "e2e-gcp" pod "e2e-gcp-gather-extra" failed: the pod ci-op-zthrcbcd/e2e-gcp-gather-extra was deleted without completing after 36s (failed containers: )
Link to step on registry info site: https://steps.ci.openshift.org/reference/gather-extra 
```
:\

PR #718, opened 2021-08-16
e2e: improve net queues setup logic

cynepco3hahue:
We should not call to any `Expect` commands outside of `BeforeEach` or `It`, othwerwise
we will affect the ginkgo dry run under cnf-tests.


fromanirh (commented):
/approve
/lgtm
it's nicer and tidier anyway, so it's a win

PR #720, opened 2021-08-23
e2e: Add RT kernel check before checking existence of rcu

shajmakh:
RCU kernel threads are available only on real time kernel nodes.Adding a condition for that in code before checking rcu in the system.

### DIFF
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -252,6 +252,13 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				ksoftirq_prio, err := strconv.Atoi(match[1])
 				Expect(err).ToNot(HaveOccurred())
 
+				if profile.Spec.RealTimeKernel == nil ||
+					profile.Spec.RealTimeKernel.Enabled == nil ||
+					*profile.Spec.RealTimeKernel.Enabled != true {
+					Expect(stalld_prio).To(BeNumerically("<", ksoftirq_prio))
+					testlog.Warning("Skip checking rcu since RT kernel is disabled")
+					return
+				}
 				rcuc_pid, err := nodes.ExecCommandOnNode([]string{"pgrep", "-f", "rcuc", "-n"}, &node)
 				Expect(err).ToNot(HaveOccurred())
 				Expect(rcuc_pid).ToNot(BeEmpty())
@@ -261,8 +268,8 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				rcuc_prio, err := strconv.Atoi(match[1])
 				Expect(err).ToNot(HaveOccurred())
 
-				Expect(stalld_prio).To(BeNumerically("<", ksoftirq_prio))
 				Expect(stalld_prio).To(BeNumerically("<", rcuc_prio))
+				Expect(stalld_prio).To(BeNumerically("<", ksoftirq_prio))
 			}
 		})
 


functests/1_performance/performance.go
```diff
@@ -252,17 +252,21 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
-				rcuc_pid, err := nodes.ExecCommandOnNode([]string{"pgrep", "-f", "rcuc", "-n"}, &node)
-				Expect(err).ToNot(HaveOccurred())
-				Expect(rcuc_pid).ToNot(BeEmpty())
-				sched_tasks, err = nodes.ExecCommandOnNode([]string{"chrt", "-ap", rcuc_pid}, &node)
-				Expect(err).ToNot(HaveOccurred())
-				match = re.FindStringSubmatch(sched_tasks)
-				rcuc_prio, err := strconv.Atoi(match[1])
-				Expect(err).ToNot(HaveOccurred())
-
+				err = nodes.HasPreemptRTKernel(&node)
```
cynepco3hahue, line 255: let's avoid using `nodes.HasPreemptRTKernel(&node)` because it can return errors not only when the RT kernel is disabled
you can just check the profile

```
if profile.Spec.RealTimeKernel == nil ||
    profile.Spec.RealTimeKernel.Enabled == nil ||
    *profile.Spec.RealTimeKernel.Enabled != true {
	testlog.Warning("Skip checking rcu since RT kernel is disabled")
	return
}
```

shajmakh: ok, thanks for letting me know, will change it.

functests/1_performance/performance.go
```diff
@@ -231,7 +231,7 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 				Expect(sched_tasks).To(ContainSubstring("scheduling priority: 10"))
 			}
 		})
-		It("[test_id:42696][crit:medium][vendor:cnf-qe@redhat.com][level:acceptance] Stalld runs in higher priority than ksoftirq and rcu{c,b}", func() {
+		FIt("[test_id:42696][crit:medium][vendor:cnf-qe@redhat.com][level:acceptance] Stalld runs in higher priority than ksoftirq and rcu{c,b}", func() {
```
fromanirh, line 234: Do we still want the Focus here?
Other than that, lgtm

shajmakh: no we don't , just removed it , Thanks !

fromanirh (approved):
/lgtm
/hold
if @cynepco3hahue is happy, we can merge.

functests/1_performance/performance.go
```diff
@@ -252,17 +252,23 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
-				match = re.FindStringSubmatch(sched_tasks)
-				rcuc_prio, err := strconv.Atoi(match[1])
-				Expect(err).ToNot(HaveOccurred())
-
+				if profile.Spec.RealTimeKernel == nil ||
+					profile.Spec.RealTimeKernel.Enabled == nil ||
+					*profile.Spec.RealTimeKernel.Enabled != true {
+					testlog.Warning("Skip checking rcu since RT kernel is disabled")
+					return
+				} else {
```
cynepco3hahue, line 260: you do not need `else` anymore

shajmakh: thanks !

PR #726, opened 2021-09-01
Latency e2e: Add new negative test case and fix regex

shajmakh:
1)Added new negative check that checks runtime as a number greatest than allowed 32bit integer
2)Fixed Skip regular expression
3)Search the executable output as a string instead of bytes avoiding Gomega truncated error.

### DIFF
--- a/functests/5_latency_testing/latency_testing.go
+++ b/functests/5_latency_testing/latency_testing.go
@@ -3,12 +3,14 @@ package __latency_testing
 import (
 	"bytes"
 	"fmt"
+	"math"
 	"os"
 	"os/exec"
 
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+	testlog "github.com/openshift-kni/performance-addon-operators/functests/utils/log"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


functests/5_latency_testing/latency_testing.go
```diff
@@ -92,16 +92,16 @@ var _ = table.DescribeTable("Test latency measurement tools tests", func(testGro
 		if err != nil {
 			fmt.Println(err.Error())
```
cynepco3hahue, line 93: what about handling an error here?

shajmakh: correcting it ..
, thank you!

functests/5_latency_testing/latency_testing.go
```diff
@@ -197,6 +195,7 @@ func getNegativeTests(toolToTest string) []latencyTest {
 
 	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "5", testMaxLatency: "1", outputMsgs: []string{latencyFailureMsg, fail}, toolToTest: toolToTest})
 	testSet = append(testSet, latencyTest{testRun: "yes", testRuntime: "5", testMaxLatency: "1", outputMsgs: []string{incorrectTestRun, fail}, toolToTest: toolToTest})
+	testSet = append(testSet, latencyTest{testRun: "true", testRuntime: "21474836773", outputMsgs: []string{incorrectRuntime, fail}, toolToTest: toolToTest})
```
Tal-or, line 198: If you want 1 above the max can this number be replaced with ```math.MaxInt + 1```

functests/5_latency_testing/latency_testing.go
shajmakh: @cynepco3hahue  actually we can't use gomega Expect on the error here , because we'll be getting an error in the negative checks & that seems to be resulted in the error,& if we use Expect on that error , we'll end up getting the gomega truncated error. We should not fail in case of negative checks.So I will add this expectation under a condition.

cynepco3hahue: I think you can just remove the check for err here, if it is an error it will fail under the positive test check

functests/5_latency_testing/latency_testing.go
```diff
@@ -28,26 +30,27 @@ const (
...
-	negativeCyclictestMaxLatency   = incorrectMsgPart1 + "\"" + cyclictestMaxLatency + "\" has a negative value" + mustBeNonNegativeInt
-	incorrectHwlatdetectMaxLatency = incorrectMsgPart1 + "\"" + hwlatdetecMaxLatency + "\"" + incorrectMsgPart2 + mustBeNonNegativeInt
-	negativeHwlatdetectMaxLatency  = incorrectMsgPart1 + "\"" + hwlatdetecMaxLatency + "\" has a negative value" + mustBeNonNegativeInt
-	incorrectTestRun               = incorrectMsgPart1 + latencyTestRun + incorrectMsgPart2
-	incorrectRuntime               = incorrectMsgPart1 + latencyTestRuntime + incorrectMsgPart2 + mustBePositiveInt
-	nonpositiveRuntime             = incorrectMsgPart1 + latencyTestRuntime + " has a nonpositive value" + mustBePositiveInt
+	incorrectMsgPart1                  = "the environment variable "
+	incorrectMsgPart2                  = " has incorrect value"
+	invalidNumber                      = " has an invalid number"
+	maxInt                             = "2147483647"
```
Tal-or, line 36: Use ```math.MaxInt```
```
maxInt := strconv.Itoa(math.MaxInt)
```
It can't be a constant but I think it's better than this magic number
BTW I see different values when I print it:
https://play.golang.org/p/eh0d8zDF5b-

shajmakh: should be math.MaxInt32

shajmakh: I have discussed the matter of changing this constant to math.MaxInt32 with Artyom , but that will require that all the dependent constants be changed to variables , so we decided to leave that as it is for now.

PR #730, opened 2021-09-13
Makefile: Add latency executable prerequisite for functests

shajmakh:
In order not to skip the tests in 5_latency_testing suite, a prerequisite was added for the functest-only target, that creates the executable test generated from 4_latency, so it will be used in 5_latency_testing.


Tal-or:
The  5_latency_testing suite should run as part of the ```functest-only``` make target?

shajmakh:
> The 5_latency_testing suite should run as part of the `functest-only` make target?
@Tal-or Well, these tests check the functionality of the latency tool tests, and thus running them regularly along the rest of the functests suite will definitely be covering & would be a good assist to detect bugs.
They run already under this target but the tests are being skipped because the load binary does not exist.

PR #732, opened 2021-09-14
cherrypick Netqueues to 4.8 

mrniranjan:


mrniranjan:
> /approve
> but we need a tracking BZ, don't we?
@fromanirh 
Can we use an existing [verified BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1899321) for this, or do we need to create a new Bugzilla.

yanirq:
@mrniranjan  Can the BZ issue be fixed so we can merge this ?

PR #733, opened 2021-09-15
tuned: remove inline comments

fromanirh:
The tuned config parser doesn't really like inline comments
(see:
https://github.com/openshift/cluster-node-tuning-operator/pull/271)

We don't really need inline comments, we can thus trivially rearrange
the comments to convey the same information but keeping them on their
own line. No intended changes of behaviour.


fromanirh:
cc @MarSik

PR #734, opened 2021-09-17
Latency e2e: Check actual CPUs number and runtime

shajmakh:
1)In latency.go, Check if actual CPUs number for latency pod is as set in the environment variable
2)In latency_testing.go: -check if actual runtime of the latency tools are as set in the environment variable  -Add new test case for oslat to check skipping the test incase there is less than 2 cpus

### DIFF
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -33,14 +33,15 @@ import (
 )
 
 const (
-	oslatTestName       = "oslat"
-	cyclictestTestName  = "cyclictest"
-	hwlatdetectTestName = "hwlatdetect"
-	defaultTestDelay    = 0
-	defaultTestRun      = false
-	defaultTestRuntime  = "300"
-	defaultMaxLatency   = -1
-	defaultTestCpus     = -1
+	oslatTestName        = "oslat"
+	cyclictestTestName   = "cyclictest"
+	hwlatdetectTestName  = "hwlatdetect"
+	defaultTestDelay     = 0
+	defaultTestRun       = false
+	defaultTestRuntime   = "300"
+	defaultMaxLatency    = -1
+	defaultTestCpus      = -1
+	minCpuAmountForOslat = 2
 )
 
 var (
@@ -132,8 +133,11 @@ var _ = Describe("[performance] Latency Test", func() {
 			// do not calculated into the Allocated, at least part of time of one of isolated CPUs will be used to run
 			// other node containers
 			// at least two isolated CPUs to run oslat + one isolated CPU used by other containers on the node = at least 3 isolated CPUs
-			if isolatedCpus.Size() < 3 {
-				Skip(fmt.Sprintf("Skip the oslat test, the profile %q has less than two isolated CPUs", profile.Name))
+			if isolatedCpus.Size() < (minCpuAmountForOslat + 1) {
+				Skip(fmt.Sprintf("Skip the oslat test, the profile %q has less than %d isolated CPUs", profile.Name, minCpuAmountForOslat))
+			}
+			if latencyTestCpus < minCpuAmountForOslat && latencyTestCpus != defaultTestCpus {
+				Skip("Skip the oslat test, LATENCY_TEST_CPUS is less than the minimum CPUs amount %d", minCpuAmountForOslat)
 			}
 		})
 
@@ -164,7 +168,8 @@ var _ = Describe("[performance] Latency Test", func() {
 				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d : \n %s", curr, maximumLatency, logFileContent)
 
 			}
-			testlog.Info(logFileContent)
+			//Use Println here so that this output will be displayed upon executing the test binary
+			fmt.Println(logFileContent)
 		})
 	})
 
@@ -206,7 +211,8 @@ var _ = Describe("[performance] Latency Test", func() {
 				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d : \n %s", curr, maximumLatency, logFileContent)
 
 			}
-			testlog.Info(logFileContent)
+			//Use Println here so that this output will be displayed upon executing the test binary
+			fmt.Println(logFileContent)
 		})
 	})
 
@@ -243,8 +249,8 @@ var _ = Describe("[performance] Latency Test", func() {
 
 			// here we don't need to parse the latency values.
 			// hwlatdetect will do that for us and exit with error if needed.
-			testlog.Info(logFileContent)
-
+			//Use Println here so that this output will be displayed upon executing the test binary
+			fmt.Println(logFileContent)
 		})
 	})
 })
@@ -340,7 +346,7 @@ func getLatencyTestPod(profile *performancev2.PerformanceProfile, node *corev1.N
 	runnerName := fmt.Sprintf("%srunner", testNamePrefix)
 	runnerPath := path.Join("usr", "bin", runnerName)
 
-	if latencyTestCpus == -1 {
+	if latencyTestCpus == defaultTestCpus {
 		// we can not use all isolated CPUs, because if reserved and isolated include all node CPUs, and reserved CPUs
 		// do not calculated into the Allocated, at least part of time of one of isolated CPUs will be used to run
 		// other node containers
@@ -428,6 +434,15 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
 	err = pods.WaitForPhase(testPod, corev1.PodRunning, 2*time.Minute)
 	Expect(err).ToNot(HaveOccurred())
 
+	if runtime, _ := strconv.Atoi(latencyTestRuntime); runtime > 1 {
+		By("Checking actual CPUs number for the running pod")
+		limitsCpusQuantity := testPod.Spec.Containers[0].Resources.Limits.Cpu()
+		RequestsCpusQuantity := testPod.Spec.Containers[0].Resources.Requests.Cpu()
+		//letancy pod is guaranteed
+		Expect(isEqual(limitsCpusQuantity, latencyTestCpus)).To(BeTrue(), fmt.Sprintf("actual limits of cpus number used for the latency pod is not as set in LATENCY_TEST_CPUS, actual number is: %s", limitsCpusQuantity))
+		Expect(isEqual(RequestsCpusQuantity, latencyTestCpus)).To(BeTrue(), fmt.Sprintf("actual requests of cpus number used for the latency pod is not as set in LATENCY_TEST_CPUS, actual number is: %s", RequestsCpusQuantity))
+	}
+
 	By("Waiting another two minutes to give enough time for the cluster to move the pod to Succeeded phase")
 	podTimeout := time.Duration(timeout + 120)
 	err = pods.WaitForPhase(testPod, corev1.PodSucceeded, podTimeout*time.Second)
@@ -462,4 +477,8 @@ func removeLogfile(node *corev1.Node, logName string) {
 		testlog.Error(err)
 	}
 
+}
+
+func isEqual(qty *resource.Quantity, amount int) bool {
+	return qty.CmpInt64(int64(amount)) == 0
 }
--- a/functests/5_latency_testing/latency_testing.go
+++ b/functests/5_latency_testing/latency_testing.go
@@ -6,6 +6,7 @@ import (
 	"math"
 	"os"
 	"os/exec"
+	"regexp"
 
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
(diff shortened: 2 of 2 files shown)


functests/4_latency/latency.go
```diff
@@ -135,6 +135,9 @@ var _ = Describe("[performance] Latency Test", func() {
 			if isolatedCpus.Size() < 3 {
 				Skip(fmt.Sprintf("Skip the oslat test, the profile %q has less than two isolated CPUs", profile.Name))
 			}
+			if latencyTestCpus < 2 && latencyTestCpus != defaultTestCpus {
+				Skip("Skip the oslat test, LATENCY_TEST_CPUS is less than two CPUs")
```
fromanirh, line 139: "is less than the minimum CPU amount or different from the default value %v", defeaultTestCpus

It is possible this check should split in two

functests/4_latency/latency.go
```diff
@@ -164,7 +167,8 @@ var _ = Describe("[performance] Latency Test", func() {
 				Expect(curr < maximumLatency).To(BeTrue(), "The current latency %d is bigger than the expected one %d : \n %s", curr, maximumLatency, logFileContent)
 
 			}
-			testlog.Info(logFileContent)
+			//Use Println here so that this output will be displayed upon executing the test binary
+			fmt.Println(logFileContent)
```
fromanirh, line 171: so this is shown inconditionally now? I wonder if this can cause regressions.

functests/5_latency_testing/latency_testing.go
```diff
@@ -61,6 +61,7 @@ const (
 	//skip messages regex
 	skipTestRun    = `Skip the latency test, the LATENCY_TEST_RUN set to false`
 	skipMaxLatency = `no maximum latency value provided, skip buckets latency check`
+	skipCpuNumber  = `Skip the oslat test, LATENCY_TEST_CPUS is less than two CPUs`
```
fromanirh, line 64: "is less than the minimum CPUs value"

functests/utils/pods/pods.go
```diff
@@ -223,3 +224,27 @@ func GetPerformanceOperatorPod() (*corev1.Pod, error) {
 
 	return &pods.Items[0], nil
 }
+
+//GetCpusNumber returns the cpus number that are used by mco pod
```
fromanirh, line 228: by MCO pod or by the pod passed as argument?

functests/utils/pods/pods.go
```diff
@@ -223,3 +224,27 @@ func GetPerformanceOperatorPod() (*corev1.Pod, error) {
...
+func GetCpusNumber(pod *corev1.Pod) (int, error) {
+	cmdGetCPU := []string{"bash", "-c", fmt.Sprintf("cat /sys/fs/cgroup/cpuset/cpuset.cpus")}
+	stdout, err := ExecCommandOnPod(testclient.K8sClient, pod, cmdGetCPU)
+	if err != nil {
+		return -1, fmt.Errorf("unable to get cpus number of the pod : %w", err)
+	}
+	if strings.Contains(string(stdout), "-") {
```
fromanirh, line 235: best to use cpuset.Parse, see examples in functests/1_performance/cpu_management.go.

functests/utils/pods/pods.go
shajmakh: yes by the object pod, will modify that, Thank you for noticing me!

functests/utils/pods/pods.go
```diff
@@ -223,3 +224,27 @@ func GetPerformanceOperatorPod() (*corev1.Pod, error) {
 
 	return &pods.Items[0], nil
 }
+
+//GetCpusNumber returns the cpus number that are used by mco pod
+func GetCpusNumber(pod *corev1.Pod) (int, error) {
+	cmdGetCPU := []string{"bash", "-c", fmt.Sprintf("cat /sys/fs/cgroup/cpuset/cpuset.cpus")}
```
fromanirh, line 230: I'm not sure this path is correct. It seems to lack the cgroup name at very least.

functests/4_latency/latency.go
shajmakh: In fact, this is already covered when we parsed the cpus' number here :
[Here](https://github.com/openshift-kni/performance-addon-operators/blob/master/functests/4_latency/latency.go#L297 
)
So at the point we get to this condition, we have these options :
latencyTestCpus = default value (-1)
latencyTestCpus > -1 
& this condition will be true if the latencyTestCpus is 0 or 1 & only in this case we will skip the test.
However, I agree with changing "two CPUs" to "minimum allowed CPU amount for oslat tool which is 2"

functests/4_latency/latency.go
shajmakh: Yes indeed, it will be printed unconditionally 
If we want to change it back to the original method we can, & then I think that we can workaround it in a way to save the output in a file and verify it from that file. Do you think we may face a complication with this solution?

functests/utils/pods/pods.go
shajmakh: Can you illustrate more? 
I see these groups under cgroup:
[root@oslat-sdzlq cgroup]# ls 
blkio  cpu,cpuacct  cpuset   freezer  memory   net_cls,net_prio  perf_event  rdma
cpu    cpuacct	    devices  hugetlb  net_cls  net_prio		 pids	     systemd

functests/4_latency/latency.go
fromanirh: not sure, if test pass (but all of them including the intended usage with cnf-tests) I'm ok with showing unconditionally.

functests/4_latency/latency.go
```diff
@@ -428,6 +434,16 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
 	err = pods.WaitForPhase(testPod, corev1.PodRunning, 2*time.Minute)
 	Expect(err).ToNot(HaveOccurred())
 
+	if runtime, _ := strconv.Atoi(latencyTestRuntime); runtime > 1 {
+		By("Checking actual CPUs number for the running pod")
+		limitsCpus := testPod.Spec.Containers[0].Resources.Limits.Cpu()
+		RequestsCpus := testPod.Spec.Containers[0].Resources.Requests.Cpu()
```
fromanirh, line 440: this doesn't seem correct, being both `testPod.Spec.Containers[0].Resources.Limits` and `testPod.Spec.Containers[0].Resources.Requests` a `type ResourceList map[ResourceName]resource.Quantity`

functests/4_latency/latency.go
```diff
@@ -428,6 +434,16 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
...
+		//letancy pod is guaranteed
+
+		Expect(fmt.Sprint(limitsCpus)).To(Equal(fmt.Sprint(latencyTestCpus)), fmt.Sprintf("actual limits of cpus number used for the latency pod is not as set in LATENCY_TEST_CPUS, actual number is: %s", limitsCpus))
```
fromanirh, line 443: better to use the methods of a `resource.Quantity`, like `Cmp`

shajmakh: Glad to know that. Modified it. Thanks!

functests/4_latency/latency.go
shajmakh: Updated.Thanks!

functests/4_latency/latency.go
shajmakh: I am not sure I understand what you mean

fromanirh: nevermind :) I mean that unless we know this breaks something, let's keep it like you wrote it.
I don't think this is going to break anything, though.

functests/4_latency/latency.go
```diff
@@ -428,6 +434,17 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
 	err = pods.WaitForPhase(testPod, corev1.PodRunning, 2*time.Minute)
 	Expect(err).ToNot(HaveOccurred())
 
+	if runtime, _ := strconv.Atoi(latencyTestRuntime); runtime > 1 {
+		By("Checking actual CPUs number for the running pod")
+		limitsCpusQuantity := testPod.Spec.Containers[0].Resources.Limits.Cpu()
+		RequestsCpusQuantity := testPod.Spec.Containers[0].Resources.Requests.Cpu()
```
fromanirh, line 440: cool, I just learned about these helpers :)

functests/4_latency/latency.go
```diff
@@ -428,6 +434,17 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
...
+		//letancy pod is guaranteed
+		//CmpInt64 returns 0 if the quantity is equal to y, -1 if the quantity is less than y, or 1 if the quantity is greater than y
+		//func (q *Quantity) CmpInt64(y int64) int{..}
+		Expect(limitsCpusQuantity.CmpInt64(int64(latencyTestCpus))).To(Equal(0), fmt.Sprintf("actual limits of cpus number used for the latency pod is not as set in LATENCY_TEST_CPUS, actual number is: %s", limitsCpusQuantity))
```
fromanirh, line 444: I'd use
```
func isEqual(qty *resource.Quantity, int amount) bool {
    return qty.CmpInt64(int64(amount)) == 0
}
```
note the pointer only because this is what `Cpu()` returns :\

and then
```
Expect(isEqual(limitsCpusQuantity, latencyTestCpus).To(BeTrue(), "actual limits of cpus number used for the latency pod is not as set in LATENCY_TEST_CPUS, actual number is: %s", limitsCpusQuantity)
```
(same for `Requests`)

functests/4_latency/latency.go
```diff
@@ -428,6 +434,17 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
 	err = pods.WaitForPhase(testPod, corev1.PodRunning, 2*time.Minute)
 	Expect(err).ToNot(HaveOccurred())
 
+	if runtime, _ := strconv.Atoi(latencyTestRuntime); runtime > 1 {
+		By("Checking actual CPUs number for the running pod")
+		limitsCpusQuantity := testPod.Spec.Containers[0].Resources.Limits.Cpu()
+		RequestsCpusQuantity := testPod.Spec.Containers[0].Resources.Requests.Cpu()
+		//letancy pod is guaranteed
+		//CmpInt64 returns 0 if the quantity is equal to y, -1 if the quantity is less than y, or 1 if the quantity is greater than y
```
fromanirh, line 442: please remove the comment once we agree about the solution

shajmakh: okay will remove it. I just found it handy to illustrate why we expect it to be 0 (as it returns three values). Thank you!

functests/4_latency/latency.go
shajmakh: sounds better to me, modified it. Thank you!

functests/4_latency/latency.go
shajmakh: ok, will keep an eye on that. Thank you!

functests/5_latency_testing/latency_testing.go
```diff
@@ -102,6 +105,20 @@ var _ = table.DescribeTable("Test latency measurement tools tests", func(testGro
 				testlog.Error(err.Error())
 			}
 			Expect(string(output)).NotTo(MatchRegexp(unexpectedError), "Unexpected error was detected in a positve test")
+			//Check runtime argument in the pod's log only if the tool is expecetd to be executed
+			if strings.Contains(string(output), success) {
```
fromanirh, line 109: `success` is a regex. Using `strings.Contains` seems wrong here. We should use a regex match.

fromanirh (commented):
one last comment and we are good to go.

functests/5_latency_testing/latency_testing.go
shajmakh: Thanks for letting me notice! updated it.

functests/5_latency_testing/latency_testing.go
```diff
@@ -102,6 +105,24 @@ var _ = table.DescribeTable("Test latency measurement tools tests", func(testGro
 				testlog.Error(err.Error())
 			}
 			Expect(string(output)).NotTo(MatchRegexp(unexpectedError), "Unexpected error was detected in a positve test")
+			//Check runtime argument in the pod's log only if the tool is expecetd to be executed
+			res, err := regexp.MatchString(success, string(output))
```
fromanirh, line 109: in golang, we usually name the bool return value like `ok` or `matched` for regexps. Next time, it'll look a bit better like
```
matched, err := regexp.MatchString(success, string(output))
if err != nil {
  testlog.Error(err.Error())
}
if matched {
   /...
}
```

fromanirh (commented):
/approve
/lgtm

functests/5_latency_testing/latency_testing.go
shajmakh: sounds better indeed, I will consider that next time, thank you!

PR #739, opened 2021-09-27
Netqueues Tests: Add function to get performanceProfile name instead of u…

mrniranjan:
Add function to get performanceprofile name instead of using hardcoded profile name "performance"


functests/1_performance/netqueues.go
```diff
@@ -357,3 +361,9 @@ func getRandomDevice(devices map[string][]int) string {
 	}
 	return device
 }
+
+func getPerformanceProfile(profileList *performancev2.PerformanceProfileList) string {
```
fromanirh, line 365: I think we should not overwrite the argument. The function should read like
```
func getPerformanceProfile() string {
        profileList := performancev2.PerformanceProfileList{}
	err := testclient.Client.List(context.TODO(), &profileList)
	return profileList.Items[0].Name
}
```
what if there are more than one profile? or no profiles? should we handle these cases?

mrniranjan: i am not sure either,  but in the case of multiple profiles , we can pick the first profile. If there is no profile i can skip the test ?

fromanirh: With multiple profiles we should pick any (the first returned is fine), but we should add a log line
With no profiles, logging + skipping the test seems a fine thing to do

mrniranjan: okay will update the patch

functests/1_performance/netqueues.go
```diff
@@ -357,3 +361,17 @@ func getRandomDevice(devices map[string][]int) string {
 	}
 	return device
 }
+
+func getPerformanceProfile() string {
```
cynepco3hahue, line 365: We have a helper method that fetches the PP by the node selector, does it not fit good enough?
See - https://github.com/openshift-kni/performance-addon-operators/blob/968c23e1a7216135f9e7a4deb77f8339844b5576/functests/1_performance/hugepages.go#L52

fromanirh (commented):
codewise looks good, have you verified this change works as intended? Besides checking existing e2e tests still pass (for non-regression), I mean.

fromanirh:
> codewise looks good, have you verified this change works as intended? Besides checking existing e2e tests still pass (for non-regression), I mean.

settled offline. Looks good.

PR #740, opened 2021-09-28
[WIP] Add script to collect must-gather

mrniranjan:
This patch adds a script to run Must-gather. 
TODO: Add script to run PPC on the collected must-gather.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -219,6 +219,8 @@ deps-update:
 	go mod tidy && \
 	go mod vendor
 
+
+
 .PHONY: deploy
 deploy: cluster-deploy
 	# TODO - deprecated, will be removed soon in favor of cluster-deploy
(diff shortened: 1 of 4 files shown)


hack/run-must-gather.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -e
+
+# expect oc to be in PATH by default
+OC_TOOL="${OC_TOOL:-oc}"
+# Current Mustgather version
+VERSION="4.9-snapshot"
+# Must gather image url
+MUST_GATHER_IMG="quay.io/openshift-kni/performance-addon-operator-must-gather:$VERSION"
+OUTPUT_DIR="/tmp/mustgather-$VERSION"
```
cynepco3hahue, line 10: maybe it is better to use `/tmp/mustgather-$VERSION-$timestamp`

mrniranjan: @cynepco3hahue okay will make the changes, but does the system on which this directory exists persist for subsequent runs ?,  Because if i am looking a way to pass this directory to another script which will run Performance profile creator to generate the profile.

cynepco3hahue: I am thinking about the case when you will want to test it locally, or under some persistent environment.

mrniranjan:
closing as issue is no longer relevant

PR #741, opened 2021-09-28
gosec

yuvalk:
- add gosec to verify target
- fix CWE-22
- fix all CWE-118 issues
- fix CWE-276
- fix CWE-703
- ignore unchecked os.Unsetenv from gosec reports
- fix CWE-703
- ignore G204 in latency_testing


cynepco3hahue:
Probably we need to set GOARCH and compiler variables under our CI.
```
Golang errors in file: [functests/utils/images]:
  > [line 0 : column 0] - loading files from package "functests/utils/images": could not determine GOARCH and Go compiler
```

hack/gosec.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+which gosec
+if [ $? -ne 0 ]; then
+	echo "Downloading gosec tool"
+	go get -u github.com/securego/gosec/v2/cmd/gosec
+fi
+
+time GO111MODULE=off gosec -conf gosec.conf.json ./...
```
yuvalk, line 9: I've added this `GO111MODULE` to make it work under prow
but I dont know WHY it work (without it, we're seeing `could not determine GOARCH and Go compiler` error )
if anyone can explain...
@fromanirh @cynepco3hahue maybe?

functests/utils/utils.go
```diff
@@ -36,7 +36,7 @@ func ExecAndLogCommandWithStderr(name string, arg ...string) ([]byte, []byte, er
 
 	var stdout bytes.Buffer
 	var stderr bytes.Buffer
-	cmd := exec.CommandContext(ctx, name, arg...)
+	cmd := exec.CommandContext(ctx, name, arg...) // #nosec G204
```
yuvalk, line 39: @fromanirh I'm not actually sure this is a good, 
do you have any idea how to make this from a closed list?

fromanirh: Hardly. The whole purpose of this API is to run arbitrary commands on arbitrary nodes. Having and maintaining a allow list would be quite a big burden. So yes, the main defense here is carefully review HOW we use this API.

hack/gosec.sh
cynepco3hahue: so with the GO111MODULE=on it works in the context of go modules and GO111MODULE=off in the context of GOPATH, so I am assuming something is configured on the PATH level that not configured on the module level(probably because of the CI)

cynepco3hahue: Did you try to run
time GOARCH=<> GOOS=<> gosec -conf gosec.conf.json ./...?

yanirq:
@yuvalk  please close this PR and open a new one on the NTO repo if needed.

PR #743, opened 2021-09-30
Revert "Bug 2006675: Test delay needs units to be valid duration"

Tal-or:
Reverts openshift-kni/performance-addon-operators#736

Agreed to keep the parameter as an Int instead of Duration in order to keep the backward compatibility of oslat tool which is using Int as well.

After this PR will get merged, we'll need to change cyclictest and hwlatdetect runners to accept Int instead of Duration type.


MarSik (approved):
/lgtm

PR #744, opened 2021-09-30
e2e: Add TC to verify hugepages count on different NUMA nodes

shajmakh:
Add two tests to verify:
-Uneven split of hugepages count between 2 numa nodes
-Even split of hugepages count between 2 numa nodes

### DIFF
--- a/functests/2_performance_update/updating_profile.go
+++ b/functests/2_performance_update/updating_profile.go
@@ -15,6 +15,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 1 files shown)


shajmakh:
> No chance it will work under the CI, so please run the test locally to make sure that it is working.

both tests passed on a matching cluster. You can find the result [here](https://docs.google.com/document/d/1vLBwZ5zXzbIQ4p3gpr-1M_9zEITdpYwZZ0Z26BztmqQ/edit?usp=sharing)

functests/2_performance_update/updating_profile.go
```diff
@@ -159,6 +159,86 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 		})
 	})
 
+	Context("Verify hugepages count split on two NUMA nodes", func() {
+		hpSize2M := performancev2.HugePageSize("2M")
+		t := true
```
cynepco3hahue, line 164: nit: you can use pointer library to create pointers without the need to define new variables, see for example https://github.com/openshift-kni/performance-addon-operators/blob/50d3c1067cff0d74267e7a356d163ba09187e702/functests/1_performance/cpu_management.go#L428

PR #746, opened 2021-10-06
Makefile: Add latency executable prerequisite for functests

shajmakh:
1)In order not to skip the tests in 5_latency_testing suite, a prerequisite was added for the functest-only target, that creates the executable test generated from 4_latency, so it will be used in 5_latency_testing.
2)Adjust cnf-tests image version to 4.9

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -249,7 +249,7 @@ cluster-clean:
 functests: cluster-label-worker-cnf functests-only
 
 .PHONY: functests-only
-functests-only:
+functests-only: dist-latency-tests
 	hack/run-functests.sh
 
 .PHONY: functests-latency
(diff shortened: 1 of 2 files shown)


shajmakh:
After further examination, the main reason the test fails on timeout is that it seems to take a **very** long time while downloading container images over the network, thus the pod will stay in Pending phase very long which will exceed the timeout. In a debug attempt, I kept the pods and the namespace & one can notice that the image was eventually downloaded and a container was successfully created, then, the tool was executed & pod phase moved to "Succeeded".
for more details please find pod "hwlatdetect-xkrbt" in this file:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-kni_performance-addon-operators/730/pull-ci-openshift-kni-performance-addon-operators-master-e2e-gcp/1447927779378073600/artifacts/e2e-gcp/gather-extra/artifacts/pods.json

PR #753, opened 2021-10-12
deps: bump

fromanirh:
bump k8s deps to 1.22.3
bump operator-framework/api deps to v0.10.7


deploy/olm-catalog/performance-addon-operator/4.10.0/performance-addon-operator.v4.10.0.clusterserviceversion.yaml
```diff
@@ -130,6 +130,8 @@ metadata:
   namespace: placeholder
 spec:
   apiservicedefinitions: {}
+  cleanup:
```
cynepco3hahue, line 133: What does the cleanup flag mean?

fromanirh: Yeah, I cut a corner here. So let's dive deeper.
The official doc is... inconclusive:
https://docs.openshift.com/container-platform/4.9/rest_api/operatorhub_apis/clusterserviceversion-operators-coreos-com-v1alpha1.html
I'll find better sources

fromanirh: found it: https://github.com/operator-framework/enhancements/pull/46

fromanirh:
```
With spec.cleanup.enabled: true OLM will add the operatorframework.io/cleanup-apis finalizer on the CSV to prevent operator removal before cleanup. Now when the user deletes the CSV, OLM will delete all CRs for the operator's owned CRDs in the target namespaces of that operator. Once the CRs are successfully removed (after the operator clears its finalizers), OLM will clear the finalizer from the CSV to let the CSV and operator be uninstalled.

With spec.cleanup.enabled: false OLM will not run cleanup on operator uninstall, and remove the operatorframework.io/cleanup-apis finalizer if it was already present on the CSV from a previous opt-in.

This is also the default behavior if spec.cleanup is unspecified. This ensures operators that haven't opted-in to cleanup are not subjected to the finalizer and have their CSV lifecycle remain unchanged.
```

It seems to me that `cleanup: false` is the right call *in this PR*. Should we decide to opt-in, we should do in a separate PR.

cynepco3hahue: for more information https://github.com/operator-framework/enhancements/pull/46
The default behavior was false, so we do not change it.

PR #757, opened 2021-10-13
Add AcceleratedStartup mode

lack:
As demonstrated in cnf-features-deploy originally, this pair of systemd
services and associated script unpin systemd, crio, and a few other
critical system processes at boot and shutdown, to allow a more
responsive reboot when CPUs are otherwise restricted.

The default settings will unrestict the processes at startup until
either:
- 10 minutes have passed, or
- The number of running containers has not changed by more than 2%
  within 2 minutes, provided the count is greater than 40

At shutdown, the processes will be unrestricted until either:
- 30 minutes have passed, or
- The shutdown completes

More details and how to override the default values are in
docs/configuration_hotfixes.md

Signed-off-by: Jim Ramsay <jramsay@redhat.com>


lack:
The `ci/prow/ci` lane will start passing after https://github.com/openshift-kni/performance-addon-operators/pull/762 is merged to get the `jq` tool into the build image.

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -232,6 +251,35 @@ func getIgnitionConfig(assetsDir string, profile *performancev2.PerformanceProfi
...
+
+	if profile.Spec.CPU != nil && profile.Spec.CPU.AcceleratedStartup != nil && *profile.Spec.CPU.AcceleratedStartup {
+		script := "accelerated-container-startup"
+		src := filepath.Join(assetsDir, "scripts", fmt.Sprintf("%s.sh", script))
+		mode := 0700
+		scriptPath := getBashScriptPath(script)
+		if err := addFile(ignitionConfig, src, scriptPath, &mode); err != nil {
+			return nil, err
+		}
+		accelerateStartupService, err := getSystemdContent(getAcceleratedStartupUnitOptions(scriptPath))
```
cynepco3hahue, line 265: Why do we need two services?

cynepco3hahue (commented):
We have some plans to move the PAO under the NTO, it can delay the PR merge(mostly because of the API changes).
Also, I have some additional questions:
1. How can we be sure, that some latency-sensitive workload will not start before the script end?
2. How the script recognize if the node is during the shutdown?

We need more eyes to review it. @MarSik @fromanirh

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
lack: One for startup and one for shutdown, each with different env variables driving the process.

lack:
> We have some plans to move the PAO under the NTO, it can delay the PR merge(mostly because of the API changes).

If this is better to do after moving PAO, I'd be happy to wait on it and redo this PR there.

> Also, I have some additional questions:
> 1. How can we be sure, that some latency-sensitive workload will not start before the script end?

A good question.  I think the answer has a couple parts:
- first of all, we don't.  We have a 10-minute timeout and a pod count check that may exit the script early if pods numbers are steady.
- if workloads use guaranteed pods, they should be entirely unaffected, even when the script is active. We periodically refresh our list of "allowed" CPUs from the list in `/var/lib/kubelet/cpu_manager_state` defaultCpuSet. Any CPUs not in this list because they have been guaranteed to a workload will soon be removed from the processes we unpin, leaving those CPUs exclusively to the guaranteed pods while the script continues to run.
- If this becomes a large issue, the script is self-cleaning, always resetting CPUs back when it exits for any reason. A customer could, in theory, set up a mechanism to stop the systemd service just before they begin critical processing to ensure it won't interfere.

> 2. How the script recognize if the node is during the shutdown?

There are two systemd services. One that runs at startup, and a different one at shutdown, with `WantedBy="shutdown.target reboot.target halt.target"`

> We need more eyes to review it. @MarSik @fromanirh

api/v2/performanceprofile_types.go
```diff
@@ -95,6 +95,12 @@ type CPU struct {
 	// Defaults to "true"
 	// +optional
 	BalanceIsolated *bool `json:"balanceIsolated,omitempty"`
+	// AcceleratedStartup toggles an optional addition to systemd on the host that will temporarily remove the "reserved"
```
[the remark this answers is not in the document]
MarSik: Ian, v1 has different defaults for upgrade reasons. Kubernetes API versions are just different views of the same data as we learned and must support the same features.

PR #765, opened 2021-10-19
functests: utils: add and use event reporting

ffromani:
Add a function to fetch the events associated to a object,
and use it into the 4_latency suite to troubleshoot the
pod startup issues.

Also add a trivial tester program.


tools/getevents/getevents.go
```diff
@@ -0,0 +1,46 @@
+package main
```
cynepco3hahue, line 1: Why do we want to have a separate binary to get events?

fromanirh: it was a easy way to test the new code, but I can strip it off if we don't want it

cynepco3hahue: If you have usage for it I do not have a problem, it is just hard for me to imagine situation when I will call this binary instead of `oc get events` :)

cynepco3hahue:
> nice! we should totally explore this option in a followup PR. Maybe let's file an issue to track this?

Works for me

PR #770, opened 2021-10-29
Latency e2e: Verify node has sufficient cpu

shajmakh:
The machines on which CI runs have few cpus (1-2 typically) for cost/scalability reasons.
Our tests want to exclusively allocate 1 or more cpus, which is impossible or impractical.
Trying to run our tests in these machines will lead to false negatives.

To keep the lane reliable, we add a check for worker nodes to verify nodes have sufficient
cpu amount to run the tests. Should not be the case, we skip the test (instead of failing).

### DIFF
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -99,7 +99,15 @@ var _ = Describe("[performance] Latency Test", func() {
 		Expect(err).ToNot(HaveOccurred(), "error looking for the optional selector: %v", err)
 
 		Expect(workerRTNodes).ToNot(BeEmpty())
-		workerRTNode = &workerRTNodes[0]
+
+		//At least one worker node should have cpu.Allocatable greater than the quantity requested by each test, else skip the test
+		workerRTNodesWithSufficientCpu := nodes.GetByCpuAllocatable(workerRTNodes, latencyTestCpus)
+		if len(workerRTNodesWithSufficientCpu) == 0 {
+			Skip("Insufficient cpu to run the test")
+
+		}
+		workerRTNode = &workerRTNodesWithSufficientCpu[0]
+
 	})
 
 	AfterEach(func() {
@@ -452,7 +460,7 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
 		By("Checking actual CPUs number for the running pod")
 		limitsCpusQuantity := testPod.Spec.Containers[0].Resources.Limits.Cpu()
 		RequestsCpusQuantity := testPod.Spec.Containers[0].Resources.Requests.Cpu()
-		//letancy pod is guaranteed
+		//latency pod is guaranteed
 		Expect(isEqual(limitsCpusQuantity, latencyTestCpus)).To(BeTrue(), fmt.Sprintf("actual limits of cpus number used for the latency pod is not as set in LATENCY_TEST_CPUS, actual number is: %s", limitsCpusQuantity))
 		Expect(isEqual(RequestsCpusQuantity, latencyTestCpus)).To(BeTrue(), fmt.Sprintf("actual requests of cpus number used for the latency pod is not as set in LATENCY_TEST_CPUS, actual number is: %s", RequestsCpusQuantity))
 	}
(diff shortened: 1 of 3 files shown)


functests/4_latency/latency.go
```diff
@@ -99,7 +99,20 @@ var _ = Describe("[performance] Latency Test", func() {
 		Expect(err).ToNot(HaveOccurred(), "error looking for the optional selector: %v", err)
 
 		Expect(workerRTNodes).ToNot(BeEmpty())
-		workerRTNode = &workerRTNodes[0]
+
+		//At least one worker node should have cpu.Allocatable greater than the quantity requested by each test, else skip the test
+		for _, wrkNode := range workerRTNodes {
+			allocatableCPU, _ := wrkNode.Status.Allocatable.Cpu().AsInt64()
+			Expect(err).ToNot(HaveOccurred(), "error getting allocatable cpus on the node: %v", err)
```
Tal-or, line 106: Which error you're checking here?

functests/4_latency/latency.go
```diff
@@ -99,7 +99,20 @@ var _ = Describe("[performance] Latency Test", func() {
...
+			if allocatableCPU >= int64(latencyTestCpus) {
+				workerRTNode = &wrkNode
+				break
+			}
+		}
+		if workerRTNode == nil {
```
Tal-or, line 112: Superfluous. we're already checking that it's not empty on line 101

functests/4_latency/latency.go
shajmakh: Actually, there is no error to check, I must have forgotten to remove this line. Thanks for catching that,I removed it

functests/4_latency/latency.go
shajmakh: On line 101 we are checking that the list of the nodes is not empty, while here we are checking that we found a worker node with at least the same CPU quantity to run the test.

functests/5_latency_testing/latency_testing.go
```diff
@@ -105,12 +110,12 @@ var _ = table.DescribeTable("Test latency measurement tools tests", func(testGro
 				testlog.Error(err.Error())
 			}
 			Expect(string(output)).NotTo(MatchRegexp(unexpectedError), "Unexpected error was detected in a positve test")
-			//Check runtime argument in the pod's log only if the tool is expecetd to be executed
-			res, err := regexp.MatchString(success, string(output))
+			//Check runtime argument in the pod's log only if the tool is expected to be executed
+			ok, err := regexp.MatchString(success, string(output))
 			if err != nil {
-				testlog.Error(err.Error())
+				testlog.Info(err.Error())
```
fromanirh, line 116: why `Error` -> `Info`

functests/4_latency/latency.go
```diff
@@ -99,7 +99,19 @@ var _ = Describe("[performance] Latency Test", func() {
 		Expect(err).ToNot(HaveOccurred(), "error looking for the optional selector: %v", err)
 
 		Expect(workerRTNodes).ToNot(BeEmpty())
-		workerRTNode = &workerRTNodes[0]
+
+		//At least one worker node should have cpu.Allocatable greater than the quantity requested by each test, else skip the test
```
fromanirh, line 103: This is a good change. It should be moved on a new function in `functests/utils/nodes/nodes.go` (or like that)

fromanirh (commented):
good, let's improve a bit more

functests/5_latency_testing/latency_testing.go
shajmakh: Well..Because the error would give us information on why the parsing failed which will help in debugging only.
Do you believe we better keep it as Error? (if so, I will consider changing the logging for similar errors)

fromanirh: It depends on the purpose of the message (this is true in general).
As rule of thumb:

- If the condition is expected to pass, so the failure is exceptional, use Error
- If the condition can pass or fail, and the failure is expected and recoverable, use Info

fromanirh (commented):
/approve
/lgtm

PR #771, opened 2021-11-02
E2e tests: do not panic in discovery mode with no profile for netqueues suite

yanirq:
If discovery mode is enabled but no profile is installed the tests currently panic.
Here, we are forcing all the tests to skip if that happens, providing the reason.
This check was missing for the netqueues test suite


fromanirh:
/lgtm
/hold
to let other reviewers chime in

PR #772, opened 2021-11-02
Update PAO resources regression test to account for workload partitioning

dgonyier:
If workload partitioning is enabled on cluster, check `Resources.Requests["management.workload.openshift.io/cores"]`
instead of `Resources.Requests.CPU`

Same case for `Resources.Limits.CPU`


cynepco3hahue:
Ok I saw the same tests failed under different PR - https://github.com/openshift-kni/performance-addon-operators/pull/775
/hold cancel

cynepco3hahue:
e2e-gcp-lane fails because of CPU load balance test
But the upgrade one has some different failure
```

[Fail] [rfe_id:27368][performance] Performance Operator [It] [test_id:38109] Should run on the control plane nodes 
/go/src/github.com/openshift-kni/performance-addon-operators/functests/1_performance/performance.go:86
```

PR #780, opened 2021-11-19
kubelet: cpumanager: use pcpus-only cpumanager option when configuring single-numa-node policy

fromanirh:
When we require the strictest NUMA alignment, we also enable  the cpumanager policy option to require full physical cores only,
a feature available in kube >= 1.23.


fromanirh:
/hold
the PR is feature complete but we need to solve the CI woes, and I want to doublecheck the e2e test.
the PR is reviewable.

fromanirh:
/hold
need to add test to demonstrate we can optional override the defaults using the kubelet arguments

fromanirh:
Install failure caused by
```
minKubeVersion: 1.23.0
```
so we need to wait for OCP to rebase

fromanirh:
fumbled with `git push` again! continuing on https://github.com/openshift-kni/performance-addon-operators/pull/790

PR #781, opened 2021-11-22
Augment gather_sysinfo to collect machineinfo data

marioferh:
gather_sysinfo should collect machineinfo data in the sysinfo.tgz it produces
The data should be stored in /etc/machineinfo/machineinfo.json


tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -44,9 +47,27 @@ func newSnapshotCommand(knitOpts *cmd.KnitOptions) *cobra.Command {
 	// use this to debug container behaviour
 	snap.Flags().IntVar(&opts.sleepTime, "sleep", 0, "amount of seconds to sleep once done, before exit")
 
+	collectMachineinfo()
```
fromanirh, line 50: better to call this function on line 40, just before `makeSnapshot`

tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -44,9 +47,27 @@ func newSnapshotCommand(knitOpts *cmd.KnitOptions) *cobra.Command {
...
+
 	return snap
 }
 
+func collectMachineinfo() {
+	err := os.MkdirAll("/etc/machineinfo/", os.ModePerm)
+	if err != nil {
+		panic(err)
```
fromanirh, line 58: please no panics, always return error

tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -44,9 +47,27 @@ func newSnapshotCommand(knitOpts *cmd.KnitOptions) *cobra.Command {
...
+	}
+	outfile, err := os.Create("/etc/machineinfo/machineinfo.json")
+	if err != nil {
+		panic(err)
+	}
+
+	mih := machineinformer.Handle{
+		Out: outfile,
```
fromanirh, line 66: you need to set the value of `Root` here using `sysfsRoot` from `knitOpts`

tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -149,5 +170,6 @@ func kniExpectedCloneContent() []string {
 		"/sys/class/dmi/id/product_name",
 		"/sys/class/dmi/id/product_sku",
 		"/sys/class/dmi/id/product_version",
+		"/etc/machineinfo/machineinfo.json",
```
fromanirh, line 173: let's keep these sorted alphabetically

tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -44,9 +47,27 @@ func newSnapshotCommand(knitOpts *cmd.KnitOptions) *cobra.Command {
 	// use this to debug container behaviour
 	snap.Flags().IntVar(&opts.sleepTime, "sleep", 0, "amount of seconds to sleep once done, before exit")
 
+	collectMachineinfo()
+
 	return snap
 }
 
+func collectMachineinfo() {
+	err := os.MkdirAll("/etc/machineinfo/", os.ModePerm)
```
fromanirh, line 56: I recommended this path, but on retrospected it seems less and less nice to temper `/etc`, so I'd just create `machineinfo.json` in `/var/log`. This way we save ourselves a `MkdirAll`.

tools/gather-sysinfo/gather-sysinfo.go
marioferh: I don't understand this, what Root is used for?

fromanirh: It is used to find the root of the sysfs. The name is probably a bit misleading, the concept is mixed with the mount point.
Normally, this value is `/sys`. But we cannot just hardcode `/sys`, because this tool is meant to be used inside a container, so the host sysfs is usally not mounted on `/sys`, but in `/host-sys` or something like this.

It is passed to `GetRaw` - https://github.com/openshift-kni/debug-tools/blob/main/pkg/machineinformer/informer.go#L61
Which use the value to initialize a `relocatablesysfs` - https://github.com/openshift-kni/debug-tools/blob/main/pkg/machineinformer/informer.go#L63
Which in turn builds the path using `root` as prefix - https://github.com/openshift-kni/debug-tools/blob/main/pkg/machineinformer/relocatablesysfs.go#L79

Easy example: https://github.com/openshift-kni/debug-tools/blob/main/pkg/machineinformer/relocatablesysfs.go#L85

api/v1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
cynepco3hahue, line 1: so it go1.17 change, I do not have a problem moving forward to 1.17, but it will be under a separate file and include changes to our tools image https://github.com/openshift-kni/performance-addon-operators/blob/c1dfe722119075fae8368ad05f0df85a733f0b91/openshift-ci/Dockerfile.tools#L5

fromanirh: Yes, better to postpone the move to golang 1.17. @marioferh please switch back to golang 1.16.z for the time being

tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -47,25 +51,24 @@ func newSnapshotCommand(knitOpts *cmd.KnitOptions) *cobra.Command {
...
-	}
-	outfile, err := os.Create("/etc/machineinfo/machineinfo.json")
+func collectMachineinfo(knitOpts *cmd.KnitOptions) error {
+	outfile, err := os.Create("/var/log/machineinfo.json")
 	if err != nil {
-		panic(err)
+		return err
 	}
 
+	root := knitOpts.SysFSRoot
```
fromanirh, line 63: could be inlined in line 66, but no big deal

fromanirh (commented):
/lgtm
/hold

must-gather/node-gather/daemonset.yaml
```diff
@@ -64,3 +66,7 @@ spec:
         hostPath:
           path: /var/lib/kubelet/pod-resources
           type: Directory
+      - name: log
```
fromanirh, line 69: ok, this is needed because of this: https://github.com/openshift-kni/performance-addon-operators/blob/master/tools/gather-sysinfo/gather-sysinfo.go#L78

So when we create the archive, the code looks for `root` + path, so in `/host/var/log/machineinfo.json`.

So adding this mount works, but it's unnecessary and actually wrong: we should not write back in the host filesystem.

So we need to change a bit how we declare the paths to add in the archive.

tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -149,5 +171,6 @@ func kniExpectedCloneContent() []string {
 		"/sys/class/dmi/id/product_name",
 		"/sys/class/dmi/id/product_sku",
 		"/sys/class/dmi/id/product_version",
+		"/var/log/machineinfo.json",
```
fromanirh, line 174: let's remove this path from there

and let's append the container-local paths in line 102, *after* the processing done by `chrootFileSpecs`

something like

```
fileSpecs = dedupExpectedContent(fileSpecs, kniContainerLocalCloneContent())
```

this will break a little the (already rarely used) `if opts.dumpList` flow, but it's ok for now.

must-gather/node-gather/daemonset.yaml
```diff
@@ -63,4 +63,4 @@ spec:
       - name: podres
         hostPath:
           path: /var/lib/kubelet/pod-resources
-          type: Directory
+          type: Directory
```
fromanirh, line 66: please remove this change

tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -47,6 +54,21 @@ func newSnapshotCommand(knitOpts *cmd.KnitOptions) *cobra.Command {
 	return snap
 }
 
+func collectMachineinfo(knitOpts *cmd.KnitOptions, opts *snapshotOptions) error {
+	outfile, err := os.Create("/var/log/machineinfo.json")
```
fromanirh, line 58: please make this a private named constant and reference it here

tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -78,6 +100,9 @@ func makeSnapshot(cmd *cobra.Command, knitOpts *cmd.KnitOptions, opts *snapshotO
 		fileSpecs = chrootFileSpecs(fileSpecs, opts.rootDir)
 	}
 
+	knitContainerLocalContent := []string{"/var/log/machineinfo.json"}
```
fromanirh, line 103: ...and here

fromanirh (commented):
ok, looks good now. A couple minor comments and we are done.
Please file a github issue about adding e2e tests for this code.

go.mod
```diff
@@ -10,10 +10,10 @@ require (
 	github.com/coreos/ignition/v2 v2.9.0
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
 	github.com/google/go-cmp v0.5.6
-	github.com/jaypipes/ghw v0.7.1-0.20210419135914-b8b1e31b27f5
+	github.com/jaypipes/ghw v0.8.1-0.20210605191321-eb162add542b
```
fromanirh, line 13: next time, please split updates of go.mod, go.sum and vendor in its own separate commit

tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -47,6 +52,25 @@ func newSnapshotCommand(knitOpts *cmd.KnitOptions) *cobra.Command {
 	return snap
 }
 
+// intentionally ignore errors, keep collecting data
+// machineinfo data is accessory, if we fail to collect
+// we want to keep going and try to collect /proc and /sys data,
+// which is more important
```
fromanirh, line 58: this comment is misplaced...

tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -86,6 +110,12 @@ func makeSnapshot(cmd *cobra.Command, knitOpts *cmd.KnitOptions, opts *snapshotO
 		scratchDir = filepath.Join(scratchDir, opts.rootDir)
 	}
 
+	localPath := filepath.Join(scratchDir, machineInfoFilePath)
+	err = collectMachineinfo(knitOpts, localPath)
```
fromanirh, line 114: should actually be there

tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -86,6 +110,12 @@ func makeSnapshot(cmd *cobra.Command, knitOpts *cmd.KnitOptions, opts *snapshotO
...
+	if err != nil {
+		return err
```
fromanirh, line 116: and to make the code consistent with the comment, we should just do
```
if err != nil {
    log.Printf("error collecting machine info data: %v - continuing", err)
}
```

must-gather/node-gather/daemonset.yaml
```diff
@@ -64,3 +64,4 @@ spec:
         hostPath:
           path: /var/lib/kubelet/pod-resources
           type: Directory
```
fromanirh, line 66: if you re-upload, please remove this change - it's unnecessary

fromanirh (commented):
we only need a fix in the error handling code and that's it.

tools/gather-sysinfo/gather-sysinfo.go
marioferh: done

tools/gather-sysinfo/gather-sysinfo.go
marioferh: done

tools/gather-sysinfo/gather-sysinfo.go
marioferh: done

must-gather/node-gather/daemonset.yaml
marioferh: Done, sorry, its was a copy paste issue

PR #785, opened 2021-11-23
add 'pao' to csv keywords

yuvalk:


yuvalk:
/retest
looks like a service was down during the test

fromanirh:
but why? I mean, I don't think "pao" as shortcut for "performance-addon-operator" became widespread general knowledge

yuvalk:
> but why? I mean, I don't think "pao" as shortcut for "performance-addon-operator" became widespread general knowledge

I was sitting next to someone yesterday who actually tried to find it that way :-)
so, why not have it there?

ppl do use the phrase PAO..
it'll just make life easier

fromanirh:
> > but why? I mean, I don't think "pao" as shortcut for "performance-addon-operator" became widespread general knowledge
> 
> I was sitting next to someone yesterday who actually tried to find it that way :-) so, why not have it there?
> 
> ppl do use the phrase PAO.. it'll just make life easier

we, people working in this area/field we likely do, but I simply have no evidence anyone else does, so I'm not convinced it's useful, rather than confusing, for the general audience which is the expected audience of the search keywords.

I'm not sold on this change, but I won't oppose it either.

cynepco3hahue:
I also do not have a strong opinion here, but initially, I had the same thoughts as Francesco. In general, keywords are needed to find operators used for some specific tasks, so if you do not know about the existence of a performance-addon operator no one will search for keyword pao. But I also can not see any harm in it, so if it helps to you I will not opose.

PR #788, opened 2021-11-25
Bug 2034158: [release-4.8]/Netqueues-E2E Tests  Do not panic in discovery mode with no profile

mrniranjan:

### DIFF
--- a/functests/1_performance/netqueues.go
+++ b/functests/1_performance/netqueues.go
@@ -21,6 +21,7 @@ import (
 	testutils "github.com/openshift-kni/performance-addon-operators/functests/utils"
 	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/cluster"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/discovery"
 	testlog "github.com/openshift-kni/performance-addon-operators/functests/utils/log"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/nodes"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/pods"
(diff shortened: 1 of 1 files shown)


functests/1_performance/netqueues.go
```diff
@@ -53,6 +54,9 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", func() {
 	})
 
 	BeforeEach(func() {
+		if discovery.Enabled() && testutils.ProfileNotFound {
+			Skip("Discovery mode enabled, performance profile not found")
+		}
 		profile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
 		Expect(err).ToNot(HaveOccurred())
 		if profile.Spec.Net == nil {
```
cynepco3hahue, line 62: in general, I am unsure we should run such tests under the `1_performance` test suite, like in case when we will update a customer performance profile and for some reason, the rollback will fail, we can get a lot of troubles.

I am proposing to skip the test if the profile does not have some specific net fields and set fields under the performance profile used for CI.

It probably should not be part of this PR, but @mrniranjan can you please open the following PR?

mrniranjan: @cynepco3hahue  i can move the netqueues to functests/2_performance_update. if that's okay ? In this automation we tests the network fields . So it can't be  set in performance profile before hand.

cynepco3hahue: it will work for me, but still we need to have something under 1_performance test

mrniranjan: @cynepco3hahue can you approve this PR, for splitting the netqueues, i will open a new PR .

mrniranjan:
closing it as the bz is closed

PR #790, opened 2021-12-01
kubelet: cpumanager: add pcpus-only

fromanirh:
When we require the strictest NUMA alignment, we also enable
the cpumanager policy option to require full physical cores only,
a feature available in kube >= 1.23.

### DIFF
--- a/pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
+++ b/pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
@@ -28,6 +28,7 @@ const (
 	// Please avoid specifying them and use the relevant API to configure these parameters.
 	experimentalKubeletSnippetAnnotation = "kubeletconfig.experimental"
 	cpuManagerPolicyStatic               = "static"
+	cpuManagerPolicyOptionFullPCPUsOnly  = "full-pcpus-only"
 	memoryManagerPolicyStatic            = "Static"
 	defaultKubeReservedMemory            = "500Mi"
 	defaultSystemReservedMemory          = "500Mi"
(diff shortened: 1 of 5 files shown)


pkg/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
```diff
@@ -116,6 +117,17 @@ func New(profile *performancev2.PerformanceProfile, profileMCPLabels map[string]
 						},
 					}
 				}
+
+				// require full physical CPUs only to ensure maximum isolation
+				if topologyPolicy == kubeletconfigv1beta1.SingleNumaNodeTopologyManagerPolicy {
```
MarSik, line 122: We need to make sure this works with the kubeletconfig.experimental annotation.

fromanirh: I don't recall this annotation, could you please add a link?

fromanirh: it was: https://github.com/openshift-kni/performance-addon-operators/blob/master/docs/kubelet-config-snippets.md

fromanirh: so, we can force the option off if needed using the annotation:
https://github.com/openshift-kni/performance-addon-operators/pull/790/commits/c91534895179be901561bf4e49a105139991978e#diff-0f0074e84690443303677c9a59e7c67616cb32d197bdec89eeb9bae667b05d83R153

fromanirh:
closed by mistake (`git push -d` instead of `git push -f`) continuing in https://github.com/openshift-kni/performance-addon-operators/pull/830

PR #792, opened 2021-12-07
Set isolated and required CPU parameter in CRD

marioferh:
## Description

Set isolated and reserved to required CPU parameters in CRD

Bug 4.7: https://bugzilla.redhat.com/show_bug.cgi?id=1986681

## Type of change

Documentation fix

## Testing steps
In performance_profile.yaml:

  cpu:
    isolated: "1-3"
    ##reserved: "0" <----       Remove reserved

Test of performanceProfile crashes in 4.7 

Besides that, must be compulsory in every Release after 4.7

### DIFF
--- a/api/v2/performanceprofile_types.go
+++ b/api/v2/performanceprofile_types.go
@@ -78,7 +78,7 @@ type CPUSet string
 // CPU defines a set of CPU related features.
 type CPU struct {
 	// Reserved defines a set of CPUs that will not be used for any container workloads initiated by kubelet.
-	Reserved *CPUSet `json:"reserved,omitempty"`
+	Reserved *CPUSet `json:"reserved"`
 	// Isolated defines a set of CPUs that will be used to give to application threads the most execution time possible,
 	// which means removing as many extraneous tasks off a CPU as possible.
 	// It is important to notice the CPU manager can choose any CPU to run the workload
--- a/config/crd/bases/performance.openshift.io_performanceprofiles.yaml
+++ b/config/crd/bases/performance.openshift.io_performanceprofiles.yaml
@@ -494,6 +494,7 @@ spec:
                     type: string
                 required:
                 - isolated
+                - reserved
                 type: object
               globallyDisableIrqLoadBalancing:
                 description: GloballyDisableIrqLoadBalancing toggles whether IRQ load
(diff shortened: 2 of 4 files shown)


yanirq:
replace CDR with CRD in description and title

yanirq:
The files you are changing here (manifests) are actually an outcome of the PAO generators.
These manifests are created by https://github.com/openshift-kni/performance-addon-operators/blob/1e11d22699f63140f80ea1f6c34441e1ad8b83e6/Makefile#L210-L211 
Skimming over the code and docs it might be a good idea to follow the trail of that target and learn how the manifests are generated. 
I would also examine https://github.com/openshift-kni/performance-addon-operators/blob/1e11d22699f63140f80ea1f6c34441e1ad8b83e6/Makefile#L205-L207 
To learn how the documents are generated.
You might need adjustments in both fronts.

fromanirh:
> ## Description
> 
> Set isolated and reserved to required CPU parameters in CDR

Typo: "CRD" (Custom Resource Definition)
Btw what's the rationale of this change?

marioferh:
> > ## Description
> > Set isolated and reserved to required CPU parameters in CDR
> 
> Typo: "CRD" (Custom Resource Definition) Btw what's the rationale of this change?

There is a crash in 4.7 when not CPU Reserved field is set. 
Reserved should be mandatory.

fromanirh:
> > Btw what's the rationale of this change?
> 
> There is a crash in 4.7 when not CPU Reserved field is set. Reserved should be mandatory.

thanks. This is surprising though, do we have a BZ?

fromanirh:
xref: https://github.com/openshift-kni/performance-addon-operators/pull/467#discussion_r544544381
xref: https://github.com/openshift-kni/performance-addon-operators/commit/dbec255b975aa60ac6ee2b25ec6f07a29fca5ba3

marioferh:
> > > Btw what's the rationale of this change?
> > 
> > 
> > There is a crash in 4.7 when not CPU Reserved field is set. Reserved should be mandatory.
> 
> thanks. This is surprising though, do we have a BZ?

https://bugzilla.redhat.com/show_bug.cgi?id=1986681

fromanirh:
> > > > Btw what's the rationale of this change?
> > > 
> > > 
> > > There is a crash in 4.7 when not CPU Reserved field is set. Reserved should be mandatory.
> > 
> > 
> > thanks. This is surprising though, do we have a BZ?
> 
> https://bugzilla.redhat.com/show_bug.cgi?id=1986681

ok, thanks, this wraps it up. Let's indeed update the docs and make sure the code doesn't crash with nil pointer deference anyway (the latter can require a 4.7-only fix). And also please make sure to mention the bug in both the PR initial message/cover letter *and* in the commit message.

PR #797, opened 2021-12-14
controller: watch tuned profiles for changes under the status

cynepco3hahue:
Watching tuned profiles will allow us to quickly reflect the tuned profile
status under the performance profile.

### DIFF
--- a/controllers/performanceprofile_controller.go
+++ b/controllers/performanceprofile_controller.go
@@ -104,6 +104,19 @@ func (r *PerformanceProfileReconciler) SetupWithManager(mgr ctrl.Manager) error 
 		},
 	}
 
+	tunedProfilePredicates := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if !validateUpdateEvent(&e) {
+				return false
+			}
+
+			tunedProfileOld := e.ObjectOld.(*tunedv1.Profile)
+			tunedProfileNew := e.ObjectNew.(*tunedv1.Profile)
+
+			return !reflect.DeepEqual(tunedProfileOld.Status.Conditions, tunedProfileNew.Status.Conditions)
+		},
+	}
+
 	err := ctrl.NewControllerManagedBy(mgr).
 		For(&performancev2.PerformanceProfile{}).
 		Owns(&mcov1.MachineConfig{}, builder.WithPredicates(p)).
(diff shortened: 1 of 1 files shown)


controllers/performanceprofile_controller.go
```diff
@@ -87,7 +87,8 @@ func (r *PerformanceProfileReconciler) SetupWithManager(mgr ctrl.Manager) error
 			kubeletOld := e.ObjectOld.(*mcov1.KubeletConfig)
 			kubeletNew := e.ObjectNew.(*mcov1.KubeletConfig)
 
-			return !reflect.DeepEqual(kubeletOld.Status.Conditions, kubeletNew.Status.Conditions)
+			return kubeletOld.GetGeneration() != kubeletNew.GetGeneration() ||
```
yanirq, line 90: how did this sneak in this PR?

controllers/performanceprofile_controller.go
```diff
@@ -156,6 +175,36 @@ func (r *PerformanceProfileReconciler) mcpToPerformanceProfile(mcpObj client.Obj
 	return requests
 }
 
+func (r *PerformanceProfileReconciler) tunedProfileToPerformanceProfile(tunedProfileObj client.Object) []reconcile.Request {
+	node := &corev1.Node{}
```
yanirq, line 179: Did you mean getting Node here ?  not the profile itself?

controllers/performanceprofile_controller.go
```diff
@@ -156,6 +175,36 @@ func (r *PerformanceProfileReconciler) mcpToPerformanceProfile(mcpObj client.Obj
...
+	profiles := &performancev2.PerformanceProfileList{}
+	if err := r.List(context.TODO(), profiles); err != nil {
+		klog.Error("failed to get performance profiles")
+		return nil
+	}
+
+	var requests []reconcile.Request
+	for i, profile := range profiles.Items {
+		profileNodeSelector := labels.Set(profile.Spec.NodeSelector)
+		nodeLabels := labels.Set(node.Labels)
```
yanirq, line 199: Can we have a situation here where we dont have a 1:1 match ? 
If Matches does not do so, is a contains labels check more suitable here ?

controllers/performanceprofile_controller.go
cynepco3hahue: Just fixed on the way, because the logic is not correct, do you want a separate commit or PR for it?

controllers/performanceprofile_controller.go
cynepco3hahue: yep, the profile will not help us to fetch the relevant performance profile, but because the tuned profile has the same name as node we do not have a problem

controllers/performanceprofile_controller.go
cynepco3hahue: For sure we have such a situation, the node can have a lot of different labels but we should match one that is specified under the node selector?

controllers/performanceprofile_controller.go
fromanirh: please move this (important) fix in its own commit. Please make sure to (briefly) explain the fix in the commit message.

controllers/performanceprofile_controller.go
yanirq: if this is the functionality here then we are good.  just make sure its not the other way around

controllers/performanceprofile_controller.go
cynepco3hahue: Moved to the separate PR - https://github.com/openshift-kni/performance-addon-operators/pull/806

PR #803, opened 2021-12-15
Pin ginkgo version in Dockerfile.tools

jlojosnegros:
The build system is trying to download and compile the latest ginkgo
version, which is too new and doesn't work anymore with the golang
version we use

### DIFF
--- a/openshift-ci/Dockerfile.tools
+++ b/openshift-ci/Dockerfile.tools
@@ -6,6 +6,7 @@ ENV GOVERSION=1.16.3
 ENV GOPATH=/go
 ENV GOBIN=${GOPATH}/bin
 ENV PATH=${PATH}:${GOROOT}/bin:${GOBIN}
+ENV GINKGO_VERSION v1.12.1
 
 ARG GO_PACKAGE_PATH=github.com/openshift-kni/performance-addon-operators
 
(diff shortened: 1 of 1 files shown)


openshift-ci/Dockerfile.tools
```diff
@@ -6,6 +6,8 @@ ENV GOVERSION=1.16.3
 ENV GOPATH=/go
 ENV GOBIN=${GOPATH}/bin
 ENV PATH=${PATH}:${GOROOT}/bin:${GOBIN}
+ENV GO111MODULE=on
+ENV GINKGO_VERSION v1.12.1
```
yanirq, line 10: why is the convention here is different from the rest ?
`ENV GINKGO_VERSION v1.12.1`
and not `ENV GINKGO_VERSION=v1.12.1 ` ? 
(might not have a negative affect but still)

jlojosnegros: I have just used the same line as in the previous branches, so no strong reason so far.
Will change to keep it homogeneous.

fromanirh: Well, it is a bit annoying indeed but it's cosmetic more than functional: https://docs.docker.com/engine/reference/builder/#env (see: "alternatyve syntax")
next time let's be consistent with the surrounding code, but it's not a big deal (and PAO is already too messy already; OTOH that's not a reason to increase the mess...)

fromanirh:
/override ci/prow/e2e-gcp

overrides must be used sparingly. In this case it's ok because we are 100% sure this specific change cannot possibly affect production code, thus causing test failures.

PR #804, opened 2021-12-15
Pin ginkgo version in Dockerfile.tools

jlojosnegros:
The build system is trying to download and compile the latest ginkgo
version, which is too new and doesn't work anymore with the golang
version we use


fromanirh:
/override ci/prow/e2e-gcp
known issue being handled on separate PRs.
This change is not relevant.

PR #812, opened 2021-12-17
Bump deps k8s to 1.23

fromanirh:
bump kube deps to 1.23.

This exposed a latent bug in how we use the
cpuset.Builder.Result() method.

Per doc, the intent of the method was to finalize the builder
and to return the built cpuset. Any subsequent call of Add()
was expected to be a NOP.

Due to a previous bug, this behaviour was not enforced, so
calls to Add() after Result(), albeit wrong, were working.

We use like this in our code because we want to accumulate CPUs
only up a threshold. So the only option is to carry a separate
counter, which allows us to efficiently accumulate as much as
we need and to call Result() exactly once, as it should be.


go.mod
```diff
@@ -98,81 +98,81 @@ require (
...
+	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.11.0
+	sigs.k8s.io/controller-tools => sigs.k8s.io/controller-tools v0.6.2
 )
 
 // Other pinned deps
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v14.2.0+incompatible
 	github.com/coreos/prometheus-operator => github.com/coreos/prometheus-operator v0.40.0
 	github.com/mtrmac/gpgme => github.com/mtrmac/gpgme v0.1.1
-	github.com/openshift-kni/debug-tools => github.com/openshift-kni/debug-tools v0.1.3-0.20211201134627-ad03be247ef2 // main
```
yanirq, line 175: this pinning was removed on purpose ?

pkg/profilecreator/profilecreator.go
```diff
@@ -344,6 +344,34 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
+	count   int
+}
+
+func newCPUAccumulator() *cpuAccumulator {
+	return &cpuAccumulator{
+		builder: cpuset.NewBuilder(),
+	}
+}
+
+// AddCoresadds logical cores from the slice of *cpu.ProcessorCore to a CPUset till the cpuset size is equal to the max value speicifed
```
yanirq, line 358: s/AddCoresadds/AddCores adds

go.mod
fromanirh: Yes, we can just consume the `v0.1.3` tag.

pkg/profilecreator/profilecreator.go
fromanirh: thanks, fixed

go.mod
fromanirh: yes, we can just consume the `v0.1.3` (or equivalent `v0.1.4`) tag.

pkg/profilecreator/profilecreator.go
```diff
@@ -344,6 +344,34 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
+
+func newCPUAccumulator() *cpuAccumulator {
+	return &cpuAccumulator{
+		builder: cpuset.NewBuilder(),
+	}
+}
+
+// AddCores adds logical cores from the slice of *cpu.ProcessorCore to a CPUset till the cpuset size is equal to the max value specified
+// In case the max is specified as allCores, all the cores from the slice of *cpu.ProcessorCore are added to the CPUSet
+func (ca *cpuAccumulator) AddCores(max int, cores []*cpu.ProcessorCore) {
```
yanirq, line 360: We might want to add a unit test for AddCores. your call

fromanirh: done

PR #814, opened 2021-12-19
tuned base profile to support new scheduler nobs

yanirq:
tuned was updated to handle new kernels (5.13 and newer) 
see https://github.com/redhat-performance/tuned/pull/365
the new kernels moved some sched_ and numa_ knobs from
the sysctl to the debugfs, thus add and abstract these knobs under the
scheduler plugin. With help of this abstraction it will write
the tuning to the correct place according to the kernel used.

Example:
[scheduler]
sched_migration_cost_ns = 500000

Will work on the old kernel the same way as:
[sysctl]
kernel.sched_migration_cost_ns = 500000

all occurences of sched_ and numa_ knobs were added
under the [scheduler] section in the base tuned profile in PAO

### DIFF
--- a/build/assets/tuned/openshift-node-performance
+++ b/build/assets/tuned/openshift-node-performance
@@ -44,6 +44,10 @@ banned_cpus=""
 runtime=0
 group.ksoftirqd=0:f:11:*:ksoftirqd.*
 group.rcuc=0:f:11:*:rcuc.*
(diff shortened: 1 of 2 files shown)


yanirq:
/hold -
still need to verify what happens with having both values and the infliction on RHCOS. 
functional tests might need adjustments.

PR #828, opened 2022-01-19
hack: show cluster version before to run functests

fromanirh:
For debug purposes, to doublcheck the reported cluster version

### DIFF
--- a/hack/show-cluster-version.sh
+++ b/hack/show-cluster-version.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# expect oc to be in PATH by default
+OC_TOOL="${OC_TOOL:-oc}"
+
+echo "Cluster version"
+${OC_TOOL} version || :
+${OC_TOOL} get nodes -o custom-columns=VERSION:.status.nodeInfo.kubeletVersion || :
+${OC_TOOL} get clusterversion || :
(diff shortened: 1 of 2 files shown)


hack/show-cluster-version.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# expect oc to be in PATH by default
+OC_TOOL="${OC_TOOL:-oc}"
+
+echo "Cluster version"
+${OC_TOOL} version || :
+${OC_TOOL} get nodes -o custom-columns=VERSION:.status.nodeInfo.kubeletVersion || :
```
yanirq, line 8: Do we want to inspect all nodes cluster version or one should be enough ?

fromanirh: In theory, just one node is fine, but checking a random node isn't much cheaper than this approach I reckon

yanirq: Not thinking about the cheaper approach here, just the printout clutter in case many nodes are present. 
This is just a suggestion and the current approach should be fine as well. 
Another suggestion - we can add the node names in the printout if we want to present all.

fromanirh: thought about it, but adding the node names could disclose unnecessary detail about the CI infrastructure.

PR #831, opened 2022-01-26
Bug 2046463: Switch to node-device nested maps for netqueues tests

yanirq:
Netqueues tests where previously done by taking a random device from all nodes then applying cross checks against nodes that do not necessarily contain the device.
An example for that could be a cluster where there are both virtual and bare metal worker nodes. 

The fix maps devices to nodes so the tests could check and match the right settings.

### DIFF
--- a/functests/1_performance/netqueues.go
+++ b/functests/1_performance/netqueues.go
@@ -83,15 +83,15 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", func() {
 
 	Context("Updating performance profile for netqueues", func() {
 		It("[test_id:40308][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Network device queues Should be set to the profile's reserved CPUs count ", func() {
-			devices := make(map[string][]int)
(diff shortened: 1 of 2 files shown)


functests/1_performance/netqueues.go
```diff
@@ -83,15 +83,15 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", func() {
 
 	Context("Updating performance profile for netqueues", func() {
 		It("[test_id:40308][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Network device queues Should be set to the profile's reserved CPUs count ", func() {
-			devices := make(map[string][]int)
+			nodesDevices := make(map[string]map[string]int)
```
fromanirh, line 86: please copy/paste the PR description in the commit message so we don't depend on github (but rather on git history) for the rationale of this change

fromanirh (commented):
just a single fix and LGTM

fromanirh:
/lgtm
/hold
to give time to other reviewers to chime in

PR #833, opened 2022-01-27
Embed assets under the go compiled code

cynepco3hahue:
Instead of copying assets under the operator image and loading them during the run, we can embed them during the compilation.
The PR should improve a number of things:
- caching of files, the code will not need to read files from the filesystem for every reconcile loop.
- vendoring performance-addon-operator under the NTO. We do not need to change the NTO image to include PAO assets.

/hold wait for 4.10 branching

### DIFF
--- a/openshift-ci/Dockerfile.deploy
+++ b/openshift-ci/Dockerfile.deploy
@@ -1,11 +1,9 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 
 ARG BIN_DIR=
-ARG ASSETS_DIR=build/assets
 
 ENV LANG=en_US.utf8
 
-COPY ${ASSETS_DIR} /assets
 COPY ${BIN_DIR}performance-addon-operators /usr/local/bin/performance-operator
 COPY ${BIN_DIR}performance-profile-creator /usr/local/bin/performance-profile-creator
 USER 1001
--- a/pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -9,16 +9,18 @@ import (
 	"path/filepath"
 	"text/template"
 
+	"github.com/openshift-kni/performance-addon-operators/build/assets"
+
 	"github.com/coreos/go-systemd/unit"
 	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/pointer"
+
 	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
 	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
 	profilecomponent "github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/profile"
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/utils/pointer"
 )
 
 const (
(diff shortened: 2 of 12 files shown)


pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -34,19 +36,20 @@ const (
...
-	crioRuntimesConfig  = "99-runtimes.conf"
-	ociHooks            = "low-latency-hooks"
+	bashScriptsDir     = "/usr/local/bin"
+	crioConfd          = "/etc/crio/crio.conf.d"
+	crioRuntimesConfig = "99-runtimes.conf"
 	// OCIHooksConfigDir is the default directory for the OCI hooks
 	OCIHooksConfigDir = "/etc/containers/oci/hooks.d"
 	// OCIHooksConfig file contains the low latency hooks configuration
-	OCIHooksConfig     = "99-low-latency-hooks"
+	OCIHooksConfig     = "99-low-latency-hooks.json"
```
yanirq, line 45: is this the only json file that now needs a specification for the extension ?   how will this play with cnf-depoly ?

cynepco3hahue: hm, what do you mean, it is the file that we are injecting under the node to make OCI hooks work.

openshift-ci/Dockerfile.deploy
```diff
@@ -1,11 +1,9 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 
 ARG BIN_DIR=
-ARG ASSETS_DIR=build/assets
```
yanirq, line 4: this is only true for this docker file or should we also align other docker files we maintain ?

cynepco3hahue: Yes, we will need to update relevant d/s dockerfile, but it not critical, because it will continue to work also with assets files under the operator image, it will just not use them

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -137,56 +140,40 @@ func getIgnitionConfig(assetsDir string, profile *performancev2.PerformanceProfi
 	// add script files under the node /usr/local/bin directory
 	mode := 0700
 	for _, script := range []string{hugepagesAllocation, ociHooks, setRPSMask} {
-		src := filepath.Join(assetsDir, "scripts", fmt.Sprintf("%s.sh", script))
-		if err := addFile(ignitionConfig, src, getBashScriptPath(script), &mode); err != nil {
+		dst := getBashScriptPath(script)
+		content, err := assets.Scripts.ReadFile(fmt.Sprintf("scripts/%s.sh", script))
```
yanirq, line 144: ReadFile is done from cache here ?

pkg/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -166,7 +166,7 @@ func NewNodePerformance(assetsDir string, profile *performancev2.PerformanceProf
 		}
 	}
 
-	profileData, err := getProfileData(getProfilePath(components.ProfileNamePerformance, assetsDir), templateArgs)
+	profileData, err := getProfileData(filepath.Join("tuned", components.ProfileNamePerformance), templateArgs)
```
yanirq, line 169: maybe use a constant instead of "tuned" string

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
yanirq: ack. I see its already done for the rest here for the injection (they already had the files extension before)

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
cynepco3hahue: Yep.

pkg/controller/performanceprofile/components/tuned/tuned.go
cynepco3hahue: I think it is a single place, so not really worth it, but if you insist I can do it.

fromanirh (commented):
I like the idea. Makes our tests also a bit easier and more robust. Let's just make sure we don't break d/s and cnf-tests (we should not, but worth doublechecking)

cynepco3hahue:
> I like the idea. Makes our tests also a bit easier and more robust. Let's just make sure we don't break d/s and cnf-tests (we should not, but worth doublechecking)

Yep, it should not break anything, until the d/s update, it will continue to copy files under the image, but will not use them.

cynepco3hahue:
We will need future d/s alignment once this PR is merged, so I will unhold it and process after with the d/s.

PR #836, opened 2022-02-03
Latency e2e: fix skipping report when cpus amount is less than minimum

shajmakh:
Missing fmt.Sprintf in the skip message leads to outputting the conversion character %d instead of minCpuAmountForOslat actual value.

Fix the message formatting by outputting fmt.Sprintf value and validate that in Latency_testing.go.

### DIFF
--- a/functests/5_latency_testing/latency_testing.go
+++ b/functests/5_latency_testing/latency_testing.go
@@ -35,6 +35,7 @@ const (
 	incorrectMsgPart2                  = " has incorrect value"
 	invalidNumber                      = " has an invalid number"
 	maxInt                             = "2147483647"
+	minimumCpuForOslat                 = "2"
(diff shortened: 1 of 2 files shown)


functests/5_latency_testing/latency_testing.go
```diff
@@ -35,6 +35,7 @@ const (
 	incorrectMsgPart2                  = " has incorrect value"
 	invalidNumber                      = " has an invalid number"
 	maxInt                             = "2147483647"
+	minimumCpuForOslat                 = "2"
```
Tal-or, line 38: If you have similar variables in multiple packages, you can move them into a common package and import them from there.
But this can be done on a separate PR.

shajmakh: good idea, I'll consider that next. Thanks!

PR #838, opened 2022-02-07
e2e upgrade: wait for other components to handle PAO resources updates

cynepco3hahue:
It possible that the PAO will need to update KubeletConfig, MachineConfig or
Tuned profiles after the upgrade, in this case we should wait until relevant
controllers will finish work to update nodes with relevant changes.


functests/utils/tuned/tuned.go
```diff
@@ -0,0 +1,55 @@
...
+				Name:      tunedProfileName,
+				Namespace: components.NamespaceNodeTuningOperator,
+			}
+
+			if err := testclient.Client.Get(context.TODO(), key, profile); err != nil {
+				klog.Errorf("failed to get tuned profile %q: %v", tunedProfileName, err)
+				return false, nil
+			}
+
+			appliedCondition, err := GetConditionByType(profile.Status.Conditions, tunedv1.TunedProfileApplied)
```
yanirq, line 32: Is there a scenario here where we will get the applied = true for the old state ? (before upgrade)

functests-extended/1_performance_operator_upgrade/upgrade_operator.go
```diff
@@ -95,6 +117,18 @@ var _ = Describe("[rfe_id:28567][performance] Performance Addon Operator Upgrade
...
+		// during the PAO upgrade, so the best we can do here is to wait some time,
+		// to give other controllers time to notify changes
+		time.Sleep(2 * time.Minute)
+		mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+
+		var workerRTNodesNames []string
+		for _, workerRTNode := range workerRTNodes {
+			workerRTNodesNames = append(workerRTNodesNames, workerRTNode.Name)
+		}
+		Expect(tuned.WaitForAppliedCondition(workerRTNodesNames, corev1.ConditionTrue, 5*time.Minute))
```
yanirq, line 131: suggestion (ignore if this is handled): 
If we want to check that tuned daemonsets where reloaded before profile applied check 
we can keep their pids for example `/run/tuned/tuned.pid`  then check if it was changed (or check the pod generation)

functests/utils/tuned/tuned.go
cynepco3hahue: sure, it the reason why I am waiting two minutes before

functests-extended/1_performance_operator_upgrade/upgrade_operator.go
yanirq: well its either that or the 2 minutes wait you added above, but because we also need to wait for kubeconfig and MCs to be updated then maybe we can keep the more general (not "precise") approach

yanirq:
/lgtm
lets release the pipeline asap

PR #844, opened 2022-02-09
Latency testing:  enable LATENCY_TEST_DELAY tests; delete namespace if image pre-pulling failed 

shajmakh:
Now that BZ 200667 is resolved, enable the tests that set LATENCY_TEST_DELAY to a non-zero value for cyclictest and hwlatdetect.

Fatalf calls os.Exit(1) thus AfterSuite won't be reached to delete the namespace.
Delete the namespace before logging fatal to guarantee cleaning the cluster from the suite's configuration.

### DIFF
--- a/functests/5_latency_testing/5_latency_testing_suite_test.go
+++ b/functests/5_latency_testing/5_latency_testing_suite_test.go
@@ -28,11 +28,7 @@ var prePullNamespace = &corev1.Namespace{
 }
 
 var _ = AfterSuite(func() {
-	prePullNamespaceName := prePullNamespace.Name
-	err := testclient.Client.Delete(context.TODO(), prePullNamespace)
-	testlog.Infof("deleted namespace %q err=%v", prePullNamespace.Name, err)
-	Expect(err).ToNot(HaveOccurred())
-	err = namespaces.WaitForDeletion(prePullNamespaceName, 5*time.Minute)
+	deleteNamespace()
 })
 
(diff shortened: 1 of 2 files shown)


functests/5_latency_testing/5_latency_testing_suite_test.go
```diff
@@ -50,6 +46,7 @@ func Test5LatencyTesting(t *testing.T) {
 	if err != nil {
 		data, _ := json.Marshal(ds) // we can safely skip errors
 		testlog.Infof("DaemonSet %s/%s image=%q status:\n%s", ds.Namespace, ds.Name, images.Test(), string(data))
+		deleteNamespace()
```
Tal-or, line 49: In general the ```Test5LatencyTesting``` function should be as minimal as possible and just call the entry point for ginkgo test:
```
	RegisterFailHandler(Fail)
	rr := []Reporter{}
	if ginkgo_reporters.Polarion.Run {
		rr = append(rr, &ginkgo_reporters.Polarion)
	}
	rr = append(rr, junit.NewJUnitReporter("latency_testing"))
	RunSpecsWithDefaultAndCustomReporters(t, "Performance Addon Operator latency tools testing", rr)
```

Then you can have all the extra logic, pre-pulling the image, declaring the client, etc. on a ```BeforeSuite()```
Then even if the test will fail, the ```AfterSuite()``` will get called anyway

PR #848, opened 2022-02-21
Allow more update time for netqueues tests

yanirq:
netqueues tests were adjusted with polling timers in order to allow
enough time for the tuned daemon and netqueues on the nodes
to be adjusted.

### DIFF
--- a/functests/1_performance/netqueues.go
+++ b/functests/1_performance/netqueues.go
@@ -10,6 +10,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


mrniranjan:
LGTM

functests/1_performance/netqueues.go
```diff
@@ -67,6 +71,14 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", func() {
 			By("Updating the performance profile")
 			profiles.UpdateWithRetry(profile)
 		}
+		// Verify the tuned profile is updated on the worker-cnf nodes
```
fromanirh, line 74: I'd actually use a `Eventually` block. `Sleep`s are always suspicious (and yes, I'm guilty of using them sometimes. So let's stop :) )

functests/1_performance/netqueues.go
```diff
@@ -153,6 +158,15 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", func() {
 				}
 				By("Updating the performance profile")
 				profiles.UpdateWithRetry(profile)
+
+				// Verify the tuned profile is updated on the worker-cnf nodes
+				time.Sleep(20 * time.Second)
```
fromanirh, line 163: same

functests/1_performance/netqueues.go
```diff
@@ -251,6 +280,14 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", func() {
 					},
 				}
 				profiles.UpdateWithRetry(profile)
+				// Verify the tuned profile is updated on the worker-cnf nodes
+				time.Sleep(20 * time.Second)
```
fromanirh, line 284: same

functests/1_performance/netqueues.go
yanirq: this is actually a bit tricky here, I would even remove the check of the tuned profile now that I think of it since it can remain from test to test and the sleep is what does the job.
since the tuned profile will have the same name in every test there will have to be another way to check if the update was applied (will need to think about this further)

PR #850, opened 2022-02-22
Enable timer migration

MarSik:
This is needed for low latency on OCP as a timer might have been
scheduled on a core just before it was given to a guaranteed
workload. Later that timer will expire and interrupt the workload.
This option lets the kernel move the timer to another cpu first.

### DIFF
--- a/build/assets/tuned/openshift-node-performance
+++ b/build/assets/tuned/openshift-node-performance
@@ -59,9 +59,9 @@ kernel.hung_task_timeout_secs = 600
 kernel.nmi_watchdog = 0
 #> realtime
 kernel.sched_rt_runtime_us = -1
-#> cpu-partitioning and realtime for RHEL disable it (= 0)
+# cpu-partitioning and realtime for RHEL disable it (= 0)
 # OCP is too dynamic when partitioning and needs to evacuate
-# scheduled timers when starting a guaranteed workload (= 1)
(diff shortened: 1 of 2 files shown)


build/assets/tuned/openshift-node-performance
```diff
@@ -59,8 +59,10 @@ kernel.hung_task_timeout_secs = 600
 kernel.nmi_watchdog = 0
 #> realtime
 kernel.sched_rt_runtime_us = -1
-#> cpu-partitioning (= 1) #realtime (= 0)
-kernel.timer_migration = 0
+#> cpu-partitioning and realtime for RHEL disable it (= 0)
+# OCP is too dynamic when partitioning and needs to evacuate
+# scheduled timers when starting a guaranteed workload (= 1)
```
yanirq, line 64: I believe that if we want to keep the comments inline we should use the same format as in other place we have here.
meaning - move `>` to the end of the comment section here (line 62 to line 64)

yanirq: make sure to adjust the [testdata/render-expected-output/manual_tuned.yaml](https://github.com/openshift-kni/performance-addon-operators/pull/850/files#diff-853529c02df5410a1582f7d76352236ee3b9de3550af446eba701685027e4156)  if you edit the file as I mentioned above

MarSik: Fixed.

PR #851, opened 2022-02-23
e2e (netqueues): do not skip from called functions

yanirq:
Skip from called functions will add an unwanted panic stacktrace that will litter the test logs.
Use Skip only from main test functions.


functests/1_performance/netqueues.go
```diff
@@ -309,9 +321,9 @@ func checkDeviceSupport(workernodes []corev1.Node, nodesDevices map[string]map[s
 		}
 	}
 	if noSupportedDevices {
-		Skip("Skipping Test: There are no supported Network Devices")
+		return false, err
 	}
-	return err
+	return true, err
```
Tal-or, line 326: Why not instead of:
```
if noSupportedDevices {
		return false, err
	}
	return true, err
```
Just:
```return !noSupportedDevices, err``` ?

yanirq: I could use that, but this is clearer, the tests are already a bit too large in code

PR #857, opened 2022-02-28
WIP: Remove OLM operator on startup

yanirq:
Once PAO is merged and shipped with NTO (as one controller) there is a need to clean up the old OLM operator, especially in upgrade use cases. 
This PR should be aligned with https://github.com/openshift/cluster-node-tuning-operator/pull/314 .

Functional tests and the existing deployment schemes in this repo will align either in this or in an additional PR/s


controllers/performanceprofile_controller.go
```diff
@@ -135,6 +142,91 @@ func (r *PerformanceProfileReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+	}
+
+	if err := r.Get(context.TODO(), key, subscription); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+	} else {
+		klog.Infof("Removing performance-addon-operator subscription %s", subscription.Name)
+		if subscription.Status.CurrentCSV != paoCSV {
+			klog.Warningf("Subscription to be removed contains a current CSV version %s which is different from %s", subscription.Status.CurrentCSV, paoCSV)
```
cynepco3hahue, line 162: probably we should exit here

controllers/performanceprofile_controller.go
```diff
@@ -135,6 +142,91 @@ func (r *PerformanceProfileReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Complete(r)
 }
 
+// uninstall PAO OLM operator and all of its artifacts
+// this should apply only from version 4.11
+func (r *PerformanceProfileReconciler) removeOLMOperator() error {
```
cynepco3hahue, line 147: you will need a bunch of new permissions for OLM related resources

controllers/performanceprofile_controller.go
yanirq: I wonder what can bring us to this scenario at all. maybe this is an overkill

cynepco3hahue: only if a user will not update a subscription from 4.9 or earlier version(but it should be the unsupported scenario, so I think we should exit with the error that first, you should upgrade to 4.10) and will upgrade to 4.11.

yanirq:
This is implemented under the new repository for PAO:
https://github.com/openshift/cluster-node-tuning-operator/pull/336

yanirq:
This is implemented under the new repository for PAO:
https://github.com/openshift/cluster-node-tuning-operator/pull/336

PR #858, opened 2022-03-07
Implement a new workload hints functionality

cynepco3hahue:
The new API should make it easier to configure nodes for specific types of workloads via the performance profile.

### How does it look?
```
...
spec:
  workloadHints:
    realTime: true
```

```
...
spec:
  workloadHints:
    powerSaving: true
...
```

### How does it work?

The tuned profile has:
1. Have a separate set of kernel arguments for each workload hint.
2. Set `sysctl`, `cpu` and `scheduler` section parameters according to
workload hints.


api/v1/performanceprofile_types.go
```diff
@@ -70,6 +70,9 @@ type PerformanceProfileSpec struct {
 	// Defaults to "false"
 	// +optional
 	GloballyDisableIrqLoadBalancing *bool `json:"globallyDisableIrqLoadBalancing,omitempty"`
+	// WorkloadHints defines hints for different types of workloads. It will allow defining exact set of tuned and
+	// kernel arguments that should be applied on top of the node.
+	WorkloadHints *WorkloadHints `json:"workloadHints,omitempty"`
```
fromanirh, line 75: I wonder if this warrants a API bump. Probably not, but worth checking. Also, we need to check we're doing OK with the API conversion rules (v2 <-> v1).
The API per se looks nice.

pkg/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -166,6 +168,21 @@ func NewNodePerformance(profile *performancev2.PerformanceProfile) (*tunedv1.Tun
 		}
 	}
 
+	// TODO: I think we need to parse additional kernel arguments and in case when it intersect with workload hints
+	// kernel arguments we should or override it or drop an error
+	// set default values for workload hints
```
fromanirh, line 173: I tend to agree to the comment.

fromanirh (commented):
VERY initial review. Everything so far looks reasonnable, but let's make sure we do our due diligence wrt the OCP API guarantees and requirements

api/v1/performanceprofile_types.go
cynepco3hahue: yep, we should have v1 and v2 API consistent because of OpenShift rules.

pkg/controller/performanceprofile/components/tuned/tuned.go
cynepco3hahue: I will provide it after as an improvement.

fromanirh: fine for me

fromanirh (commented):
did another pass, and pending the future extension and the API concerns we already talked about, LGTM.
Nice work!

build/assets/tuned/openshift-node-performance
```diff
@@ -121,12 +129,24 @@ avc_cache_threshold=8192
...
 {{if .StaticIsolation}}
-cmdline_realtime=+tsc=nowatchdog intel_iommu=on iommu=pt isolcpus=domain,managed_irq,${isolated_cores} systemd.cpu_affinity=${not_isolated_cores_expanded}
+cmdline_isolation=+isolcpus=domain,managed_irq,${isolated_cores}
 {{else}}
-cmdline_realtime=+tsc=nowatchdog intel_iommu=on iommu=pt isolcpus=managed_irq,${isolated_cores} systemd.cpu_affinity=${not_isolated_cores_expanded}
+cmdline_isolation=+isolcpus=managed_irq,${isolated_cores}
+{{end}}
+
+{{if .RealTime}}
+cmdline_realtime=+tsc=nowatchdog nosoftlockup nmi_watchdog=0 audit=0 mce=off skew_tick=1
```
browsell, line 143: nmi_watchdog is already disabled at line 66 above.
Disabling audit is likely not acceptable.
Why would we disable mce ?

build/assets/tuned/openshift-node-performance
```diff
@@ -121,12 +129,24 @@ avc_cache_threshold=8192
...
 {{end}}
+
+{{if not .PowerSaving}}
+cmdline_power_performance=+processor.max_cstate=1 intel_idle.max_cstate=0 idle=poll intel_pstate=disable
```
browsell, line 147: Since you idle=poll is being set why would processor.max_cstate=1 intel_idle.max_cstate=0 need to be set ?

build/assets/tuned/openshift-node-performance
```diff
@@ -121,12 +129,24 @@ avc_cache_threshold=8192
 initrd_remove_dir=
 initrd_dst_img=
 initrd_add_dir=
+
 # overrides cpu-partitioning cmdline
-cmdline_cpu_part=+nohz=on rcu_nocbs=${isolated_cores} tuned.non_isolcpus=${not_isolated_cpumask} intel_pstate=disable nosoftlockup
+cmdline_cpu_part=+nohz=on rcu_nocbs=${isolated_cores} tuned.non_isolcpus=${not_isolated_cpumask} systemd.cpu_affinity=${not_isolated_cores_expanded} intel_iommu=on iommu=pt
```
browsell, line 134: Why do we not include nohz_full ?

cynepco3hahue: @MarSik IIRC it was some downsides regarding using `nohz_full`, can you please confirm?

fromanirh:
The (important) comments about changing profiles should be addressed here or on a separate PR? Because this PR, AFAIK is about how existing data is combined to form the final output.

cynepco3hahue:
> The (important) comments about changing profiles should be addressed here or on a separate PR? Because this PR, AFAIK is about how existing data is combined to form the final output.

I will do it under the current PR, because anyway, we should move it under the NTO.

build/assets/tuned/openshift-node-performance
cynepco3hahue: Added under the real time workload hint - https://github.com/openshift-kni/performance-addon-operators/pull/858/files#diff-73267c64c70b410234c47702b68c892fbbe79bef657f0658068e3de80e07b94dR143

cynepco3hahue:
@browsell Can you please answer https://github.com/openshift-kni/performance-addon-operators/pull/858#discussion_r841722345 and https://github.com/openshift-kni/performance-addon-operators/pull/858#discussion_r841720292? I want to finalize it and progress with it.

cynepco3hahue:
close in favor of https://github.com/openshift/cluster-node-tuning-operator/pull/339

PR #830, opened 2022-01-26
enable cpumanager policy options - full-pcpus-only

fromanirh:
When we require the strictest NUMA alignment, we also enable the cpumanager policy option to require full physical cores only,
a feature available in kube >= 1.23.

Reprises https://github.com/openshift-kni/performance-addon-operators/pull/790

PR #864, opened 2022-03-15
Latency testing: stabilize tests 

shajmakh:
Due to modification done in #830, tests requesting an odd number of CPU are failing due to rejecting the pod from running with SMTAlignmentError.
Adjust the tests to use an even number of CPUs.

And now that BZ 200667 is resolved, enable the tests that set LATENCY_TEST_DELAY to a non-zero value for cyclictest and hwlatdetect.

### DIFF
--- a/functests/5_latency_testing/latency_testing.go
+++ b/functests/5_latency_testing/latency_testing.go
@@ -208,11 +208,9 @@ func getValidValuesTests(toolToTest string) []latencyTest {
 	var testSet []latencyTest
 	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "5", testMaxLatency: guaranteedLatency, testCpus: "2", outputMsgs: []string{success}, toolToTest: toolToTest})
 	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "1", testMaxLatency: guaranteedLatency, testCpus: "5", outputMsgs: []string{success}, toolToTest: toolToTest})
-	//BZ https://bugzilla.redhat.com/show_bug.cgi?id=2006675
-	if toolToTest == oslat {
-		testSet = append(testSet, latencyTest{testDelay: "1", testRun: "true", testRuntime: "2", testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
-		testSet = append(testSet, latencyTest{testDelay: "60", testRun: "true", testRuntime: "2", testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
-	}
+	testSet = append(testSet, latencyTest{testDelay: "1", testRun: "true", testRuntime: "2", testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
+	testSet = append(testSet, latencyTest{testDelay: "60", testRun: "true", testRuntime: "2", testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
+
 	if toolToTest != hwlatdetect {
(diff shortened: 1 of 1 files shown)


functests/5_latency_testing/latency_testing.go
```diff
@@ -206,28 +206,28 @@ func clearEnv() {
 
 func getValidValuesTests(toolToTest string) []latencyTest {
 	var testSet []latencyTest
-	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "5", testMaxLatency: guaranteedLatency, testCpus: "2", outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "1", testMaxLatency: guaranteedLatency, testCpus: "5", outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "1", testRun: "true", testRuntime: "2", testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
+	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, testCpus: "2", outputMsgs: []string{success}, toolToTest: toolToTest})
```
fromanirh, line 209: could you add a rationale for these values? notes in the commit message are fine

shajmakh: sure, I'll add them, thank you for your comment.

functests/5_latency_testing/latency_testing.go
```diff
@@ -206,28 +206,32 @@ func clearEnv() {
...
-	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "1", testMaxLatency: guaranteedLatency, testCpus: "5", outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "1", testRun: "true", testRuntime: "2", testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
+
+	//testRuntime: let runtime be 10 seconds for most of the tests and not less, that is to let the tools
+	//have their time to measure latency properly hence stabilizing the tests
+	//testCpus: for tests that expect a success output message, note that an even CPU number is needed, otherwise the test would fail with SMTAlignmentError
+	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, testCpus: "2", outputMsgs: []string{success}, toolToTest: toolToTest})
+	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, testCpus: "6", outputMsgs: []string{success}, toolToTest: toolToTest})
```
fromanirh, line 214: I get why testCpus was moved from 5 to 6, but why testRuntime was changed from 1 to 10?

shajmakh: lsee line 210-211:
```
testRuntime: let runtime be 10 seconds for most of the tests and not less, that is to let the tools
have their time to measure latency properly hence stabilizing the tests
```
I have noticed that 5 is not good enough when we expect success message, that is, when expected latency is set to 1 and hwlatdetect runs for 5 seconds there are actually no samples to compare the latency with, thus adjusted the time to most of the tests to be 10 seconds

fromanirh: ok, makes sense. Next time please considering splitting the commits such as
- one commit changes the cpu count to avoid SMTAlignmentError
- one commit change the test time to improve the stability of the tests

a commit should be a single unit of change and in general we should avoid mixing changes together unless they are really tightly coupled (this is not the case)

fromanirh (commented):
/lgtm
/approve

functests/5_latency_testing/latency_testing.go
shajmakh: you're right. I will take that into consideration, thank you for elaborating!

PR #872, opened 2022-03-25
WIP: Add capability to offline cpus via PAO

marioferh:
WIP:
- Add PerformanceProfile API ✓ 
- Test PAO API
- Measure per-core offline savings
- Must-gather collects data
- Fill Doc


pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -221,6 +228,24 @@ func getIgnitionConfig(profile *performancev2.PerformanceProfile) (*igntypes.Con
 		})
 	}
 
+	if profile.Spec.CPU.Offlined != nil {
+		offlinedCPUSList, err := cpuset.Parse(string(*profile.Spec.CPU.Offlined))
+		if err != nil {
+			return nil, err
+		}
+		offlinedCPUSstring := offlinedCPUSList.StringSplit()
```
fromanirh, line 236: from where this comes from?

fromanirh: `String()` here is not ideal because the implementation can coalesce ranges (e.g. 0,1,2,5 can become 0-2,5) which will make the life of the script harder. Better to use something like `ToSlice()` and create an explicit string with the separator which most helps the script, like `0,1,5` or `0 1 5`

yanirq:
This PR will now have to be implemented under https://github.com/openshift/cluster-node-tuning-operator  since PAO code base has moved to NTO

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
marioferh: [cpuset.go](https://github.com/openshift-kni/performance-addon-operators/pull/872/files#diff-d98fdf4c4970d1faa33fb971ca756472db77e98f1536480b8ab938ccd769777a)
Its there, sorry but I didn't realize that was in a vendor file. with ToSlice() I get 0-5 instead of 0,1,2,3,4,5. But yep its betteer to use ToSlice and create an explicit string to help the script. 
Thanks

marioferh: now better?

pkg/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -352,3 +399,7 @@ func renderCrioConfigSnippet(profile *performancev2.PerformanceProfile, src stri
 
 	return crioConfig.Bytes(), nil
 }
+
+func stringSplit(cpuListSlice []int) string {
+	return strings.Trim(strings.Join(strings.Fields(fmt.Sprint(cpuListSlice)), ","), "[]")
+}
```
fromanirh, line 405:
```
func listToString(cpus []int) string {
    items := make([]string, len(cpus))
    for idx, cpu := range cpus {
        items[idx] = strconv.FormatInt(cpu, 10)
    }
    return strings.Join(cpus, ",")
}
```
(this will require a unit test)

PR #875, opened 2022-04-12
Add Hyper-threading aware scheduling test cases.

dgonyier:
Test case IDs to execute based on current cluster toplogy:
* SNO vs non-SNO
* Hyper-Threading enabled vs disbled
* SNO with Workload Partitioning enabled vs disabled


yanirq:
@dgonyier  please note that this should go under https://github.com/openshift/cluster-node-tuning-operator  which is the new repo for PAO and related PAO tests

PR #876, opened 2022-04-14
Add Hyper-threading aware scheduling test cases.

dgonyier:
Test case IDs to execute based on current cluster toplogy:

* SNO vs non-SNO
* Hyper-Threading enabled vs disbled
* SNO with Workload Partitioning enabled vs disabled


yanirq:
@dgonyier  lets close this PR since it moved to NTO

PR #887, opened 2022-05-18
must-gather gather-sysinfo unit tests

marioferh:
Add unit test of gather-sysinfo

gather-sysinfo is must-gather main tool

### DIFF
--- a/tools/gather-sysinfo/gather-sysinfo_suite_test.go
+++ b/tools/gather-sysinfo/gather-sysinfo_suite_test.go
@@ -1,0 +1,13 @@
+package main
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestComponetsUtils(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Tuned Suite")
+}
--- a/tools/gather-sysinfo/gather-sysinfo_test.go
+++ b/tools/gather-sysinfo/gather-sysinfo_test.go
@@ -1,0 +1,93 @@
+package main
+
+import (
+	"io/ioutil"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openshift-kni/debug-tools/pkg/knit/cmd"
+	"github.com/spf13/cobra"
+)
+
+var kniEntries = []string{
+	"/host/proc/cmdline",
+	"/host/proc/interrupts",
+	"/host/proc/irq/default_smp_affinity",
+	"/host/proc/irq/*/*affinity_list",
+	"/host/proc/irq/*/node",
+	"/host/proc/softirqs",
+	"/host/sys/devices/system/cpu/smt/active",
+	"/host/proc/sys/kernel/sched_domain/cpu*/domain*/flags",
+	"/host/sys/devices/system/cpu/offline",
+	"/host/sys/class/dmi/id/bios*",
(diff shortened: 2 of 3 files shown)


tools/gather-sysinfo/gather-sysinfo_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"/host/proc/sys/kernel/sched_domain/cpu*/domain*/flags",
+	"/host/sys/devices/system/cpu/offline",
+	"/host/sys/class/dmi/id/bios*",
+	"/host/sys/class/dmi/id/product_family",
+	"/host/sys/class/dmi/id/product_name",
+	"/host/sys/class/dmi/id/product_sku",
+	"/host/sys/class/dmi/id/product_version",
+}
+
+var dupEntries = []string{"/host/proc/cmdline"}
```
yanirq, line 29: can we make the var names more clear ?

tools/gather-sysinfo/gather-sysinfo_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package main
```
fromanirh, line 1: I'm not a fan of using ginkgo for unit tests. We add a great amount of baggage with unclear benefits. What are the benefits of using ginkgo here?

tools/gather-sysinfo/gather-sysinfo_test.go
```diff
@@ -0,0 +1,91 @@
+package main
```
fromanirh, line 1: we may want to strip the logic from the main source and move them into an **internal** package to make unit test easier

tools/gather-sysinfo/gather-sysinfo_test.go
marioferh: right

tools/gather-sysinfo/gather-sysinfo_suite_test.go
marioferh: Same opinion here, but we decided to use ginkgo for everything.

tools/gather-sysinfo/gather-sysinfo_test.go
fromanirh: `tools/gather-sysinfo/$WHATEVER` can be a nice start

marioferh: ok, thanks

marioferh:
@yanirq please override gpc and operator upgrade. Thanks

fromanirh (commented):
these are not unit tests for must gather, but unit tests for the main tool which we use in must-gather, `gather-sysinfo`. So this should be reflected in the commit message (which should also be less terse).

I still believe we should split the logic from the `gather-sysinfo` main package to make the tests easier and nicer, but I won't push too much for that.

So, conditional LGTM, need to have the commit message fixed and then we can move forward.

marioferh:
> these are not unit tests for must gather, but unit tests for the main tool which we use in must-gather, `gather-sysinfo`. So this should be reflected in the commit message (which should also be less terse).
> 
> I still believe we should split the logic from the `gather-sysinfo` main package to make the tests easier and nicer, but I won't push too much for that.
> 
> So, conditional LGTM, need to have the commit message fixed and then we can move forward.

done

marioferh:
thanks @fromanirh next step: https://github.com/openshift-kni/performance-addon-operators/issues/897

marioferh:
@fromanirh @yanirq needs override

PR #891, opened 2022-05-30
E2e latency fixes

shajmakh:
- add the runtime to stabilize the pod-phase condition
- verify tool args 
- change ImagePullPolicy to default
- fail the test if CPUs number is less than 1

### DIFF
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -313,7 +313,7 @@ func getLatencyTestCpus() (int, error) {
 		if err != nil {
 			return val, fmt.Errorf("the environment variable LATENCY_TEST_CPUS has incorrect value %q, it must be a positive integer with maximum value of %d: %w", latencyTestCpusEnv, math.MaxInt32, err)
 		}
-		if val < 0 || val > math.MaxInt32 {
+		if val <= 0 || val > math.MaxInt32 {
 			return val, fmt.Errorf("the environment variable LATENCY_TEST_CPUS has an invalid number %q, it must be a positive integer with maximum value of %d", latencyTestCpusEnv, math.MaxInt32)
 		}
 		return val, nil
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -454,12 +454,18 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
 	Expect(err).ToNot(HaveOccurred())
 
 	By("Waiting two minutes to download the latencyTest image")
-	err = pods.WaitForPhase(testPod, corev1.PodRunning, 2*time.Minute)
+	currentPod, err := pods.WaitForPredicate(testPod, 2*time.Minute, func(pod *corev1.Pod) (bool, error) {
+		if pod.Status.Phase == corev1.PodRunning {
+			return true, nil
+		}
+		return false, nil
+	})
+
 	if err != nil {
 		testlog.Error(err)
 		logEventsForPod(testPod)
 	}
-	Expect(err).ToNot(HaveOccurred())
+	Expect(err).ToNot(HaveOccurred(), "expected the pod to reach running phase, its current phase is %s", currentPod.Status.Phase)
 
 	if runtime, _ := strconv.Atoi(latencyTestRuntime); runtime > 1 {
 		By("Checking actual CPUs number for the running pod")
@@ -476,8 +482,9 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
 	if err != nil {
 		testlog.Error(err)
 		logEventsForPod(testPod)
+		testlog.Info(getLogFile(node, logName))
 	}
-	Expect(err).ToNot(HaveOccurred(), getLogFile(node, logName))
+	Expect(err).ToNot(HaveOccurred())
 }
 
 func extractLatencyValues(logName string, exp string, node *corev1.Node) string {
--- a/functests/5_latency_testing/latency_testing.go
+++ b/functests/5_latency_testing/latency_testing.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"strings"
 
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
@@ -122,8 +123,27 @@ var _ = table.DescribeTable("Test latency measurement tools tests", func(testGro
 				testlog.Error(matchErr.Error())
 			}
 			if ok {
-				commandRegex := fmt.Sprintf("%s command with arguments .*--duration %s", test.toolToTest, test.testRuntime)
-				Expect(string(output)).To(MatchRegexp(commandRegex), "The output of the executed tool is not as expected")
+				//verify the command is executed with the expected args
+				//this lists of args depend on the ones the latency tool runners adds to tool command in cnf-features-deploy.
+				var passedArgs []string
+				switch test.toolToTest {
+				case oslat:
+					passedArgs = []string{"--duration " + test.testRuntime, "--rtprio ", "--cpu-list ", "--cpu-main-thread "}
+				case cyclictest:
+					passedArgs = []string{"--duration " + test.testRuntime, "--priority 95", "--threads ", "--affinity ", "--histogram ", "--interval ", "--mlockall ", "--quiet"}
+				case hwlatdetect:
+					thr := test.testMaxLatency
+					if test.hwlatdetectMaxLatency != "" {
+						thr = test.hwlatdetectMaxLatency
+					}
+					passedArgs = []string{"--duration " + test.testRuntime, "--threshold " + thr, "--hardlimit " + thr, "--window ", "--width "}
+				default:
+					testlog.Error("the tool to test was not set")
+				}
+
+				for _, argument := range passedArgs {
+					Expect(strings.Contains(string(output), argument)).To(BeTrue(), "The tool command didn't pass the argument %q", argument)
+				}
 			}
 		}
 		for _, msg := range test.outputMsgs {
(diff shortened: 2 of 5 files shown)


shajmakh:
complement PR to #889

functests/5_latency_testing/latency_testing.go
```diff
@@ -214,7 +214,7 @@ func getValidValuesTests(toolToTest string) []latencyTest {
 	if toolToTest == oslat {
 		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: "10", testMaxLatency: "1", oslatMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
 		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: "10", oslatMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
-		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, testCpus: "1", outputMsgs: []string{skip, skipOslatCpuNumber}, toolToTest: toolToTest})
+		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, testCpus: "0", outputMsgs: []string{skip, skipOslatCpuNumber}, toolToTest: toolToTest})
```
fromanirh, line 220: however requesting 0 cpus is misleading for another reasons. Maybe ask 2?

functests/5_latency_testing/latency_testing.go
```diff
@@ -267,3 +270,22 @@ func getNegativeTests(toolToTest string) []latencyTest {
...
+		return []string{"--duration " + tst.testRuntime, "--rtprio ", "--cpu-list ", "--cpu-main-thread "}
+	case cyclictest:
+		return []string{"--duration " + tst.testRuntime, "--priority 95", "--threads ", "--affinity ", "--histogram ", "--interval ", "--mlockall ", "--quiet"}
+	case hwlatdetect:
+		thr := tst.testMaxLatency
+		if tst.hwlatdetectMaxLatency != "" {
+			thr = tst.hwlatdetectMaxLatency
+		}
+		return []string{"--duration " + tst.testRuntime, "--threshold " + thr, "--hardlimit " + thr, "--window ", "--width "}
+	default:
```
fromanirh, line 287: you want to fail very loudly here

functests/5_latency_testing/latency_testing.go
```diff
@@ -267,3 +270,22 @@ func getNegativeTests(toolToTest string) []latencyTest {
 	}
 	return testSet
 }
+
+func getToolArgsRegex(tst latencyTest) []string {
+	//this lists of args depend on the ones the latency tool runners adds to tool command in cnf-features-deploy.
+	switch tst.toolToTest {
+	case oslat:
```
fromanirh, line 277: Shouldn't this be in quotes ("oslat")? same for all the cases below

functests/5_latency_testing/latency_testing.go
```diff
@@ -122,8 +122,11 @@ var _ = table.DescribeTable("Test latency measurement tools tests", func(testGro
 				testlog.Error(matchErr.Error())
 			}
 			if ok {
-				commandRegex := fmt.Sprintf("%s command with arguments .*--duration %s", test.toolToTest, test.testRuntime)
-				Expect(string(output)).To(MatchRegexp(commandRegex), "The output of the executed tool is not as expected")
+				//verify the command is executed with the expected args
+				args := getToolArgsRegex(test)
```
fromanirh, line 126: I don't really like this for a couple reasons:
1. this is actually hiding more (should be less) the purpose of the test. Test code should be way of factoring and adding helpers, because the test code has a different purpose than production code (https://about.sourcegraph.com/go/advanced-testing-in-go/#Repeat-yourself)
2. this doesn't seem to return regexes - rather plain strings and the purpose of the code is not really clear here

fromanirh (commented):
if there are urgent fixes (e.g. to unblock CI), I'd split from this PR, because some areas require a bit more work.

functests/5_latency_testing/latency_testing.go
```diff
@@ -267,3 +270,22 @@ func getNegativeTests(toolToTest string) []latencyTest {
...
+	case hwlatdetect:
+		thr := tst.testMaxLatency
+		if tst.hwlatdetectMaxLatency != "" {
+			thr = tst.hwlatdetectMaxLatency
+		}
+		return []string{"--duration " + tst.testRuntime, "--threshold " + thr, "--hardlimit " + thr, "--window ", "--width "}
+	default:
+		return nil
+	}
+	return nil
```
yanirq, line 290: this is probably the unreachable code that the prow-images lane is complaining about.

functests/5_latency_testing/latency_testing.go
shajmakh: it is either 0 or 1 are the invalid for oslat, as it works fine with at least 2 CPUS

functests/5_latency_testing/latency_testing.go
shajmakh: there is a constant for each tool name, just for simplicity and better maintenance.
https://github.com/openshift-kni/performance-addon-operators/blob/a6ae635a2d4d099c5412b24f42184634ce51f2ac/functests/5_latency_testing/latency_testing.go#L19

functests/5_latency_testing/latency_testing.go
shajmakh: thanks for pointing to that!

functests/5_latency_testing/latency_testing.go
fromanirh: this link should work: https://about.sourcegraph.com/go/advanced-testing-in-go/#Repeat-yourself

functests/4_latency/latency.go
```diff
@@ -459,7 +460,9 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
 		testlog.Error(err)
 		logEventsForPod(testPod)
 	}
-	Expect(err).ToNot(HaveOccurred())
+	updatedPod := &corev1.Pod{}
+	_ = testclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(testPod), updatedPod)
```
yanirq, line 464: do we really need an empty assignment here ?

functests/5_latency_testing/latency_testing.go
```diff
@@ -122,8 +123,27 @@ var _ = table.DescribeTable("Test latency measurement tools tests", func(testGro
...
+					thr := test.testMaxLatency
+					if test.hwlatdetectMaxLatency != "" {
+						thr = test.hwlatdetectMaxLatency
+					}
+					passedArgs = []string{"--duration " + test.testRuntime, "--threshold " + thr, "--hardlimit " + thr, "--window ", "--width "}
+				default:
+					testlog.Error("the tool to test was not set")
+				}
+
+				for _, argText := range passedArgs {
```
yanirq, line 144: suggestion:    s/argText/arguments

functests/4_latency/latency.go
shajmakh: I don't see the point of checking that twice (already checked in  https://github.com/shajmakh/performance-addon-operators/blob/0f197be30938d8e25b7556eab7667349614f93e8/functests/utils/pods/pods.go#L114) and we surely don't want to fail the test for this as it is used only for reporting.

fromanirh: I think the spirit of the comment was that you can just use
```
testclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(testPod), updatedPod)
```
(provided the linter likes this way)

functests/5_latency_testing/latency_testing.go
shajmakh: I'm sorry didn't get that, can you elaborate please?

functests/5_latency_testing/latency_testing.go
shajmakh: following the commit: https://github.com/openshift-kni/performance-addon-operators/pull/891/commits/0f197be30938d8e25b7556eab7667349614f93e8
this test set is no longer relevant. Will address an alternative in a later PR when managing the PP within the suite.

functests/4_latency/latency.go
shajmakh: oh i see, makes sense, will fix it

functests/5_latency_testing/latency_testing.go
yanirq: Just a suggestion to change `argText` to `argument`   (to keep it simple and readable)

functests/5_latency_testing/latency_testing.go
```diff
@@ -199,32 +199,32 @@ func clearEnv() {
...
 	//testCpus: for tests that expect a success output message, note that an even CPU number is needed, otherwise the test would fail with SMTAlignmentError
-	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, testCpus: "2", outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, testCpus: "6", outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "1", testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "60", testRun: "true", testRuntime: "2", testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
+	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "30", testMaxLatency: guaranteedLatency, testCpus: "2", outputMsgs: []string{success}, toolToTest: toolToTest})
```
Tal-or, line 225: same comment as https://github.com/openshift/cluster-node-tuning-operator/pull/368/commits/ec3a9dc4c1a1083eb58a8c594424add5e0016c3d#r890002911

functests/utils/pods/pods.go
```diff
@@ -121,6 +123,14 @@ func WaitForPhase(pod *corev1.Pod, phase corev1.PodPhase, timeout time.Duration)
...
 		return false, nil
 	})
+	if err != nil {
+		updatedPod := &corev1.Pod{}
+		if err := testclient.Client.Get(context.TODO(), key, updatedPod); err != nil {
+			testlog.Errorf("could not get the pod's object: %v", err)
+		}
+		testlog.Errorf("Expected the pod to reach phase %s but its phase is %s", phase, updatedPod.Status.Phase)
+	}
+	return err
```
shajmakh, line 133: I did this change in WaitForPhase and not WaitForPredicate because in the predicate:
1. we need to pass a pred function, which will be actually the same functionality we are checking in WaitForPhase, so avoid duplicated code for the pred function.
2. the Function can be checking anything so pointing to the failing condition of pod's phase in WaitForPhase is more clear.

shajmakh:
This PR is the same as 4.11 one (https://github.com/openshift/cluster-node-tuning-operator/pull/368) except that 4.11 has few TODO comments that are relevant for 4.11 only

fromanirh (commented):
/lgtm
/approve
the same changes are queued for 4.11 but blocked until we branch out. Let's move forward and unblock the lanes.

PR #893, opened 2022-06-01
e2e: Latency: initialize PP to avoid nil pointer 

shajmakh:
Following https://github.com/openshift-kni/performance-addon-operators/pull/889, running a latency tool test with no CPUs specified is currently panicking on nil pointer of the performance profile as it was not initialized.

Initialize the PP before checking if to skip the test due to odd CPU number.

### DIFF
--- a/functests/5_latency_testing/latency_testing.go
+++ b/functests/5_latency_testing/latency_testing.go
@@ -214,7 +214,7 @@ func getValidValuesTests(toolToTest string) []latencyTest {
 	if toolToTest == oslat {
 		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: "10", testMaxLatency: "1", oslatMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
 		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: "10", oslatMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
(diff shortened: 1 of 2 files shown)


shajmakh:
complement PR to https://github.com/openshift-kni/performance-addon-operators/pull/889

functests/5_latency_testing/latency_testing.go
```diff
@@ -214,7 +214,8 @@ func getValidValuesTests(toolToTest string) []latencyTest {
 	if toolToTest == oslat {
 		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: "10", testMaxLatency: "1", oslatMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
 		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: "10", oslatMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
-		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, testCpus: "1", outputMsgs: []string{skip, skipOslatCpuNumber}, toolToTest: toolToTest})
+		//TODO: update isolated CPUs in PP to 1 and restore the original set post test
+		//testSet = append(testSet, latencyTest{testRun: "true", testRuntime: "10", oslatMaxLatency: guaranteedLatency, outputMsgs: []string{skip,skipOslatCpuNumber}, toolToTest: toolToTest})
```
fromanirh, line 218: (non blocking if we address it in #889 ) why commented and not deleted?

shajmakh: unfortuanatly I didn't address it there, and the comment is a little different from the original test set,now it points to the correct form of the test set once the TODO is done

fromanirh (commented):
/lgtm

PR #895, opened 2022-06-08
Gather all pods information.

jlojosnegros:
To be able to check if IRQs affinity lists collide with those CPUs exclusively assigned to a "Guaranteed" pod we need to have the QoSClass for each pod in the cluster.

This change will save each of the pods manifests in `must-gather/namespaces/<namespace>/core/pods/*.yaml`

see [here](https://gitlab.cee.redhat.com/ccx/ccx-ocp-core/-/merge_requests/480) for more info

### DIFF
--- a/must-gather/collection-scripts/gather
+++ b/must-gather/collection-scripts/gather
@@ -20,6 +20,9 @@ resources+=(performanceprofile)
 # machine/node resources
 resources+=(nodes machineconfigs machineconfigpools featuregates kubeletconfigs tuneds)
 
+# pods information
+resources+=(pods)
+
 # run the collection of resources using must-gather
 for resource in ${resources[@]}; do
   /usr/bin/oc adm inspect --dest-dir must-gather --all-namespaces ${resource}


must-gather/collection-scripts/gather
```diff
@@ -20,6 +20,9 @@ resources+=(performanceprofile)
 # machine/node resources
 resources+=(nodes machineconfigs machineconfigpools featuregates kubeletconfigs tuneds)
 
+# pods information
+resources+=(pods)
```
fromanirh, line 24: does the base must-gather image collect this info?

jlojosnegros: Not really they get pod information from some namespaces but not all of them and to make accurate checks we need to know all the "guaranteed" pods in order to be able to calculate all the exclusively assigned cpus.

jlojosnegros: Not exactly just `openshift-*` namespaces but a list of namespaces ([see](https://github.com/openshift/must-gather/blob/87c1fd3aee6a6ec7359406a14e1ad0d8d1f8784e/collection-scripts/gather#L28)

As far as I understand we need all the exclusively assigned CPUs to intersect them with the IRQs affinity lists, and to get them we need to know at least all the containers with an integer CPU request in any "guaranteed" pod.

marioferh:
the test are failing due to PAO is now in NTO, tests should be override

must-gather/collection-scripts/gather
marioferh: named_resources+=(ns/default ns/openshift ns/kube-system ns/openshift-etcd)
https://github.com/openshift/must-gather/blob/master/collection-scripts/gather#L28

PR #896, opened 2022-06-09
[release-4.10]  Bug 2055267: latency-tests: oslat: increase minimal number of CPUs

Tal-or:
Oslat should exclude the sibling cpu of the main thread, hence we need one additional isolated CPU for making the test meaningful.
If there are not enough isolated CPUs, the test will skip.

### DIFF
--- a/functests/4_latency/latency.go
+++ b/functests/4_latency/latency.go
@@ -42,7 +42,7 @@ const (
 	defaultTestRuntime   = "300"
 	defaultMaxLatency    = -1
 	defaultTestCpus      = -1
-	minCpuAmountForOslat = 2
+	minCpuAmountForOslat = 3
 )
 
 var (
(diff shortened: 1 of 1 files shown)


functests/4_latency/latency.go
```diff
@@ -42,7 +42,7 @@ const (
 	defaultTestRuntime   = "300"
 	defaultMaxLatency    = -1
 	defaultTestCpus      = -1
-	minCpuAmountForOslat = 2
+	minCpuAmountForOslat = 3
```
shajmakh, line 45: 3 is an odd number this will result with SMTAlignment error in pods events, to avoid that we need an even number so it'll start at 4.

shajmakh: please ignore above comment, I am mistaken here, as this only indicates the minimal value and doesn't apply it anywhere

Tal-or: This is also a tricky check, meaning do we want to ensure that an even number of CPUs is about to be tested?

shajmakh: yes, we want to allow tests with even cpus number only, otherwise the test pod will not start because of the SMTAlignment error

Tal-or: But what if SMT alignment is disabled?

shajmakh: is that an option? HT is supported since 4.10 and I guess should also be spilled to 4.11

Tal-or:
Should be handled on the runner side.
I'll issue another PR that should report the SMT alignment error more clearly.

PR #899, opened 2022-06-14
Bug 2100544: Set rps for virtual interfaces only in crio hook

yanirq:
Fix the low-latency-hooks crio script
settings of rps mask for network interfaces (i.e. the veth)
to target only virtual interfaces instead of
all network devices visible in the container since
the physical NICs are already covered by a udev rule

This is a downport of https://github.com/openshift/cluster-node-tuning-operator/pull/371


fromanirh (commented):
/lgtm
/approve

PR #900, opened 2022-06-14
Gather some pod info

jlojosnegros:
Need to gather some new pod information to use it in insight rules later

To check if some ISRs are executing on exclusively assigned CPUs we need to know which CPUs are assigned to Guaranteed pods. To do that we need to cross the qosClass with the PodResources API information where the assigned cpu_ids are.


fromanirh (commented):
please explain a bit in the commit message (and in the PR itself) what "some" informations are and how we plan to use it, possibly mentioning the podresources API gaps and how this would help

marioferh (commented):
/lgtm

remember two ci lanes need override

jlojosnegros:
> thanks for the updates. Can you please also fix the commit message header? (and the PR title?) it would be nice to also mention that
> 
> * we need to deal with older podresources API which provide the esclusive cpuset or the default one, making impossible to learn the exclusive cpus without extra help, provided here
> * we want to provide a unified flow, so we will keep this logic for the time being until k8s < 1.23 goes out of support, but we can drop this extra logic once the minimum supported kube version is 1.23
> * we decided to collect only the minimal amount of extra information needed to cover this gap

Ok working on that.

jlojosnegros:
> /lgtm
> 
> remember two ci lanes need override

@marioferh which two lanes should be overridden?

fromanirh:
/approve
/lgtm
nice updates, thanks.

PR #904, opened 2022-06-15
e2e: latency: Update PP isolated cpus to match test requirement

shajmakh:
Some tests in the suites require setting the number of CPUs to be used by the latency pod. If the available cpus is not sufficient, the tests will be skipped.
The goal of the suite is to verify that the main tests of the latency tools are executed as expected.
 So to overcome this, update the isolated CPUs for this suite to CPU set that guarantees to execute the tests properly, and restore the initial set when finished.

### DIFF
--- a/functests/5_latency_testing/5_latency_testing_suite_test.go
+++ b/functests/5_latency_testing/5_latency_testing_suite_test.go
@@ -3,6 +3,7 @@ package __latency_testing_test
 import (
 	"context"
 	"encoding/json"
+	"reflect"
 	"testing"
 	"time"
 
@@ -12,12 +13,19 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 
+	performancev2 "github.com/openshift-kni/performance-addon-operators/api/v2"
+	testutils "github.com/openshift-kni/performance-addon-operators/functests/utils"
 	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/images"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/junit"
 	testlog "github.com/openshift-kni/performance-addon-operators/functests/utils/log"
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/namespaces"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/nodes"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/profiles"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/profilesupdate"
+
 	ginkgo_reporters "kubevirt.io/qe-tools/pkg/ginkgo-reporters"
 )
 
@@ -26,32 +34,73 @@ var prePullNamespace = &corev1.Namespace{
 		Name: "testing-prepull",
 	},
 }
+var profile *performancev2.PerformanceProfile
 
-var _ = AfterSuite(func() {
-	prePullNamespaceName := prePullNamespace.Name
-	err := testclient.Client.Delete(context.TODO(), prePullNamespace)
-	testlog.Infof("deleted namespace %q err=%v", prePullNamespace.Name, err)
-	Expect(err).ToNot(HaveOccurred())
-	err = namespaces.WaitForDeletion(prePullNamespaceName, 5*time.Minute)
-})
-
-func Test5LatencyTesting(t *testing.T) {
-	RegisterFailHandler(Fail)
-
+var _ = BeforeSuite(func() {
 	if !testclient.ClientsEnabled {
-		t.Fatalf("client not enabled")
+		testlog.Errorf("client not enabled")
+	}
+
+	// update PP isolated CPUs. the new cpu set for isolated should have an even number of CPUs to avoid failing the pod on SMTAlignment error,
+	// and should be greater than what is requested by the test cases in the suite so the test runs properly
+	var err error
+	profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+	Expect(err).ToNot(HaveOccurred())
+	workerNodes, err := nodes.GetByLabels(testutils.NodeSelectorLabels)
+	Expect(err).ToNot(HaveOccurred())
+
+	initialIsolated := profile.Spec.CPU.Isolated
+	initialReserved := profile.Spec.CPU.Reserved
+	//updated both sets to ensure there is no overlap
+	latencyIsolatedSet := performancev2.CPUSet("1-9")
+	latencyReservedSet := performancev2.CPUSet("0")
+
+	totalCpus := cpuset.MustParse(string(latencyIsolatedSet)).Size() + cpuset.MustParse(string(latencyReservedSet)).Size()
+	nodesWithSufficientCpu := nodes.GetByCpuAllocatable(workerNodes, totalCpus)
+	//before applying the changes verify that there are compute nodes with sufficient cpus
+	if len(nodesWithSufficientCpu) != 0 {
+		if *initialIsolated != latencyIsolatedSet || *initialReserved != latencyReservedSet {
+			testlog.Info("Update the isolated and reserved cpus sets of the profile")
+			err = profilesupdate.UpdateIsolatedReservedCpus(latencyIsolatedSet, latencyReservedSet)
+			if err != nil {
+				testlog.Error("could not update the profile with the desired CPUs sets")
+			}
+		}
 	}
 
 	if err := createNamespace(); err != nil {
-		t.Fatalf("cannot create the namespace: %v", err)
+		testlog.Errorf("cannot create the namespace: %v", err)
 	}
 
 	ds, err := images.PrePull(testclient.Client, images.Test(), prePullNamespace.Name, "cnf-tests")
 	if err != nil {
 		data, _ := json.Marshal(ds) // we can safely skip errors
 		testlog.Infof("DaemonSet %s/%s image=%q status:\n%s", ds.Namespace, ds.Name, images.Test(), string(data))
-		t.Fatalf("cannot prepull image %q: %v", images.Test(), err)
+		testlog.Errorf("cannot prepull image %q: %v", images.Test(), err)
 	}
+})
+
+var _ = AfterSuite(func() {
+	prePullNamespaceName := prePullNamespace.Name
+	err := testclient.Client.Delete(context.TODO(), prePullNamespace)
+	if err != nil {
+		testlog.Errorf("namespace %q could not be deleted err=%v", prePullNamespace.Name, err)
+	}
+	namespaces.WaitForDeletion(prePullNamespaceName, 5*time.Minute)
+
+	currentProfile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+	Expect(err).ToNot(HaveOccurred())
+	if reflect.DeepEqual(currentProfile.Spec, profile.Spec) != true {
+		testlog.Info("Restore initial performance profile")
+		err = profilesupdate.ApplyProfile(profile)
+		if err != nil {
+			testlog.Error("could not restore the initial profile")
+		}
+	}
+})
+
+func Test5LatencyTesting(t *testing.T) {
+	RegisterFailHandler(Fail)
 
 	rr := []Reporter{}
 	if ginkgo_reporters.Polarion.Run {
--- a/functests/5_latency_testing/latency_testing.go
+++ b/functests/5_latency_testing/latency_testing.go
@@ -12,6 +12,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+
 	testlog "github.com/openshift-kni/performance-addon-operators/functests/utils/log"
 )
 
@@ -145,9 +146,6 @@ var _ = table.DescribeTable("Test latency measurement tools tests", func(testGro
 					Expect(strings.Contains(string(output), argument)).To(BeTrue(), "The tool command didn't pass the argument %q", argument)
 				}
 			}
-		}
-		for _, msg := range test.outputMsgs {
-			Expect(string(output)).To(MatchRegexp(msg), "The output of the executed tool is not as expected")
 		}
 	}
 },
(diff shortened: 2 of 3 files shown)


functests/utils/profilesupdate/profile_update.go
```diff
@@ -0,0 +1,54 @@
...
+	updatedProfile := profile.DeepCopy()
+	updatedProfile.Spec.CPU = &performancev2.CPU{
+		Isolated: &isolatedSet,
+		Reserved: &reservedSet,
+	}
+
+	ApplyProfile(updatedProfile)
+}
+
+func ApplyProfile(profile *performancev2.PerformanceProfile) {
```
Tal-or, line 33: I would expect this function to return an error status whether the update finished successfully.

Tal-or (commented):
Nice work. I did an initial review and left a few comments.
I'll have another one in the following days.

functests/5_latency_testing/latency_testing.go
```diff
@@ -88,57 +93,96 @@ type latencyTest struct {
...
+			table.Entry("[test_id:42853] Oslat - Verify that the latency tool test should print an expected error message when passing invalid environment variables values", getNegativeTests(oslat), negativeTesting),
+			table.Entry("[test_id:42115] Cyclictest - Verify that the tool is working properly with valid environment variables values", getValidValuesTests(cyclictest), positiveTesting),
+			table.Entry("[test_id:42852] Cyclictest - Verify that the latency tool test should print an expected error message when passing invalid environment variables values", getNegativeTests(cyclictest), negativeTesting),
+			table.Entry("[test_id:42849] Hwlatdetect - Verify that the tool is working properly with valid environment variables values", getValidValuesTests(hwlatdetect), positiveTesting),
+			table.Entry("[test_id:42856] Hwlatdetect - Verify that the latency tool test should print an expected error message when passing invalid environment variables values", getNegativeTests(hwlatdetect), negativeTesting),
+		)
+	})
+
+	Context("Verify tests output with existed performance profile", func() {
+		It("[test_id:---] should skip oslat test when isolated CPUs amount is less than 2", func() {
```
yanirq, line 152: blank test id ?

functests/5_latency_testing/5_latency_testing_suite_test.go
```diff
@@ -26,32 +31,58 @@ var prePullNamespace = &corev1.Namespace{
...
 
-	if !testclient.ClientsEnabled {
-		t.Fatalf("client not enabled")
+	initialIsolated = profile.Spec.CPU.Isolated
+	initialReserved = profile.Spec.CPU.Reserved
+	//updated both sets to ensure there is no overlap
+	latencyIsolatedSet := performancev2.CPUSet("1-9")
+	latencyReservedSet := performancev2.CPUSet("0")
+	profileNeedsUpdate = false
+	//if string(*initialIsolated) != string(latencyIsolatedSet) || string(*initialReserved) != string(latencyReservedSet) {
```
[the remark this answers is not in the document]
shajmakh: nope, removing, thanks for catching that

functests/5_latency_testing/latency_testing.go
shajmakh: not intended, will fix it, thanks for catching that!

shajmakh: not intended, will fix it, thanks for catching that!

shajmakh:
/hold
#891 should be merged first

functests/5_latency_testing/5_latency_testing_suite_test.go
```diff
@@ -26,32 +31,58 @@ var prePullNamespace = &corev1.Namespace{
...
+	if *initialIsolated != latencyIsolatedSet || *initialReserved != latencyReservedSet {
+		testlog.Info(" the profile needs update ")
+		profileNeedsUpdate = true
+		profilesupdate.UpdateIsolatedReservedCpus(latencyIsolatedSet, latencyReservedSet)
 	}
 
 	if err := createNamespace(); err != nil {
-		t.Fatalf("cannot create the namespace: %v", err)
```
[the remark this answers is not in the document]
shajmakh: yes, the namespace is created to hold the prepull daemon to make the process of pulling the pod image faster as it will be already on the nodes. but if prepull is not successful it should not block the testing because the test itself pulls the image:
  https://github.com/openshift-kni/performance-addon-operators/blob/0c9fb1fe23e10f8c89a7f8e5861662bddeb8d7cd/functests/4_latency/latency.go#L452

fromanirh:
please rebase and we will review again

functests/5_latency_testing/5_latency_testing_suite_test.go
```diff
@@ -26,32 +31,58 @@ var prePullNamespace = &corev1.Namespace{
...
+	// and should be greater than what is requested by the test cases in the suite so the test runs properly
+	profile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+	Expect(err).ToNot(HaveOccurred())
 
-	if !testclient.ClientsEnabled {
-		t.Fatalf("client not enabled")
+	initialIsolated = profile.Spec.CPU.Isolated
+	initialReserved = profile.Spec.CPU.Reserved
+	//updated both sets to ensure there is no overlap
+	latencyIsolatedSet := performancev2.CPUSet("1-9")
```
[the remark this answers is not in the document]
shajmakh: Well, I agree it should not be hardcoded, but these tests run only u/s on the QE CI machines, which have at least 12 CPUs. Now if the machine doesn't have this set, an error message will be logged but the test will continue and the irrelevant tests will be skipped.

shajmakh:
the failures seem flaky, the results are healthy locally:

```
Jun 22 13:38:18.587: [INFO]: Update the isolated and reserved cpus sets of the profile: initial isolated %!q(*v2.CPUSet=0xc0003a0980) reserved "0", new values will be: isolated "1-9" reserved "0"
Jun 22 13:38:18.608: [INFO]: Getting MCP for profile
Jun 22 13:38:18.634: [INFO]: Verifying that mcp is ready for update
Jun 22 13:38:18.707: [INFO]: MCP "workercnf" is targeting 2 node(s)
Jun 22 13:38:18.728: [INFO]: Applying changes in performance profile and waiting until mcp will start updating
Jun 22 13:38:18.860: [INFO]: MCP "workercnf" is targeting 2 node(s)
Jun 22 13:38:48.903: [INFO]: Waiting when mcp finishes updates
Jun 22 13:38:48.995: [INFO]: MCP "workercnf" is targeting 2 node(s)
```

functests/5_latency_testing/5_latency_testing_suite_test.go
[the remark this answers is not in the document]
fromanirh: we can (and should) add a check:
1. iterate over the worker nodes
2. check all/some/at least one of them has at least the amount of CPUs we need (let's check allocatable, not capacity)
3. skip if we don't have enough worker nodes

functests/utils/profilesupdate/profile_update.go
```diff
@@ -0,0 +1,75 @@
...
+
+	ok := ApplyProfile(updatedProfile)
+	if ok {
+		testlog.Infof("successfully updated performance profile %q with new isolated cpus set: %q and new reserved cpus set: %q", profile.Name, updatedProfile.Spec.CPU.Isolated, updatedProfile.Spec.CPU.Reserved)
+	}
+	return ok
+}
+
+//ApplyProfile applies the new profile and returns true if the changes were applied indeed and false otherwise
+func ApplyProfile(profile *performancev2.PerformanceProfile) bool {
```
fromanirh, line 42: we should return error and not bool. Success should be conveyed with err == nil, failure otherwise. This should enable overall code simplification

functests/utils/profilesupdate/profile_update.go
```diff
@@ -0,0 +1,75 @@
...
+	testclient "github.com/openshift-kni/performance-addon-operators/functests/utils/client"
+	testlog "github.com/openshift-kni/performance-addon-operators/functests/utils/log"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/mcps"
+	"github.com/openshift-kni/performance-addon-operators/functests/utils/profiles"
+	"github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components"
+	profilecontroller "github.com/openshift-kni/performance-addon-operators/pkg/controller/performanceprofile/components/profile"
+)
+
+//UpdateIsolatedReservedCpus Updates the current performance profile with new sets of isolated and reserved cpus, and returns true if the update was successfull and false otherwise
+func UpdateIsolatedReservedCpus(isolatedSet performancev2.CPUSet, reservedSet performancev2.CPUSet) bool {
```
fromanirh, line 23: we should return error and not bool, Success should be conveyed with err == nil, failure otherwise.

functests/5_latency_testing/latency_testing.go
```diff
@@ -160,6 +164,41 @@ var _ = Describe("Run tests of latency measurement tools with different values o
 			table.Entry("[test_id:42856] Hwlatdetect - Verify that the latency tool test should print an expected error message when passing invalid environment variables values", getNegativeTests(hwlatdetect), negativeTesting),
 		)
 	})
+
+	Context("Verify tests output with updated performance profile", func() {
+		It("[test_id:52388] should skip oslat test when isolated CPUs amount is less than 2", func() {
```
fromanirh, line 169: is this a backport of a existing test?

functests/5_latency_testing/5_latency_testing_suite_test.go
```diff
@@ -26,32 +31,69 @@ var prePullNamespace = &corev1.Namespace{
 		Name: "testing-prepull",
 	},
 }
+var initialIsolated, initialReserved *performancev2.CPUSet
+var profileNeedsUpdate bool = false
```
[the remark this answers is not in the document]
fromanirh: ok, we can keep them for starters

functests/5_latency_testing/latency_testing.go
shajmakh: no this is a new test, we need to have it in 4.11 as well. currently 4.11 latency patches are waiting for the branching to avoid creating BZ, so testing this on 4.10 first once approved will create a PR for 4.11

functests/5_latency_testing/latency_testing.go
```diff
@@ -160,6 +164,41 @@ var _ = Describe("Run tests of latency measurement tools with different values o
...
+			profile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+			Expect(err).ToNot(HaveOccurred())
+			initialIsolated := profile.Spec.CPU.Isolated
+			initialReserved := profile.Spec.CPU.Reserved
+			//updated both sets to ensure there is no overlap
+			latencyIsolatedSet := performancev2.CPUSet("1")
+			latencyReservedSet := performancev2.CPUSet("0")
+			profilesupdate.UpdateIsolatedReservedCpus(latencyIsolatedSet, latencyReservedSet)
```
fromanirh, line 177: I'm very wary of tests which change their environment. This is intrisincally more fragile and slow. Can we maybe start _checking_ the cluster is not properly configuring and skipping/failing if not the case?

functests/5_latency_testing/latency_testing.go
```diff
@@ -160,6 +164,41 @@ var _ = Describe("Run tests of latency measurement tools with different values o
...
+			testDescription := setEnvAndGetDescription(test)
+			By(testDescription)
+			if _, err := os.Stat("../../build/_output/bin/latency-e2e.test"); os.IsNotExist(err) {
+				Skip("The executable test file does not exist , skipping the test.")
+			}
+			output, err := exec.Command("../../build/_output/bin/latency-e2e.test", "-ginkgo.focus", test.toolToTest).Output()
+			if err != nil {
+				testlog.Error(err.Error())
+			}
+			Expect(string(output)).NotTo(MatchRegexp(unexpectedError), "Unexpected error was detected in a positive test")
```
fromanirh, line 195: In this specific case we can maybe simplify like
```
Expect(string(output)).NotTo(MatchRegexp(unexpectedError), "Unexpected error was detected in a positive test: %v", err)
```

functests/utils/profilesupdate/profile_update.go
shajmakh: sure, done. Thanks!

functests/5_latency_testing/latency_testing.go
shajmakh: if we don't change the environment this test will always be skipped.

functests/5_latency_testing/latency_testing.go
```diff
@@ -89,76 +90,77 @@ type latencyTest struct {
...
-		if _, err := os.Stat("../../build/_output/bin/latency-e2e.test"); os.IsNotExist(err) {
-			Skip("The executable test file does not exist , skipping the test.")
-		}
-		output, err := exec.Command("../../build/_output/bin/latency-e2e.test", "-ginkgo.focus", test.toolToTest).Output()
-		if err != nil {
-			//we don't log Error level here because the test might be a negative check
-			testlog.Info(err.Error())
-		}
+var _ = Describe("Run tests of latency measurement tools with different values of latency environment variables", func() {
+	Context("Verify tests output with existed performance profile", func() {
```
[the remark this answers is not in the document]
shajmakh: we don't need it in the first commit, but we do need it in the second one when adding a new test outside the describetable.I moved the change accordingly.

functests/5_latency_testing/latency_testing.go
fromanirh: ok, that's fair. But do we need to change the environment _within_ the test? any alternatives?

shajmakh: because oslat requires currently for 4.10 at least  >=2, that leaves us with 1 for this test, but setting 1 will be skipped because of the SMTalignment.
@Tal-or are we planning to change oslat minimum cpus to 3 in 4.10? don't we have bug for that? 
and what about  4.9 do we expect any changes?

Tal-or: > @Tal-or are we planning to change oslat minimum cpus to 3 in 4.10? don't we have bug for that?

We do. once the fix is in I'll backport to 4.10

> and what about 4.9 do we expect any changes?

Nothing that I'm aware of

shajmakh: thanks for the info Talor, @fromanirh shall we keep this new test or wait for the bug fix?
(if we keep it and re-adjust it later it means replacing the whole [new context](https://github.com/openshift-kni/performance-addon-operators/pull/904/commits/04715f26f102aa4669692148a9347a18cea63c51#diff-430c986deb22f7ffd2962b8759ac1068ad8d55477c7588d04163522a33d85af0R168) with one line like other [testcases](https://github.com/openshift-kni/performance-addon-operators/pull/904/commits/04715f26f102aa4669692148a9347a18cea63c51#diff-430c986deb22f7ffd2962b8759ac1068ad8d55477c7588d04163522a33d85af0R276), but OTOH if we don't merge it we don't have a test covering this scenario (which is a minor case) until the new fix lands.

fromanirh: it seems the lesser evil is to merge this test as-is and re-evaluate later

PR #906, opened 2022-06-16
Improve pod information to check IRQ Load Balancing

jlojosnegros:
Added "resources.requests.cpu" and "resources.limits.cpu" information
for each of the containers on each pod.
This information is required to look for exclusively assigned CPUs.

CPUs are assigned to be used exclusively only to those containers
belonging to a pod with a "Guaranteed" qosClass and with an integer cpu
request
Thats why we need to add this information to must-gather.

### DIFF
--- a/must-gather/collection-scripts/gather_nodes
+++ b/must-gather/collection-scripts/gather_nodes
@@ -71,7 +71,7 @@ do
 
     oc exec $pod -n perf-node-gather -- gather_sysinfo snapshot --debug --root=/host --output=- > $NODE_PATH/sysinfo.tgz 2> $NODE_PATH/sysinfo.log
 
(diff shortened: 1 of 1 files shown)


jlojosnegros:
cc @fromanirh

marioferh (commented):
checked it and working
/lgtm

fromanirh:
so we need to cover the fact that is not enough to have the pod QoS to learn which pods got exclusive cpus, because old podresources API will return the exclusive set OR the default, and without knowing if the containers requested integral cpus or not, we still don't have enough infos.

This is unfortunate, but the need is clear and very legit.
However, this is not the right direction to solve this. In general, we should have less `oc` calls, not more. The previous PR (#900) was already an exception. We're doubling down in the wrong direction.

This extra info is needed to cover a podresources gap. It is better collected in `gather-sysinfo`.
If we are in a rush to complete, we can have one more exception, iff we track this technical debt (filing issues) and we come up with a plan to remove it in the future.

jlojosnegros:
I am not sure if we are in a rush ... I mean, the final target is to improve our check for IRQ load balancing and as it has been as it is now for a while I think we are not in a rush and we can do this in the proper way.
But that is my impression, maybe @MarSik could give us another point of view.

PR #910, opened 2022-06-22
must-gather: add podresources command

fromanirh:
Backport podresources and podinfo command


tools/gather-sysinfo/gather-sysinfo.go
```diff
@@ -20,7 +21,11 @@ type snapshotOptions struct {
 }
 
 func main() {
-	root := cmd.NewRootCommand(newSnapshotCommand)
+	root := cmd.NewRootCommand(
+		k8s.NewPodResourcesCommand,
```
jlojosnegros, line 25: More of a question than a comment. 
Why should we add these two commands but we do not need to add the other commands from debug-tools we also use?

fromanirh: because the other commands are automatically and implicitely registered: https://github.com/openshift-kni/debug-tools/blob/main/pkg/knit/cmd/root.go#L82
OTOH commands which will require a large set of deps (k8s-based commands are a egregious example) need to be registered explicitely, to give to the client code the choice if the large deps should be pulled in or not

PR #916, opened 2022-06-23
[release 4.10] Bug 2100544: Fix RPS default physical and virtual settings

yanirq:
Issue:
1. The RPS cpu mask was set to the reserved CPUs taken from the performance profile for all interfaces by default.
2. A background script that runs as an oci pre-start hook (low-latency-hooks) for guaranteed pods sets the rps mask of all network devices visible from the root path of devices (/sys/devices).

Fix:
1. Set RPS cpu mask to reserved CPUs taken from the performance profile excluding physical and veth interfaces.
2. Change the devices path in the script to set rps mask only for virtual interfaces.

Added special annotation capability:
A new annotation option to the performance profile: `"performance.openshift.io/enable-physical-dev-rps"`.
When added to a performance profile,it will enable RPS mask setting with systemd for all network devices by including physical interfaces from netdev-rps rule.


pkg/controller/performanceprofile/components/profile/profile.go
```diff
@@ -54,3 +54,16 @@ func IsPaused(profile *performancev2.PerformanceProfile) bool {
...
 	return false
 }
+
+// IsRpsEnabled or not RPS mask should be set for all physical net devices
+func IsRpsEnabled(profile *performancev2.PerformanceProfile) bool {
+	if profile.Annotations == nil {
+		return false
+	}
+	isRpsEnabled, ok := profile.Annotations[performancev2.PerformanceProfileEnableRpsAnnotation]
+	if ok && isRpsEnabled == "true" {
```
fromanirh, line 64: we didn't use `strconv.ParseBool` in the end, didn't we?

fromanirh (commented):
/lgtm
/approve

pkg/controller/performanceprofile/components/profile/profile.go
yanirq: at the end we did'nt.  Its using the same annotation logic we have in the function above (isPaused)

fromanirh: too bad, but no place to fix this in a backport.

PR #911, opened 2022-06-22
[release-4.8][manual] Backport podresources and podinfo commands

jlojosnegros:

PR #918, opened 2022-06-27
[release 4.8] Modify regex to reflect new "data"

jlojosnegros:
Due to some dependency update profile.spec.data field is now written as
a multiple lines string instead of as a single line string.
We are using regex **in tests** to check for some content in this field and these new
line feeds could make them fail so we have updated the regex to check
only the data we are interested on despite what could be in between.

See https://github.com/openshift-kni/performance-addon-operators/pull/911#discussion_r904649553

### DIFF
--- a/pkg/controller/performanceprofile/components/tuned/tuned_test.go
+++ b/pkg/controller/performanceprofile/components/tuned/tuned_test.go
@@ -173,7 +173,7 @@ var _ = Describe("Tuned", func() {
 					reservedSet, err := cpuset.Parse(string(*profile.Spec.CPU.Reserved))
 					Expect(err).ToNot(HaveOccurred())
 					reserveCPUcount := reservedSet.Size()
-					channelsRegex := regexp.MustCompile(`\s*channels=combined ` + strconv.Itoa(reserveCPUcount) + `\s*`)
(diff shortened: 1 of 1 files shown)


jlojosnegros:
/assign @fromanirh
https://github.com/openshift-kni/performance-addon-operators/pull/911#discussion_r904649553

fromanirh (commented):
/cc @yanirq

marioferh:
maybe it worth mention that its a test change

jlojosnegros:
> maybe it worth mention that its a test change

done in PR message

yanirq:
/lgtm
lets see the e2e tests pass + worth testing this with #911 locally with running `make unittests`

jlojosnegros:
> lets see the e2e tests pass + worth testing this with https://github.com/openshift-kni/performance-addon-operators/pull/911 locally with running make unittests

That was already done. This change was done in #911 to solve the failure of the tests once dependencies were updated but it was change to its own PR.

fromanirh:
/lgtm
/approve
next time please consider adding tags in the commit message header to identify the area, like
`tuned: fix regex to catch multi-line data`
`tuned: controller: fix regex to catch multi-line data`

PR #919, opened 2022-07-07
Gather NRO related data

jlojosnegros:
Need to gather data regarding new NUMA resources operator but currently
must-gather does not collect this data.
Different approachs were discussed but in the end we have decided to
collect NRO data here to avoid creating new must-gather images.


must-gather/collection-scripts/gather_nro
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# get crds
+for crd_name in $(oc get crds -o=custom-columns=NAME:.metadata.name --no-headers | grep -E "^*\.nodetopology\.openshift\.io$")
+do
+    oc adm inspect --dest-dir must-gather crd/${crd_name}
+done
+
+# Get NRO namespace and gather all the data there
+. namespace
```
Tal-or, line 10: Maybe I'm missing something here, but this is a bash script.
Is there such a command ```. namespace```?

marioferh: Yep, executing this: https://github.com/openshift-kni/performance-addon-operators/blob/master/must-gather/collection-scripts/namespace

marioferh (commented):
/lgtm

must-gather/collection-scripts/namespace
```diff
@@ -11,3 +11,19 @@ function namespace() {
...
+  ns=$( oc get subs -A --field-selector metadata.name='numaresources-operator-subscription' -o=jsonpath='{.items[0].metadata.namespace}{"\n"}' 2> /dev/null )
+
+  # trying again with the pods, which are _usually_ reliable - but users can change them
+  [ -z "${ns}" ] && ns=$( oc get pods -A -l name='resource-topology' -o=jsonpath='{.items[0].metadata.namespace}{"\n"}' 2> /dev/null ) 
+
+  # namespace suggested by the documentation. This is a fancier way to check for its existence
+  [ -z "${ns}" ] && ns=$(oc get ns openshift-numaresources -o=jsonpath='{.metadata.name}{"\n"}' 2> /dev/null )
+
+  # we should never get there. This is the last resort.
+  [ -z "${ns}" ] && ns="numaresources"
```
fromanirh, line 26: should be `openshift-numaresources`, or no fallback. On openshift platform we want `openshift-` prefixed namespaces for infra-like components (and a new scheduler totally is)

must-gather/collection-scripts/gather_nro
fromanirh: the command is actually `.` (dot), aliased by `source`, which means: "read and execute in the current shell context the given file"

Tal-or: But the function name has just changed

must-gather/collection-scripts/gather_nro
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# get crds
+for crd_name in $(oc get crds -o=custom-columns=NAME:.metadata.name --no-headers | grep -E "^*\.nodetopology\.openshift\.io$")
```
fromanirh, line 4: let's also make sure we collect `noderesourcetopologies.topology.node.k8s.io`

fromanirh: Additionally, does this collect also the instances or just the definition?

fromanirh (commented):
nice! how was this tested?

must-gather/collection-scripts/gather_nro
jlojosnegros: Maybe not the best filename to avoid confusions.

must-gather/collection-scripts/gather_nro
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+crdoi_list=()
```
fromanirh, line 3: crdoi?

jlojosnegros:
> nice! how was this tested?

I ran it against a cluster Shereen kindly lent to me with NRO already installed. I have also ran it against a "kind" cluster after running e2e tests but without cleaning up so it had crds instances

must-gather/collection-scripts/gather_nro
```diff
@@ -0,0 +1,21 @@
...
+crdoi_list+=$(oc get crds -o=custom-columns=NAME:.metadata.name --no-headers | grep -E "^*\.nodetopology\.openshift\.io$")
+crdoi_list+=( noderesourcetopologies.topology.node.k8s.io )
```
fromanirh, line 5: we know the names of all the crds we need, and these names are expected to be stable. What's the advantage of autodetecting some and hardcoding some other? I'd just hardcoded all of them.

must-gather/collection-scripts/gather_nro
```diff
@@ -0,0 +1,21 @@
...
+
+# get crds
+for crd_name in ${crdoi_list[@]}
+do
+    # get crd definition
+    oc adm inspect --dest-dir must-gather crd/${crd_name}
+
+    # get crd instances
+    oc adm inspect --dest-dir must-gather ${crd_name}
```
fromanirh, line 14: what's the benefit of `inspect` over plain `get -o yaml`?

fromanirh (commented):
thanks for the updates! how was this tested?

fromanirh:
> > nice! how was this tested?
> 
> I ran it against a cluster Shereen kindly lent to me with NRO already installed. I have also ran it against a "kind" cluster after running e2e tests but without cleaning up so it had crds instances

ok, so we don't have e2e coverage.

fromanirh:
I'm trying to think how a e2e test could look like for this change.
We should probably have the test running in the NROP CI (not PAO's), and should run like nightly or weekly - since it cannot run on PR changes - different repo.

must-gather/collection-scripts/gather_nro
jlojosnegros: `inspect` is the command commonly used by all the `must-gather` scripts. It defaults to a plain `get` but could execute predefined behaviours with some resource types to collect additional resources related with the one asked. 
So in the "worst" case ( not having any behaviour defined ) we get the same we would obtain with `oc get`, on the other hand we could also benefit from any specific behaviour defined.
Also `inspect` uses a defined folder layout to save the collected info to disk if we use `oc get` we should mimic that layout and be aware of any change on it so sync with

must-gather/collection-scripts/gather_nro
jlojosnegros: crd Of Interest ...

fromanirh: this deserves a comment

must-gather/collection-scripts/gather_nro
fromanirh: fair enough, let's keep `inspect` then

fromanirh (commented):
thanks, LGTM. I'll verify ASAP, but let's file a issue to track lack of e2e tests, so we can merge (and then add the e2e tests).

jlojosnegros:
> thanks, LGTM. I'll verify ASAP, but let's file a issue to track lack of e2e tests, so we can merge (and then add the e2e tests).

Issue created #920

fromanirh:
> > thanks, LGTM. I'll verify ASAP, but let's file a issue to track lack of e2e tests, so we can merge (and then add the e2e tests).
> 
> Issue created #920

I think it's more fitting for the numaresources operator, though. We're somehow halfway, but most of the resources, the pieces and the infra are more readily available on NRO CI than setting up a completely alien flow in the PAO CI.

fromanirh:
/override ci/prow/e2e-gcp
still haven't had the chance to verify

fromanirh:
/lgtm
/approve
verified!

PR #929, opened 2022-08-09
e2e: latency: skip the tests earlier in the suite

shajmakh:
This suite works properly only if the test executable exists. Currently each test checks for the existence of this common executable.
The suite performs some setup steps that are relevant for this set of tests, without considering whether the file exists or not. To avoid extra unneeded execution time, skip the suite at an early stage if the executable file does not exist.

Signed-off-by: shereenH <shajmakh@redhat.com>

### DIFF
--- a/functests/5_latency_testing/5_latency_testing_suite_test.go
+++ b/functests/5_latency_testing/5_latency_testing_suite_test.go
@@ -3,6 +3,7 @@ package __latency_testing_test
 import (
 	"context"
 	"encoding/json"
+	"os"
 	"reflect"
 	"testing"
 	"time"
@@ -29,6 +30,9 @@ import (
 	ginkgo_reporters "kubevirt.io/qe-tools/pkg/ginkgo-reporters"
 )
 
+//TODO get commonly used variables from one shared file that defines constants
+const filePath = "../../build/_output/bin/latency-e2e.test"
+
 var prePullNamespace = &corev1.Namespace{
 	ObjectMeta: metav1.ObjectMeta{
 		Name: "testing-prepull",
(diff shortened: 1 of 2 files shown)


functests/5_latency_testing/5_latency_testing_suite_test.go
```diff
@@ -29,26 +30,64 @@ import (
 	ginkgo_reporters "kubevirt.io/qe-tools/pkg/ginkgo-reporters"
 )
 
+//TODO get commonly used variables from one shared file that defines constants
+const filePath = "../../build/_output/bin/latency-e2e.test"
```
MarSik, line 34: You probably want a different variable name here. Maybe something like `testExecutablePath`?

functests/5_latency_testing/latency_testing.go
```diff
@@ -20,7 +20,9 @@ import (
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/profilesupdate"
 )
 
+//TODO get commonly used variables from one shared file that defines constants
 const (
+	filePath = "../../build/_output/bin/latency-e2e.test"
```
MarSik, line 25: Same comments as above.

functests/5_latency_testing/5_latency_testing_suite_test.go
```diff
@@ -29,26 +30,64 @@ import (
...
+
 var prePullNamespace = &corev1.Namespace{
 	ObjectMeta: metav1.ObjectMeta{
 		Name: "testing-prepull",
 	},
 }
 var profile *performancev2.PerformanceProfile
 
-var _ = BeforeSuite(func() {
+func Test5LatencyTesting(t *testing.T) {
```
MarSik, line 43: Is it a good idea to drop the Ginkgo features (Before, After etc?). Won't we lose functionality like focusing or something else this way?

shajmakh: no, those are optional nodes, so ginkgo features will still work with having as minimal as the main TestXXX(t *testing.T) only.

functests/5_latency_testing/5_latency_testing_suite_test.go
shajmakh: addressed in https://github.com/openshift-kni/performance-addon-operators/pull/930, Thanks!

shajmakh:
the commit is split into several targeted commits in https://github.com/openshift-kni/performance-addon-operators/pull/930
closing this PR.

PR #930, opened 2022-08-10
e2e: latency: skip the tests earlier in the suite

shajmakh:
This suite works properly only if the test executable exists. Currently each test checks for the existence of this common executable.
The suite performs some setup steps that are relevant for this set of tests, without considering whether the file exists or not. To avoid extra unneeded execution time, skip the suite at an early stage if the executable file does not exist.

### DIFF
--- a/functests/5_latency_testing/5_latency_testing_suite_test.go
+++ b/functests/5_latency_testing/5_latency_testing_suite_test.go
@@ -43,6 +43,10 @@ var profile *performancev2.PerformanceProfile
 func Test5LatencyTesting(t *testing.T) {
 	RegisterFailHandler(Fail)
 
+	if !testclient.ClientsEnabled {
+		t.Fatal("client not enabled")
+	}
+
 	if _, err := os.Stat(testExecutablePath); os.IsNotExist(err) {
 		testlog.Infof("The test executable file %q does not exist, skipping the suite.", testExecutablePath)
 		t.Skip()
@@ -70,10 +74,6 @@ func createNamespace() error {
 }
 
 func setup() {
-	if !testclient.ClientsEnabled {
-		testlog.Errorf("client not enabled")
-	}
-
 	// update PP isolated CPUs. the new cpu set for isolated should have an even number of CPUs to avoid failing the pod on SMTAlignment error,
 	// and should be greater than what is requested by the test cases in the suite so the test runs properly
 	var err error
--- a/functests/5_latency_testing/5_latency_testing_suite_test.go
+++ b/functests/5_latency_testing/5_latency_testing_suite_test.go
@@ -36,7 +36,33 @@ var prePullNamespace = &corev1.Namespace{
 }
 var profile *performancev2.PerformanceProfile
 
-var _ = BeforeSuite(func() {
+func Test5LatencyTesting(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	//TODO Skip the suite before setup steps in case the executable is not found
+
+	setup()
+	defer teardown()
+
+	rr := []Reporter{}
+	if ginkgo_reporters.Polarion.Run {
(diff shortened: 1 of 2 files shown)


shajmakh:
/hold 
do not merge until 4.11 batch is merged:
https://github.com/openshift/cluster-node-tuning-operator/pull/429

functests/5_latency_testing/5_latency_testing_suite_test.go
```diff
@@ -45,9 +45,15 @@ var _ = BeforeSuite(func() {
 	// and should be greater than what is requested by the test cases in the suite so the test runs properly
 	var err error
 	profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
-	Expect(err).ToNot(HaveOccurred())
```
Tal-or, line 48: Could you please explain why this change is needed?

functests/5_latency_testing/5_latency_testing_suite_test.go
```diff
@@ -36,7 +36,33 @@ var prePullNamespace = &corev1.Namespace{
...
 var profile *performancev2.PerformanceProfile
 
-var _ = BeforeSuite(func() {
+func Test5LatencyTesting(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	//TODO Skip the suite before setup steps in case the executable is not found
+
+	setup()
+	defer teardown()
```
Tal-or, line 56: Are we sure that teardown would run in case of test failure, now that it's not part of the AfterSuite?

functests/5_latency_testing/5_latency_testing_suite_test.go
```diff
@@ -39,7 +43,10 @@ var profile *performancev2.PerformanceProfile
 func Test5LatencyTesting(t *testing.T) {
 	RegisterFailHandler(Fail)
 
-	//TODO Skip the suite before setup steps in case the executable is not found
+	if _, err := os.Stat(testExecutablePath); os.IsNotExist(err) {
+		testlog.Infof("The test executable file %q does not exist, skipping the suite.", testExecutablePath)
+		t.Skip()
```
Tal-or, line 52: Is there a valid case where we expect the executable to be missing? 
If the reason for missing executable is due to something bad that happened, and not an expected scenario I think we should fail here and not skip.

Tal-or: If we fail and do not skip it can be done in the ```BeforeSuite```

shajmakh: the executable is generated via a make target, if something interrupts the creation of the binary then it would mostly be external from the tests, thus it tests should not fail as their intent is to run the provided binary with different environment variables. 
the make target to build the execfile: https://github.com/openshift-kni/performance-addon-operators/blob/release-4.10/Makefile#L115

functests/5_latency_testing/5_latency_testing_suite_test.go
shajmakh: following the conversation on the team channel: 
defer will not run if an interruption happened during the tests or if an assertion failed. thus, using defer here is not relevant.
Thanks Talor for catching this!

functests/5_latency_testing/5_latency_testing_suite_test.go
shajmakh: had this done in a separete commit to having the changes in this PR in groups. 
it was needed because it is not recommended to have Gomega assertions in the main ..Testing(t ..) function. it w

shajmakh:
following the conversation here and on the team channel, using defere() will not always clean up the env in case of interruptions and/or assertion failures, thus closing this pr in favor of #938 that introduce another alternative skipping method while preserving the before/afterSuites

PR #934, opened 2022-08-16
"Add imagePullPolicy verification test"

shaior:
Automates https://bugzilla.redhat.com/show_bug.cgi?id=2055019

### DIFF
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -1088,6 +1088,7 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 
 	Context("Verify containers specifications", func() {
 
+		// Automates https://bugzilla.redhat.com/show_bug.cgi?id=2055019
 		It("imagePullPolicy should be ifNotPresent", func() {
 			pod, err := pods.GetPerformanceOperatorPod()
 			Expect(err).ToNot(HaveOccurred())
--- a/functests/1_performance/performance.go
+++ b/functests/1_performance/performance.go
@@ -1085,6 +1085,17 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 			})
 		})
 	})
+
+	Context("Verify containers specifications", func() {
+
+		It("imagePullPolicy should be ifNotPresent", func() {
+			pod, err := pods.GetPerformanceOperatorPod()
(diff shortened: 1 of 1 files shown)


functests/1_performance/performance.go
```diff
@@ -1085,6 +1085,17 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 			})
 		})
 	})
+
+	Context("Verify containers specifications", func() {
```
MarSik, line 1089: I am not entirely sure this belongs to functional tests. But even if it does, I would expect comments or link to the bug so future readers understand WHY we have such a test in the first place.

shaior: Sure, I will add a comment with a link to the bug. 
where do you think this tc should belong? 

thanks.

shaior: Hello @MarSik , @fedepaol,
can you please review the PR?

functests/1_performance/performance.go
```diff
@@ -1085,6 +1085,17 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+
+		It("imagePullPolicy should be ifNotPresent", func() {
+			pod, err := pods.GetPerformanceOperatorPod()
+			Expect(err).ToNot(HaveOccurred())
+
+			// verify that the pod uses ifNotPresent as imagePullPolicy
```
fromanirh, line 1096: redundant comment, what is being checked is evident from the code. If anything, better reference here the BZ or the issue which this test is covering.

fromanirh (commented):
in general, we prefer to squash the fixes in the fixing comment - not adding followup fixing commits.
In this case I think it's good enough. Let's consider the comments for the upcoming fixes.

fromanirh:
wait a sec: are we fixing this bug _first_ on branch 4.10? if so, why?

mrniranjan:
@fromanirh  because this bug affects 4.10 and below only . Do we want to fix it in master. I was under the impression that master branch is mostly backports of NTO

fromanirh:
> @fromanirh because this bug affects 4.10 and below only . Do we want to fix it in master. I was under the impression that master branch is mostly backports of NTO

ok, it makes sense to me to have fixes only for PAO as independent operator.

PR #935, opened 2022-08-17
Add ethtool -k and ethtool -l to must gather collected data

marioferh:
Add ethtool -k and ethtool -l to must gather collected data
Save in ethtool_features and ethtool_channels files


yanirq:
do we have a test for that ? or what is tested in a different method ?

marioferh:
> do we have a test for that ? or what is tested in a different method ?

the tests are in NTO: https://github.com/openshift/cluster-node-tuning-operator/pull/365/files

But it is needed to upload new data. I want to do a new test with real data generated in the cluster, when that PR will be merged I will add the test for these files also.

marioferh:
After this one we can add test in NTO
https://github.com/openshift/cluster-node-tuning-operator/pull/442

marioferh:
Test in: https://github.com/openshift/cluster-node-tuning-operator/pull/450

PR #938, opened 2022-08-22
hack: skip tests under 5_latency_testing

shajmakh:
The tests in 5_latency_testing should run only on-demand and by using different make target with relevant dependencies. Skip this package using ginkgo skipPackage flag.
additionally to some cleanup and fixes, see commits for more details.

### DIFF
--- a/functests/5_latency_testing/5_latency_testing_suite_test.go
+++ b/functests/5_latency_testing/5_latency_testing_suite_test.go
@@ -37,9 +37,7 @@ var prePullNamespace = &corev1.Namespace{
 var profile *performancev2.PerformanceProfile
 
 var _ = BeforeSuite(func() {
-	if !testclient.ClientsEnabled {
-		testlog.Errorf("client not enabled")
-	}
+	Expect(testclient.ClientsEnabled).To(BeTrue())
 
 	// update PP isolated CPUs. the new cpu set for isolated should have an even number of CPUs to avoid failing the pod on SMTAlignment error,
 	// and should be greater than what is requested by the test cases in the suite so the test runs properly
--- a/functests/5_latency_testing/5_latency_testing_suite_test.go
+++ b/functests/5_latency_testing/5_latency_testing_suite_test.go
@@ -94,7 +94,7 @@ var _ = AfterSuite(func() {
 		testlog.Info("Restore initial performance profile")
 		err = profilesupdate.ApplyProfile(profile)
 		if err != nil {
-			testlog.Error("could not restore the initial profile")
+			testlog.Errorf("could not restore the initial profile: %v", err)
 		}
 	}
 })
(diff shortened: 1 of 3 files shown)


functests/5_latency_testing/5_latency_testing_suite_test.go
```diff
@@ -102,6 +113,11 @@ var _ = AfterSuite(func() {
 func Test5LatencyTesting(t *testing.T) {
 	RegisterFailHandler(Fail)
 
+	if !isTestExecutableFound() {
+		testlog.Infof("The test executable file %q does not exist, skipping the suite.", testExecutablePath)
+		t.Skip()
```
Tal-or, line 118: continue the discussion from here: https://github.com/openshift-kni/performance-addon-operators/pull/930#discussion_r951280557 

If the make target failed to create the binary or the binary is missing for any other reason, what benefit do we have by skipping and not failing the test?
As you mentioned in the PR title - the suite works properly only if the binary exists.

shajmakh: the tests run on top of the binary. It is a precondition for this suite to have the binary exists, otherwise there is no point in running the tests. we skip and don't fail the tests because it is not visible to the tests why the binary is not found. OTOH running this suite is done by make target that includes generating the binary:
https://github.com/openshift-kni/performance-addon-operators/blob/master/Makefile#L271

 running the suite in another way is not considered a valid run. In case of failure to create the binary that will be reflected upon running the make target that creates the binary which runs this file:
https://github.com/openshift-kni/performance-addon-operators/blob/master/hack/build-latency-test-bin.sh

shajmakh: it is like requesting to have a PP on the cluster before running the functional tests, otherwise there is no point to run the suites. the reason for failing to create the PP should not be handled in each suite as we know this should be handled at earlier stage before getting to the phase of running the suite.

Tal-or: Basically, if the compilation failed here:
https://github.com/openshift-kni/performance-addon-operators/blob/master/hack/build-latency-test-bin.sh#L10 

The make target that depends on it would be failed as well.

shajmakh: then target commands will not be executed if the prerequisites of the target fail, so in this case, the suite will not even run.

Tal-or: Exactly so there is no chance for the suite to start without the executable

Tal-or: So why would we need the suite to skip in case the executable is missing?

shajmakh:
Following the offline conversation with @Tal-or we agreed to skip the suite ahead from the functests-only target and fail the suite in case the binary is not found.

functests/5_latency_testing/5_latency_testing_suite_test.go
shajmakh: https://github.com/openshift-kni/performance-addon-operators/pull/938#issuecomment-1225507397

Tal-or:
/lgtm
/approve
Thanks!

PR #939, opened 2022-08-26
must-gather: numaresources: Avoid error messages when NRO is not installed

jlojosnegros:
must-gather is trying to get NRO related information (CRDs, and
namespaced artifacts) no matter what and so, when NRO is not installed
in the cluster some nasty error messages appear because those elements
are not found.

These are not errors but to avoid missleading the users it is better to
avoid them by not gathering the NRO information if we cannot detect the
NRO namespace.
A small message appears in console in this case.

### DIFF
--- a/must-gather/collection-scripts/namespace
+++ b/must-gather/collection-scripts/namespace
@@ -22,8 +22,5 @@ function nro_namespace() {
   # namespace suggested by the documentation. This is a fancier way to check for its existence
   [ -z "${ns}" ] && ns=$(oc get ns openshift-numaresources -o=jsonpath='{.metadata.name}{"\n"}' 2> /dev/null )
 
-  # we should never get there. This is the last resort.
-  [ -z "${ns}" ] && ns="openshift-numaresources"
-
   echo ${ns}
 }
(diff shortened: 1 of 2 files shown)


must-gather/collection-scripts/namespace
```diff
@@ -22,8 +22,5 @@ function nro_namespace() {
   # namespace suggested by the documentation. This is a fancier way to check for its existence
   [ -z "${ns}" ] && ns=$(oc get ns openshift-numaresources -o=jsonpath='{.metadata.name}{"\n"}' 2> /dev/null )
 
-  # we should never get there. This is the last resort.
-  [ -z "${ns}" ] && ns="openshift-numaresources"
-
```
shajmakh, line 27: why delete this?

shajmakh (commented):
Thanks for the fixes! looks good to me
have one comment below.

must-gather/collection-scripts/namespace
jlojosnegros: This is the way we find out the NRO namespace is not defined.
We were using a "default" when everything else fails but if we cannot find the namespace and we reach this point is good to assume NRO is not installed.

shajmakh: makes sense, thanks for clarifying.

jlojosnegros:
I think the two failing ci lanes need to be overridden

PR #947, opened 2022-09-14
[WIP] PAO adaptation to Hypershift proposal

jlojosnegros:
This is the proposal to adapt Performance Profile Controller to Hypershift.

I have created this pull request for sake of discussion and comments. it is not intended to be merged.

/hold


docs/enhancements/hypershift-adaptation.md
```diff
@@ -0,0 +1,99 @@
...
+- [NTO Hypershift adaptation design document](https://docs.google.com/document/d/1G9uu_EBCu-X8OgEA5L0P1GZMcrASjQJ6qIYAJSS6NbY/edit)
+
+## Proposal
+
+### Performance Profile
+
+The proposal to handle the Performance Profile (PP) mimics the way NTO handles it `Tuned` API
+
+- New `.spec.performanceProfileConfig` will be created in `NodePool` CRD
+- [Cluster Service Consumer](https://hypershift-docs.netlify.app/reference/concepts-and-personas/) will create `PerformanceProfile` objects inside of configmaps, then reference these configmaps objects in the `.spec.performanceProfileConfig` field of the `NodePool` API
```
dagrayvid, line 45: This will depend on what the HyperShift team prefers, but I wonder if it would be better to use the same `.spec.tunedConfig` field for referencing ConfigMaps that contain PerformanceProfiles and Tuned? In the NodePool controller we can use type assertion to figure out which it is. The benefit being that it keeps the NodePool API simpler.

docs/enhancements/hypershift-adaptation.md
```diff
@@ -0,0 +1,99 @@
...
+  - This operation is __NOT__ in place already, so a modification in the NodePool Controller will be needed, but it would be quite similar to the one handling `MachineConfigs` right now.
+
+> __development NOTE__: I am not sure this is the proper way to handle this because I still have some doubts regarding the way NodePool Controller handles `MachineConfig` objects, that's why I have asked @dagrayvid about that
+>
+> Here is the question:
+>
+> - Regarding NTO generated machine configs. I have seen that nodepool controller extract all the machine configurations from the configmaps,
+>  then comprress them and store them into a core.v1.Secret in the hosted-control-plane-namespace ... and I do not understand this operation.
+>  I mean, in the end machineconfig operator should handle this machine configs to apply them in the hosted cluster, right?
+>  I thought there have been some adaptations in the MCO for hypershift to handle this, but I cannot find anything in the repo so any help will be really appreciated.
```
dagrayvid, line 93: As mentioned on Slack, there are a lot of changes in how MachineConfigs are applied in HyperShift. This enhancement will give a better explanation than I can https://github.com/openshift/enhancements/blob/master/enhancements/hypershift/node-lifecycle.md.

To add onto this, if you take a look at the operators running in the control plane namespace, there is no MCO, only the ignition-server. On the hosted cluster nodes, there is no MCD DaemonSet.

docs/enhancements/hypershift-adaptation.md
```diff
@@ -0,0 +1,99 @@
...
+  - label: `hypershift.openshift.io/nodePool` : `NodePool` API name where the `PeformanceProfile` which generate this `MachineConfig` was referenced.
+- This will trigger the reconcile operation in NodePool Controller for these objects.
+
+### MCO KubeletConfig
+
+__WIP__
+
+Being this object also handled by MachineConfig Operator (MCO) as `MachineConfig`s the proposal is to handle them in a similar way.
+
+- Once Performance Profile Controller has created the `Kubeletconfig` object as usual, it will embeded the object into a `configmap` in the `hosted-control-plane-namespace`.
```
dagrayvid, line 79: I haven't tried out creating KubeletConfigs yet on HyperShift. Unless I'm missing something, [they will not be reconciled and applied](https://github.com/openshift/hypershift/blob/338770449bd0b6e14df123d2c17846c774446a63/hypershift-operator/controllers/nodepool/nodepool_controller.go#L1448). Some experimentation / discussion with the HyperShift folks is probably necessary here.

dagrayvid (commented):
Thanks @jlojosnegros. Over all this seems in-line with how I imagined PAO would function in HyperShift. I left a few comments for discussion.

docs/enhancements/hypershift-adaptation.md
jlojosnegros: Yeah I totally agree with that, in fact I was just proposing to use a different field just in case you would prefer to avoid mixing PAO and NTO configs.

docs/enhancements/hypershift-adaptation.md
jlojosnegros: On the other hand maybe talking with Hypershift people would help to dissipate any doubt.

dagrayvid: I think you're right, I had misunderstood `getConfig`, and was thinking the manifest would be ignored if it were a KubeletConfig. It looks like a KubeletConfig is added to the token secret just like MachineConfigs, and should (somehow) be applied.

jlojosnegros:
enhancement open: openshift/enhancements#1244

PR #954, opened 2022-12-01
[release-4.10] OCPBUGS-4165: Add ktimer to PAO tuned profile

jlojosnegros:
Backport of openshift/cluster-node-tuning-operator#510


fromanirh:
do we have dependencies on tuned/rhcos components possibly not available on 4.10?

jlojosnegros:
> do we have dependencies on tuned/rhcos components possibly not available on 4.10?

We depend on the corresponding patch applied to the kernel but the ticket for that is closed so I think we are ok on that side.

MarSik (approved):
/lgtm

PR #964, opened 2023-07-05
e2e: latency testing: increase the expected threshold and fix gomega truncating output

shajmakh:
lately some of the latency tests were failing and it turned out that the failure is due to that the maximum latency found was greater than anticipated. The tests in this suite do not care about the measured latency because we do not care to tune the systems in the first place. Its only goal is to run the tests of the latency tools with different values of environment variables and validate that the tool actually runs with these parameters.

Increase the maximum latency threshold so that the test doesn't fail in case of a high latency result.

### DIFF
--- a/functests/5_latency_testing/latency_testing.go
+++ b/functests/5_latency_testing/latency_testing.go
@@ -20,7 +20,7 @@ import (
 	"github.com/openshift-kni/performance-addon-operators/functests/utils/profilesupdate"
 )
 
-//TODO get commonly used variables from one shared file that defines constants
+// TODO get commonly used variables from one shared file that defines constants
 const (
 	testExecutablePath = "../../build/_output/bin/latency-e2e.test"
 	//tool to test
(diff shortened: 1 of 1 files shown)


functests/5_latency_testing/latency_testing.go
```diff
@@ -97,6 +101,7 @@ type latencyTest struct {
 }
 
 var _ = Describe("Run tests of latency measurement tools with different values of latency environment variables", func() {
+	format.MaxLength = 0
```
yanirq, line 104: where is this used?

shajmakh: this is limiting the gomega output presentation by default to 4000 chars. we set it to 0 to disable this feature. 
check commit https://github.com/openshift-kni/performance-addon-operators/pull/964/commits/34614661dc46f5b4e1d7e8d50031d9dd784c490b

PR #965, opened 2023-07-18
e2e: add expected max latency to hwlatdetec test & rename constant

shajmakh:
The test started to fail frequently in CI. The reason for the failure is that there were measured latency samples that exceeds the maximal default threshold of hwlatdetec. Since we have enabled new CI machines, it seems that they are not well tuned hence causing in high latency.

To fix the test, just add a high threshold. Note that in these tests we do not care about latency measurments but about enabling the latency tools tests with differnt variables.

### DIFF
--- a/functests/5_latency_testing/latency_testing.go
+++ b/functests/5_latency_testing/latency_testing.go
@@ -81,7 +81,7 @@ const (
 
 	// we do not care about the actual system latency because CI systems are not tuned well enough to be an example for
 	// latency measuring, besides this suite only cares about testing the test executable with different values of env vars.
-	untunedLatencyThreshold = "900000"
+	untunedLatencyThreshold = "10000000" //10s
 	negativeTesting         = false
 	positiveTesting         = true
 )
--- a/functests/5_latency_testing/latency_testing.go
+++ b/functests/5_latency_testing/latency_testing.go
@@ -286,7 +286,7 @@ func getValidValuesTests(toolToTest string) []latencyTest {
 	if toolToTest == hwlatdetect {
 		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: successRuntime, testMaxLatency: "1", hwlatdetectMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
 		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: successRuntime, hwlatdetectMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
-		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: successRuntime, outputMsgs: []string{success}, toolToTest: toolToTest})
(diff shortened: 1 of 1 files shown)


functests/5_latency_testing/latency_testing.go
```diff
@@ -286,7 +286,7 @@ func getValidValuesTests(toolToTest string) []latencyTest {
 	if toolToTest == hwlatdetect {
 		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: successRuntime, testMaxLatency: "1", hwlatdetectMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
 		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: successRuntime, hwlatdetectMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
-		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: successRuntime, outputMsgs: []string{success}, toolToTest: toolToTest})
+		testSet = append(testSet, latencyTest{testRun: "true", testRuntime: successRuntime, testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
```
ffromani, line 289: the idea is fine, but the variable name suggests this is the expected latency for guaranteed workloads rather than an arbitrary huge value just to exercise the tool

ffromani (commented):
/approve

can we improve the naming? let's give it _one_ honest try

functests/5_latency_testing/latency_testing.go
MarSik: Yep and the value of it is also pretty low.

shajmakh: sure, we can update that. Is safeLatencyThreshold sounds better? 
I'll add a comment to clarify this constant usage and increase the threshold. If you have any preferred threshold in mind that is enough to be counted as safe you are welcome to share.

ffromani: what about `UntunedLatencyThreshold` or something like it?
Unless the only usage of `guaranteedLatency` is right here, I recommend adding (not replacing) a new constant + adding the comment you suggested to it

shajmakh: sounds better, thanks!
Yes, it's only used in these tests for the same reason (not to fail on high latency results)

shajmakh: > Yep and the value of it is also pretty low.

@MarSik the value was updated to 900000 in NTO. will update it here too.

shajmakh: actually this was addressed in another PR (I assume it wasn't merged once you reviewed this one):
https://github.com/openshift-kni/performance-addon-operators/pull/964

ffromani:
LGTM now. I think everything but the actual value of the threshold is fine. I'm wondering if we can't just an astronomically high value like 10s to best convey the intent here.
I'll let others decide.

MarSik (approved):

shajmakh:
4.10 is out of maintenance, closing the PR. will reopen if needed.

